PR #456, opened 2020-11-26
Add query to check if teller service has not been deleted. Closes #428

### LINKED ISSUE #428: Add Tiller Service (Helm v2) is not deleted query for Kubernetes
### Platform
Kubernetes

### Description
Check if there is any Tiller Service present

joaomartinscx:
Check if any Teller service remains undeleted. Closes #428

### DIFF
--- a/assets/queries/k8s/tiller_service_not_deleted/query.rego
+++ b/assets/queries/k8s/tiller_service_not_deleted/query.rego
@@ -1,0 +1,54 @@
+package Cx
+
+CxPolicy [ result ] {
+    document := input.document[i]
+    keyword := "tiller"
+
+    metadata := document.metadata
+    contains(metadata.name,keyword)
+
+	result := {
+                "documentId": 		input.document[i].id,
+                "searchKey": 	    sprintf("metadata.name=%s",[metadata.name]),
+                "issueType":		"IncorrectValue",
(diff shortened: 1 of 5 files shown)


assets/queries/k8s/tiller_service_not_deleted/query.rego
```diff
@@ -0,0 +1,54 @@
...
+    keyword := "tiller"
+
+    metadata := document.metadata
+    labels:= metadata.labels
+    some j
+        contains(labels[j],keyword)
+    
+	result := {
+                "documentId": 		input.document[i].id,
+                "searchKey": 	    sprintf("metadata.labels.%s", [j]),
```
joaomartinscx, line 30: @joelsou5a In this case, I wasn't able to search with the name in the metadata, because the key is at the same level

joelsou5a: I think you should then search it on metadata itself (metadata name) otherwise we could have problems with t

"searchKey": 	    sprintf("metadata.name=%s", [metadata.name]),

its not precise but we will not have problems with mismatch of the lines.

joaomartinscx: attended requested changes

joelsou5a (approved):

PR #486, opened 2020-11-27
Add cpu_limits query Closes #483

### LINKED ISSUE #483: Add cpu_limits query for Kubernetes
### Platform
Kubernetes

### Description
Add new "CPU Limits" query, that checks if CPU limits are set

rafaela-soares:
Since attribute 'containers' is an array, it is necessary to check if all containers have CPU limits.

Closes #483

### DIFF
--- a/assets/queries/k8s/cpu_limits/query.rego
+++ b/assets/queries/k8s/cpu_limits/query.rego
@@ -33,8 +33,8 @@ CxPolicy [ result ] {
                 "documentId":        input.document[i].id,
                 "issueType":         "MissingAttribute",
                 "searchKey":         sprintf("metadata.name=%s.%s.containers.name=%s.resources", [metadata.name, specInfo.path, containers[index].name]),
-                "keyExpectedValue":  sprintf("%s.containers.name=%s. does have limits defined",[specInfo.path, containers[index].name]),
-                "keyActualValue":    sprintf("%s.containers.name=%s. doesn't have limits defined",[specInfo.path, containers[index].name])
+                "keyExpectedValue":  sprintf("%s.containers.name=%s does have limits defined",[specInfo.path, containers[index].name]),
+                "keyActualValue":    sprintf("%s.containers.name=%s doesn't have limits defined",[specInfo.path, containers[index].name])
               }
 }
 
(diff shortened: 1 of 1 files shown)


assets/queries/k8s/cpu_limits/query.rego
```diff
@@ -0,0 +1,33 @@
+package Cx
```
joelsou5a, line 1: Check this implementation to a more specifiy searchKey

Attached rego query
[Uploading query.zip…]()

joelsou5a (changes requested):

assets/queries/k8s/cpu_limits/test/positive_expected_result.json
```diff
@@ -0,0 +1,7 @@
+[
+	{
+		"queryName": "CPU Limits",
+		"severity": "MEDIUM",
+		"line": 6
```
joelsou5a, line 5: "line": 11

joelsou5a (changes requested):

assets/queries/k8s/cpu_limits/query.rego
```diff
@@ -0,0 +1,32 @@
...
+
+CxPolicy [ result ] {
+    document := input.document[i]
+    specInfo := getSpecInfo(document)
+    metadata := document.metadata
+
+    containers := document.spec.containers
+    
+    
+    object.get(containers[index]["resources"]["limits"], "cpu", "undefined") == "undefined"
```
joaomartinscx, line 11: the query should detect when either _limits_ or _resource_ are not present in the file, e.g. if you remove the limits or resources attribute in the *negative.yaml* the query is still passing, when it should not. I suggest testing these different edge cases.

joaomartinscx (changes requested):

assets/queries/k8s/cpu_limits/query.rego
```diff
@@ -0,0 +1,71 @@
...
+ 
+    result := {
+                "documentId":        input.document[i].id,
+                "issueType":         "MissingAttribute",
+                "searchKey":         sprintf("metadata.name=%s.%s.containers.name=%s.resources.limits", [metadata.name, specInfo.path, containers[index].name]),
+                "keyExpectedValue":  sprintf("%s.containers.name=%s does have CPU limits",[specInfo.path, containers[index].name]),
```
joaomartinscx, line 17: Perhaps it's clearer to specify the path where the information should be added, like in this case:
sprintf("'%s.containers[%s].resources.limits.cpu' is set",[specInfo.path, index])

joaomartinscx (changes requested):

ruigomescx (approved):

PR #560, opened 2020-11-30
Added Restart Policy With 5 Retries query #559

### LINKED ISSUE #559: Add Restart Policy With 5 Retries query for Dockerfile
### Platform
Docker

### Description
It's important that the restart policy is 'on-failure' up to 5 attempts

joelcarvalhocheckmarx:
It's important that the restart policy is 'on-failure' up to 5 attempts.

Closes #559


ruigomescx:
This is a docker-compose query, we will not support it for now.

Issue #572, opened 2020-12-02
Add Kubernetes Image should use digest query for Kubernetes


pedro-mimoso:
### Platform
Kubernetes

### Description
Sees if Kubernetes image has digest on

PR #573, opened 2020-12-02
Add Kubernetes Image should use digest query for Kubernetes #572

pedro-mimoso:
Sees if Kubernetes image has digest on #572.

### DIFF
--- a/assets/queries/k8s/kubernetes_image_digest/query.rego
+++ b/assets/queries/k8s/kubernetes_image_digest/query.rego
@@ -1,0 +1,20 @@
+package Cx
+
+CxPolicy [ result ] {
+	document := input.document[i]
+  metadata:= document.metadata
+  spec := document.spec
+  containers := spec.containers
+  not containers[c].imagePullPolicy == "Always"
+    
+	
+
+	result := {
+                "documentId": 		input.document[i].id,
+                "searchKey": 	   sprintf("metadata.name=%s.spec.containers.name=%s.imagePullPolicy", [metadata.name, containers[c].name]),
+                "issueType":		"IncorrectValue",
+                "keyExpectedValue": sprintf("spec[%s].containers[%s].imagePullPolicy is Always", [metadata.name, containers[c].name]),
+                "keyActualValue": 	sprintf("spec[%s].containers[%s].imagePullPolicy is not Always", [metadata.name, containers[c].name])
+              }
+}
+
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/google/uuid v1.1.2
 	github.com/hashicorp/hcl/v2 v2.7.0
 	github.com/mailru/easyjson v0.7.6
-	github.com/moby/buildkit v0.7.2
+	github.com/moby/buildkit v0.7.2 // indirect
```
joaomartinscx, line 13: I don't know much about the significance of this but doesn't seem necessary

assets/queries/k8s/kubernetes_image_digest/query.rego
```diff
@@ -0,0 +1,20 @@
+package Cx
+
+CxPolicy [ result ] {
+	document := input.document[i]
+  metadata:= document.metadata
+  spec := document.spec
+  containers := spec.containers
+  not containers[c].imagePullPolicy == "Always"
```
joaomartinscx, line 8: I think this verification belongs to another query, here we should look into the "image" tag and check if the digest is present, the "@".

joaomartinscx (changes requested):

assets/queries/k8s/kubernetes_image_digest/query.rego
```diff
@@ -0,0 +1,38 @@
...
+  not contains(image, "@")
+    
+	
+
+	result := {
+                "documentId": 		input.document[i].id,
+                "searchKey": 	   sprintf("metadata.name=%s.spec.containers.name=%s.image", [metadata.name, containers[c].name]),
+                "issueType":		"IncorrectValue",
+                "keyExpectedValue": sprintf("spec[%s].containers[%s].image has '@'", [metadata.name, containers[c].name]),
+                "keyActualValue": 	sprintf("spec[%s].containers[%s].image doesn't have '@'", [metadata.name, containers[c].name])
```
joaomartinscx, line 18: Metadata name comes before Spec, so you shouldn't search for Spec[Metadata.name]. At this point, it is not really necessary since the search key tracks down the faulty resource, but if you want to place it, do it like in the SearchKey.

assets/queries/k8s/kubernetes_image_digest/query.rego
```diff
@@ -0,0 +1,38 @@
...
+  object.get(containers[k],"image","undefined") == "undefined"
+    
+	
+
+	result := {
+                "documentId": 		input.document[i].id,
+                "searchKey": 	   sprintf("metadata.name=%s.spec.containers", [metadata.name]),
+                "issueType":		"MissingAttribute",
+                "keyExpectedValue": sprintf("spec[%s].containers[%s] is Defined", [metadata.name, containers[c].name]),
+                "keyActualValue": 	sprintf("spec[%s].containers[%s] is Undefined", [metadata.name, containers[c].name])
```
joaomartinscx, line 36: Same thing here

joaomartinscx (changes requested):

joaomartinscx (commented):
Everything looks good to me, I recommend approval and merge of this PR

ruigomescx (approved):

Issue #592, opened 2020-12-03
Add Docker Daemon Socket is Exposed to Containers query for Kubernetes


pedro-mimoso:
### Platform
Kubernetes

### Description
Sees if Docker Daemon Socket is not exposed to Containers

PR #593, opened 2020-12-03
Add Docker Daemon Socket is Exposed to Containers query for Kubernetes #592

pedro-mimoso:
Sees if Docker Daemon Socket is not exposed to Containers #592

### DIFF
--- a/assets/queries/k8s/kubernetes_docker_socket_volume/query.rego
+++ b/assets/queries/k8s/kubernetes_docker_socket_volume/query.rego
@@ -14,8 +14,8 @@ CxPolicy [ result ] {
                 "documentId": 		input.document[i].id,
                 "searchKey": 	   sprintf("metadata.name=%s.spec.volumes.name=%s.hostPath.path", [metadata.name, volumes[c].name]),
                 "issueType":		"IncorrectValue",
-                "keyExpectedValue": sprintf("spec[%s].volumes[%s].hostPath.path is not '/var/run/docker.sock'", [metadata.name, volumes[c].name]),
-                "keyActualValue": 	sprintf("spec[%s].volumes[%s].hostPath.path is '/var/run/docker.sock'", [metadata.name, volumes[c].name])
+                "keyExpectedValue": sprintf("spec.volumes[%s].hostPath.path is not '/var/run/docker.sock'", [volumes[c].name]),
+                "keyActualValue": 	sprintf("spec.volumes[%s].hostPath.path is '/var/run/docker.sock'", [volumes[c].name])
               }
 }
 
@@ -34,8 +34,8 @@ CxPolicy [ result ] {
                 "documentId": 		input.document[i].id,
                 "searchKey": 	   sprintf("metadata.name=%s.spec.volumes.name=%s.hostPath.path", [metadata.name, volumes[c].name]),
                 "issueType":		"IncorrectValue",
-                "keyExpectedValue": sprintf("spec[%s].volumes[%s].hostPath.path is not '/var/run/docker.sock'", [metadata.name, volumes[c].name]),
-                "keyActualValue": 	sprintf("spec[%s].volumes[%s].hostPath.path is '/var/run/docker.sock'", [metadata.name, volumes[c].name])
+                "keyExpectedValue": sprintf("spec.volumes[%s].hostPath.path is not '/var/run/docker.sock'", [volumes[c].name]),
+                "keyActualValue": 	sprintf("spec.volumes[%s].hostPath.path is '/var/run/docker.sock'", [ volumes[c].name])
               }
 }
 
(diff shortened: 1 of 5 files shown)


assets/queries/k8s/kubernetes_docker_socket_volume/query.rego
```diff
@@ -0,0 +1,19 @@
+package Cx
+
+CxPolicy [ result ] {
+	document := input.document[i]
+    metadata:= document.metadata
```
joelsou5a, line 5: validate if the kind is 'pod'

assets/queries/k8s/kubernetes_docker_socket_volume/query.rego
```diff
@@ -0,0 +1,19 @@
...
+	
+
+	result := {
+                "documentId": 		input.document[i].id,
+                "searchKey": 	   sprintf("metadata.name=%s.spec.volumes.name=%s.hostPath.path", [metadata.name, volumes[c].name]),
+                "issueType":		"IncorrectValue",
+                "keyExpectedValue": sprintf("spec[%s].volumes[%s].hostPath.path is not '/var/run/docker.sock'", [metadata.name, volumes[c].name]),
+                "keyActualValue": 	sprintf("spec[%s].volumes[%s].hostPath.path is '/var/run/docker.sock'", [metadata.name, volumes[c].name])
+              }
+}
```
joelsou5a, line 19: This seems good if it is a pod, but we also should validate the template if it is not a pod and also check the cronjob

joelsou5a (changes requested):

assets/queries/k8s/kubernetes_docker_socket_volume/query.rego
joaomartinscx: And to save you some time trying to fix conflicts, I suggest testing these 3 cases separately, and not use a function common to all cases, even though all that changes is the path where the evaluated attribute is.

assets/queries/k8s/kubernetes_docker_socket_volume/query.rego
```diff
@@ -0,0 +1,55 @@
...
+	
+
+	result := {
+                "documentId": 		input.document[i].id,
+                "searchKey": 	    document.kind,
+                "issueType":		"IncorrectValue",
+                "keyExpectedValue": "document.kind is not Pod or CronJob",
+                "keyActualValue": 	"document.kind is a Pod or CronJob",
+              }
+}
```
joaomartinscx, line 36: When the resource is neither a Pod nor a CronJob, it's still evaluated, the path where you find the volume hostPath is different, but it's also supposed to check in Resource.Spec.Template.Spec.

joaomartinscx (changes requested):

assets/queries/k8s/kubernetes_docker_socket_volume/query.rego
joaomartinscx: @pedro-mimoso Please take a look here, when it's not Pod or CronJob, you're still supposed to check, just in a different path.

assets/queries/k8s/kubernetes_docker_socket_volume/query.rego
```diff
@@ -0,0 +1,59 @@
...
+    volumes[c].hostPath.path == "/var/run/docker.sock"
+
+	
+
+	result := {
+                "documentId": 		input.document[i].id,
+                "searchKey": 	   sprintf("metadata.name=%s.spec.volumes.name=%s.hostPath.path", [metadata.name, volumes[c].name]),
+                "issueType":		"IncorrectValue",
+                "keyExpectedValue": sprintf("spec[%s].volumes[%s].hostPath.path is not '/var/run/docker.sock'", [metadata.name, volumes[c].name]),
+                "keyActualValue": 	sprintf("spec[%s].volumes[%s].hostPath.path is '/var/run/docker.sock'", [metadata.name, volumes[c].name])
```
joaomartinscx, line 18: metadata is at the same level as Spec, not inside

assets/queries/k8s/kubernetes_docker_socket_volume/query.rego
```diff
@@ -0,0 +1,59 @@
...
```
joaomartinscx, line 38: same thing here

assets/queries/k8s/kubernetes_docker_socket_volume/query.rego
```diff
@@ -0,0 +1,59 @@
...
```
joaomartinscx, line 57: here too

joaomartinscx (dismissed):

joaomartinscx (commented):
Everything looks good to me, I recommend approval and merge of this PR.

joelsou5a (approved):

joelsou5a (approved):

PR #602, opened 2020-12-03
Add query to check if API version is deprecated. Closes #601

### LINKED ISSUE #601: Add Object Is Using A Deprecated API Version query for Kubernetes
### Platform
Kubernetes

### Description
Check if any objects are using a deprecated version of API.

joaomartinscx:
Check if some objects are using a deprecated API version. Closes #601

### DIFF
--- a/assets/queries/k8s/apiversion_is_deprecated/query.rego
+++ b/assets/queries/k8s/apiversion_is_deprecated/query.rego
@@ -1,0 +1,32 @@
+package Cx
+
+CxPolicy [ result ] {
+    document := input.document[i]
+    metadata := document.metadata
+
+    recommendedVersions := {
+                            "extensions/v1beta1" : {
+                                                      "Deployment" : "apps/v1",
+                                                      "DaemonSet" : "apps/v1",
+                                                      "Ingress" : "apps/v1"
+                                                   },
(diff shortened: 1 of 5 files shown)


assets/queries/k8s/apiversion_is_deprecated/query.rego
```diff
@@ -0,0 +1,32 @@
...
+                                                      "Deployment" : "apps/v1",
+                                                      "DaemonSet" : "apps/v1",
+                                                      "StatefulSet" : "apps/v1"
+                                                   }
+                          }
+    object.get(recommendedVersions,document.apiVersion,"undefined") != "undefined"
+    object.get(recommendedVersions[document.apiVersion],document.kind,"undefined") != "undefined"
+    result := {
+                "documentId": 		input.document[i].id,
+                "searchKey": 	    sprintf("apiVersion=%s",[document.apiVersion]),
```
joaomartinscx, line 27: @joelsou5a In this case I'm not using the metadata key because apiVersion comes before, so if I identify apiVersion after metadata, it will catch the next occurrence of _apiVersion_in the file, and if I put the metadata key after apiVersion, it catches the metadata key and not apiVersion.

ruigomescx (approved):

PR #612, opened 2020-12-04
Feature/add query copy instead of add

joaoReigota1:


assets/queries/dockerfile/copy_instead_of_add/query.rego
```diff
@@ -0,0 +1,21 @@
+package Cx
+
+CxPolicy [ result ] {
+  resource := input.document[i].command[j]
+  check(resource, input.document[i].command[j+1]) == true
```
ruigomescx, line 5: We should change the logic of this index access, to ensure we always have a next element. Check the length of the array before and compare with the current index

assets/queries/dockerfile/copy_instead_of_add/query.rego
```diff
@@ -0,0 +1,21 @@
...
+                "keyExpectedValue": "Using Command COPY",
+                "keyActualValue": 	"Using Command ADD"
+              }
+}
+
+check(com, nextCom) = true{
+  com.Cmd == "add"
+  not nextCom.Cmd == "run"
+  value := nextCom.Value[_]
+  not contains(value, "tar")
```
ruigomescx, line 20: check for .tar to be more specific

ruigomescx (changes requested):

markmishaevcx (approved):

joaomartinscx:
> _No description provided._

Please refer to the issue that is being addressed in this PR

assets/queries/dockerfile/copy_instead_of_add/metadata.json
```diff
@@ -0,0 +1,8 @@
+{
+  "id": "Copy_instead_of_add",
+  "queryName": "Copy Instead of Add",
```
joaomartinscx, line 3: The query name should identify the problem, in this case, that the Dockerfile is using ADD command instead of COPY

assets/queries/dockerfile/copy_instead_of_add/query.rego
```diff
@@ -0,0 +1,23 @@
+package Cx
+
+CxPolicy [ result ] {
+  resource := input.document[i].command[j]
+  check(resource) 
+  not tarfileChecker(resource.Value, "tar")
```
joaomartinscx, line 6: Maybe should be a little more specific and check with ".tar" and/or ".tar.", so that the extension is captured and not some filename containing tar, like "tarte.txt"

assets/queries/dockerfile/copy_instead_of_add/query.rego
```diff
@@ -0,0 +1,23 @@
...
+  
+	result := {
+                "documentId": 		input.document[i].id,
+                "searchKey": 	    sprintf("ADD=%s", [resource.Value[0]]),
+                "issueType":		"IncorrectValue", 
+                "keyExpectedValue": "Using Command COPY",
+                "keyActualValue": 	"Using Command ADD"
```
joaomartinscx, line 13: Try to identify if possible the command being used inappropriately, so the user can track it down quicker.

joaomartinscx (changes requested):

markmishaevcx:
@ruigomescx What is our decision regarding this PR, do you want me to merge it?  I also modified codacy exclusion files to ignore queries folder.

ruigomescx (approved):

PR #675, opened 2020-12-07
add query IAM Password Has Min Length #674

### LINKED ISSUE #674: Add IAM Password Without Minimum Length query for CloudFormation
### Platform
CloudFormation

### Provider
AWS

### Description
IAM user resource Login Profile Password should have at least 14 characters

joaoReigota1:
Closes #674


assets/queries/cloudFormation/iam_password_has_min_length/metadata.json
```diff
@@ -0,0 +1,8 @@
+{
+  "id": "IAM_password_has_min_length",
+  "queryName": "IAM Password Has Min Length",
```
ruigomescx, line 3: The query name should reflect the vulnerability, not the fix. We should use something like the reverse of what it's now. Expand Min to Minimum

ruigomescx (changes requested):

ruigomescx (approved):

PR #683, opened 2020-12-07
Added new query - RBAC_wildcard

soniantao:
Closes #683


assets/queries/k8s/RBAC_wildcard/metadata.json
```diff
@@ -0,0 +1,8 @@
+{
+  "id": "RBAC_wildcard_in_rules",
+  "queryName": "RBAC_wildcard_in_rules",
```
ruigomescx, line 3: Query name should be something like: RBAC Wildcard In Rules

assets/queries/k8s/RBAC_wildcard/query.rego
```diff
@@ -0,0 +1,105 @@
+package Cx
+
+CxPolicy [result]  {
+	document := input.document[i]
+    metadata := document.metadata
+    document.kind == "ClusterRole"
```
ruigomescx, line 6: We could check the document.kind == ["Role", "ClusterRole"] and could delete the last 3 rules

ruigomescx (changes requested):

assets/queries/k8s/RBAC_wildcard/query.rego
soniantao: I've changed for functions

ruigomescx (approved):

PR #704, opened 2020-12-07
add Cloud Trail Multi Region #703

### LINKED ISSUE #703: Add/Update Cloud Trail Multi Region query for CloudFormation
### Platform
CloudFormation

### Provider
AWS

### Description
AWS CloudTrail should have IsMultiRegionTrail set to true

joaoReigota1:
Closes #703


assets/queries/cloudFormation/cloud_trail_multi_region_disabled/query.rego
```diff
@@ -0,0 +1,19 @@
...
+  not checkRegion(resource)
+
+	result := {
+                "documentId": 		input.document[i].id,
+                "searchKey": 	    sprintf("Resources.%s.Properties", [name]),
+                "issueType":		"IncorrectValue", 
+                "keyExpectedValue": sprintf("'Resources.%s.Properties.IsMultiRegionTrail' is true", [name]),
+                "keyActualValue": 	sprintf("'Resources.%s.Properties.IsMultiRegionTrail' is false", [name])
+              }
+}
```
Ruben-Silva, line 15: Query should be divided into two. This is correct, however, if you divide into one query that checks if 'isMultiRegionTrail' is undefined and another that checks, when defined, if it is set to false. This way, you can be more specific in the type of issue, because when 'undefined' it should be 'MissingAttribute' and when 'false' it should be 'IncorrectValue'. Additionally, with the 'IncorrectValue' query I mentioned, the searchKey should be 'sprintf("Resources.%s.Properties.IsMultiRegionTrail", [name])'.

ruigomescx (approved):

PR #729, opened 2020-12-07
add Cloud Trail Log File Validation #728

### LINKED ISSUE #728: Add Cloud Trail Log File Validation query for CloudFormation
### Platform
CloudFormation

### Provider
AWS

### Description
CloudTrail log file validation should be enabled

joaoReigota1:
Closes #728


assets/queries/cloudFormation/cloud_trail_log_file_validation_disabled/query.rego
```diff
@@ -0,0 +1,24 @@
...
+              }
+}
+
+checkFileValidation(cltr) {
+	object.get(cltr.Properties , "EnableLogFileValidation", "not found") == "not found"
+}
+
+checkFileValidation(cltr) {
+	object.get(cltr.Properties , "EnableLogFileValidation", "not found") == false
+}
```
Ruben-Silva, line 23: This is correct, however, you should have two separate queries, one for when its undefined and another when its false. This way, you can be more specific in the "issueType", being "MissingAttribute" when undefined, and "IncorrectValue" when false. Additionally, in the IncorrectValue query, the searchKey should be "sprintf("Resources.%s.Properties.EnableLogFileValidation", [name])".

Ruben-Silva (changes requested):

PR #776, opened 2020-12-10
Added new query for Custom vs default SSL certificates

### LINKED ISSUE #775: Add Vulnerable_Default_SSL_Certificate query for AWS
### Platform
AWS

### Provider
AWS

### Description
"Custom SSL certificates allow only defined users to access content by using alternate domain name instead the default one

soniantao:
Closes #775


assets/queries/cloudFormation/CustomSSLCertificates/query.rego
```diff
@@ -0,0 +1,15 @@
+package Cx
+
+CxPolicy [result] {
+  resource := input.document[i].Resources[name]
+  certificateSource := input.document[i].Resources[name].Properties.DistributionConfig.ViewerCertificate.certificateSource
```
joaomartinscx, line 5: Check the resource Type

assets/queries/cloudFormation/CustomSSLCertificates/query.rego
```diff
@@ -0,0 +1,15 @@
...
+  expectedvalue := "cloudfront"
+  certificateSource == expectedvalue
```
joaomartinscx, line 7: should use contains() instead of == operator

joaomartinscx: Because the name shouldn't be like "%cloudfront%", which means the string might not be exactly 'cloudfront'

joaomartinscx (changes requested):

assets/queries/cloudFormation/CustomSSLCertificates/query.rego
soniantao: Hi João :)
After a deep research I found that the other was not accurate. There are only 4 CertificateSource exact types: cloudfront, iam, acm and UNKNOWN_TO_SDK_VERSION (this one is not so used).

for cloudFront, if you ran the follow cmd to checkout the a well implemented distribution config you'll get the follow:
$ aws cloudfront get-distribution-config --id xxxxxxxxxx
{
    "DistributionConfig": {
        "ViewerCertificate": {
            "CertificateSource": "cloudfront",
            "CloudFrontDefaultCertificate": true,
            "MinimumProtocolVersion": "SSLv3"
        },
        
This is why I kept only cloudfront and not %cloudfront% :)

assets/queries/cloudFormation/CustomSSLCertificates/query.rego
soniantao: I forgot it. thanks!

joaomartinscx (commented):
Everything looks good to me, I recommend the approval and merge of this PR.

joelsou5a (approved):

Issue #830, opened 2020-12-11
Add Service Type is NodePort query for Kubernetes


pedro-mimoso:
### Platform
Kubernetes

### Description
Service type should not be NodePort

PR #831, opened 2020-12-11
Add Service Type is NodePort query for Kubernetes #830

pedro-mimoso:
Service type should not be NodePort #830


assets/queries/k8s/service_type_is_nodeport/query.rego
```diff
@@ -0,0 +1,17 @@
+package Cx
+
+CxPolicy [ result ] {
+	  document := input.document[i]
+    metadata:= document.metadata
+    spec := document.spec
```
Ruben-Silva, line 6: Add a check to see if the kind is service.
lower(document.kind) == "service"

Ruben-Silva (changes requested):

ruigomescx (approved):

PR #870, opened 2020-12-11
Add VPC Has FlowLog Disabled query

### LINKED ISSUE #869: Add VPC Has FlowLog Disabled query for CloudFormation
### Platform
CloudFormation

### Provider
AWS

### Description
VPC Hasn´t any FlowLog associated, this !Ref in FlowLog.ResourceId must be valid

fabioGoncalvesCx:
Added new query for AWS CloudFormation.

Provide a CloudFormation resource and a resource to ensure that FlowLog.ResourceId has a valid !Ref to the VPC

Closes #869


assets/queries/cloudFormation/vpc_has_FlowLog_disabled/query.rego
```diff
@@ -0,0 +1,34 @@
+package Cx
+
+CxPolicy [ result ] {
+  resource := input.document[i].Resources[name]
+  resource.Type = "AWS::EC2::FlowLog"
```
rogeriopeixotocx, line 5: I think you should try the other way around. Find the VPC first, then check if there's any resources of type `AWS::EC2::FlowLog` in the same document

fabioGoncalvesCx: Done

assets/queries/cloudFormation/vpc_has_FlowLog_disabled/query.rego
```diff
@@ -0,0 +1,34 @@
...
+                "keyActualValue": 	sprintf("Resources.%s has FlowLogs disabled", [name])
+              }
+}
+
+CheckFlowLogExistance (service) = result {
+    
+	documents := input.document[index].Resources[a]
+  documents.Type = "AWS::EC2::FlowLog"
+    
+  result := contains(documents, service)
```
ruigomescx, line 24: the contains function here could be deleted and we could return directly since it's a simple operation:
`result := documents[_].ResourceId == service`

ruigomescx (changes requested):

assets/queries/cloudFormation/vpc_has_FlowLog_disabled/query.rego
fabioGoncalvesCx: Done

ruigomescx (approved):

PR #889, opened 2020-12-14
Add acm_expired_certificates query Closes #887

### LINKED ISSUE #887: Add acm_expired_certificates query for AWS CloudFormation
**Platform**
AWS CloudFormation

**Provider**
AWS

**Description**
Check if they are any expired certificates in the ACM

NunoAraujoCX:
Closes #887 

Check if they are any expired certificates in the AWS Certificate Manager

### DIFF
--- a/assets/queries/cloudFormation/acm_expired_certificates/query.rego
+++ b/assets/queries/cloudFormation/acm_expired_certificates/query.rego
@@ -1,0 +1,14 @@
+package Cx
+
+CxPolicy [ result ] {
+  resource := input.document[i].Resources[name]
+  resource.Status == "EXPIRED"
+
+	result := {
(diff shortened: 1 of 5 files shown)


assets/queries/cloudFormation/acm_expired_certificates/query.rego
```diff
@@ -0,0 +1,14 @@
+package Cx
+
+CxPolicy [ result ] {
+  resource := input.document[i].Resources[name]
+  resource.Status == "EXPIRED"
```
ruigomescx, line 5: Checking the reference for AWS::CertificateManager::Certificate, there is no field Status. I think we cannot check this query via resource. Please review the query

ruigomescx (changes requested):

PR #901, opened 2020-12-14
added new query for iam user with passwordResetRequired config

### LINKED ISSUE #900: Add user IAM missing PasswordResetRequired query for cloudformation
### Platform
cloudformation

### Provider
cloudformation

### Description
IAM User Login Profile should exist and have PasswordResetRequired property set to true

soniantao:
Closes #901
Closes #900


assets/queries/cloudFormation/iam_user_passwordResetRequired_false/metadata.json
```diff
@@ -0,0 +1,8 @@
+{
+  "id": "iam_user_missing_PasswordResetRequired",
+  "queryName": "user IAM missing PasswordResetRequired",
+  "severity": "MEDIUM",
+  "category": null,
```
joaomartinscx, line 5: Missing category

assets/queries/cloudFormation/iam_user_passwordResetRequired_false/query.rego
```diff
@@ -0,0 +1,44 @@
+package Cx
+
+CxPolicy [result] {
+  resource := input.document[i].Resources[name]
+  resource.Type == "AWS::IAM::User"
+  loginProfile := resource.Properties.LoginProfile
+  loginProfile != []
+  loginProfile != null
```
joaomartinscx, line 8: Isn't LoginProfile an object? Since in this test case you are assuming its present (or not testing its presence), you won't need to check if it is defined

assets/queries/cloudFormation/iam_user_passwordResetRequired_false/query.rego
```diff
@@ -0,0 +1,44 @@
...
+CxPolicy [result] {
+  resource := input.document[i].Resources[name]
+  resource.Type == "AWS::IAM::User"
+  passwordResetRequired := resource.Properties.LoginProfile
+  count(passwordResetRequired) == 1
+  passwordResetRequired.Password
+  result := {
+                "documentId": 		input.document[i].id,
+                "searchKey": 	    sprintf("Resources.%s.Properties.LoginProfile", [name]),
+                "issueType":		"IncorrectValue",
```
joaomartinscx, line 40: Here should be "MissingAttribute"

assets/queries/cloudFormation/iam_user_passwordResetRequired_false/query.rego
```diff
@@ -0,0 +1,44 @@
...
+}
+CxPolicy [result] {
+  resource := input.document[i].Resources[name]
+  resource.Type == "AWS::IAM::User"
+  loginProfile := resource.Properties
+  object.get(loginProfile, "LoginProfile", "undefined") == "undefined"
+   result := {
+                "documentId": 		input.document[i].id,
+                "searchKey": 	    sprintf("Resources.%s.Properties", [name]),
+                "issueType":		"IncorrectValue",
```
joaomartinscx, line 26: Here should be "MissingAttribute"

joaomartinscx (changes requested):

assets/queries/cloudFormation/iam_user_passwordResetRequired_false/query.rego
soniantao: Hi João.
I've done this for the follow example
Properties:
       Path: "/"
       LoginProfile:
         Password: myP@ssW0rd
         PasswordResetRequired: true
       Policies: ...
         
This is what we are expecting. However, a user can define this object empty like this:
Properties:
        Path: "/"
        LoginProfile:
        Policies:
        ...


It's a valid case where a password it's not defined for a user for example at this stage but there is a password defined for the group where this user is inserted in.
So I've not only checked for the object presence but also if it has content.

assets/queries/cloudFormation/iam_user_passwordResetRequired_false/metadata.json
soniantao: Yes. These queries were all without category defined. However, according to previous ones, I assume the best one for this is "Identity and Access Management"

joaomartinscx: Yes, that is right!

assets/queries/cloudFormation/iam_user_passwordResetRequired_false/query.rego
joaomartinscx: I'm still a bit skeptical about having LoginProfile empty, if there's nothing to add why place it? The Group password won't be referenced by some user id, not having anything to do with this LoginProfile? Could you please provide a link or reference to that example you mention so I can dig a little deeper on this issue?

joaomartinscx (changes requested):

assets/queries/cloudFormation/iam_user_passwordResetRequired_false/query.rego
soniantao: Hi João :)

after our discussion in Teams, I've changed this query since I did not checked the next line after these 2.
This solution was not well implemented :)
I've already done new changes

joaomartinscx (commented):
Everything looks good to me, I recommend the approval and merge of this PR

joaomartinscx (approved):

Issue #792, opened 2020-12-10
Add S3 Bucket ACL Allows Read to AllUsers query for CloudFormation


rogeriopeixotocx:
### Platform
*CloudFormation*

### Description
*S3 Bucket ACL should not allow read operations to 'AllUsers' grantee*

rogeriopeixotocx:
Waiting for AWS account to confirm if **PublicAccessBlockConfiguration** needs to be considered

@ruigomescx @NunoAraujoCX

PR #905, opened 2020-12-14
Add S3 Bucket ACL Allows Read to AllUsers query for CloudFormation

rogeriopeixotocx:
Platform
_CloudFormation_

Description
_S3 Bucket ACL should not allow read operations to 'AllUsers' grantee_


rogeriopeixotocx:
Closes #792

joelsou5a (approved):

PR #927, opened 2020-12-14
Add RDS Public Interface query for AWS CloudFormation

### LINKED ISSUE #926: Add RDS Public Interface query for AWS CloudFormation
### Platform
AWS CloudFormation

### Provider
AWS

### Description
RDS must not be defined with public interface, which means the attribute 'PubliclyAccessible' must be set to false.

Ruben-Silva:
Adding RDS Public Interface query for AWS CloudFormation, that checks if the attribute 'PubliclyAccessible' is set to false.

Closes #926


assets/queries/cloudFormation/rds_public_interface/query.rego
```diff
@@ -0,0 +1,29 @@
...
+  object.get(resource.Properties, "PubliclyAccessible", "undefined") == "undefined"
+
+	result := {
+                "documentId": 		input.document[i].id,
+                "searchKey": 	    sprintf("Resources.%s.Properties",  [name]),
+                "issueType":		"MissingAttribute", 
+                "keyExpectedValue": sprintf("'Resources.%s.Properties.PubliclyAccessible' is defined", [name]),
+                "keyActualValue": 	sprintf("'Resources.%s.Properties.PubliclyAccessible' is undefined", [name]),
+              }
+}
```
ruigomescx, line 15: > The default behavior value depends on your VPC setup and the database subnet group. For more information, see the PubliclyAccessible parameter in CreateDBInstance in the Amazon RDS API Reference.

Since the default value is variable, we should not check if the attribute is missing and assume that it's public available. We can cut this part of the query

ruigomescx (changes requested):

assets/queries/cloudFormation/rds_public_interface/query.rego
Ruben-Silva: Understood. Thanks.

ruigomescx (approved):

PR #986, opened 2020-12-15
Add gem_install_without_version Closes #983

### LINKED ISSUE #983: Add pin_versions_in_gem_install query for Docker
**Platform**
Docker

**Provider**
Dockerfile

**Description**
We should use 'gem install <gem>:<version> instead of 'gem install <gem>'

NunoAraujoCX:
Closes #983 

Check if we are using 'gem install gem:version' instead of 'gem install gem'


assets/queries/dockerfile/pin_versions_in_gem_install/metadata.json
```diff
@@ -0,0 +1,8 @@
+{
+  "id": "pin_versions_in_gem_install",
+  "queryName": "Pin Versions In Gem Install",
```
joaoReigota1, line 3: The query name should reflect the vulnerability, not the fix

joaoReigota1 (changes requested):

joaoReigota1 (commented):
Approved

joelsou5a (approved):

PR #1060, opened 2020-12-16
Add query to check if apt-get is called without yes flag. Closes #1059

### LINKED ISSUE #1059: Add APT-GET Missing "-y" To Avoid Manual Input  query for Docker
### Platform
Docker

### Description
Check if apt-get calls use the flag -y to avoid user manual input.

joaomartinscx:
Check if any apt-get call does not use "-y" flag to avoid manual user input. Closes #1059


joaoReigota1 (commented):
Approve

joelsou5a (approved):

PR #1143, opened 2020-12-18
Add copy_from_without_from_alias_defined_previously Closes #1139 

### LINKED ISSUE #1139: Add COPY --from Without FROM Alias Defined Previously query for Docker
### Platform
Docker

### Description
Add a new "COPY --from Without FROM Alias Defined Previously" query, that checks if COPY --from references a previous defined FROM alias

rafaela-soares:
For this query, it is necessary to check if the alias referenced by COPY --from (COPY --from=ALIAS ....) is a previously defined FROM alias (FROM myimage:tag as ALIAS).

To do that, we need to check if the cmd copy has an attribute 'flags' like "--from=" and keep the alias. Next, we need to check if this current alias is referenced in a previously defined FROM alias ( through attribute Values of cmd from)

Closes #1139


assets/queries/dockerfile/copy_from_without_from_alias_defined_previously/query.rego
```diff
@@ -0,0 +1,31 @@
...
+isPreviousAlias(startLine, currentAlias) = allow {
+  resource := input.document[i].command[name][_]
+  resource.Cmd == "from"
+  previousAlias := resource.Value[2]
+  
+  previousAlias == currentAlias
+  resource.EndLine < startLine
+  
+  allow = true
+}
```
joaoReigota1, line 31: COPY command with the flag '--from=0' shouldn't fail.
The COPY '--from=0' line copies just the built artifact from the previous stage into this new stage. The Go SDK and any intermediate artifacts are left behind, and not saved in the final image.

assets/queries/dockerfile/copy_from_without_from_alias_defined_previously/test/negative.dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM golang:1.7.3 AS builder
+WORKDIR /go/src/github.com/alexellis/href-counter/
+RUN go get -d -v golang.org/x/net/html
+COPY app.go    .
+RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o app .
+
+---
```
joaoReigota1, line 7: remove '---' doesn't exist in docker

assets/queries/dockerfile/copy_from_without_from_alias_defined_previously/test/positive.dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM golang:1.7.3 AS builder
+WORKDIR /go/src/github.com/alexellis/href-counter/
+RUN go get -d -v golang.org/x/net/html
+COPY app.go    .
+RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o app .
+
+---
```
joaoReigota1, line 7: same as before

assets/queries/dockerfile/copy_from_without_from_alias_defined_previously/metadata.json
```diff
@@ -0,0 +1,8 @@
+{
+  "id": "copy_from_without_from_alias_defined_previously",
+  "queryName": "COPY --from Without FROM Alias Defined Previously",
```
joaoReigota1, line 3: change to "COPY '--from' Without FROM Alias Defined Previously"

assets/queries/dockerfile/copy_from_without_from_alias_defined_previously/metadata.json
```diff
@@ -0,0 +1,8 @@
...
+  "severity": "MEDIUM",
+  "category": "Multi-stage",
+  "descriptionText": "COPY --from should mention a previously defined FROM alias",
```
joaoReigota1, line 6: change to "COPY command with the flag '--from' should mention a previously defined FROM alias"

joaoReigota1 (changes requested):

joaoReigota1 (commented):
Approved

joelsou5a (approved):

Issue #1162, opened 2020-12-21
Add CloudFormation Specifying Credentials Not Safe CloudFormation


pedro-mimoso:
### Platform
CloudFormation

### Provider
AWS

### Description
Specifying credentials in the template itself is probably not safe to do.

PR #1163, opened 2020-12-21
Add CloudFormation Specifying Credentials Not Safe CloudFormation #1162

pedro-mimoso:
Specifying credentials in the template itself is probably not safe to do. #1162


joaoReigota1 (commented):
Approve

ruigomescx (approved):

PR #1177, opened 2020-12-21
Update Sns Topic Should Specify KmsMasterKeyId Attribute query for CloudFormation #1171

pedro-mimoso:
KmsMasterKeyId attribute should not be undefined #1171


joaoReigota1 (changes requested):
change query name to "SNS Topic Should Specify KmsMasterKeyId Attribute"
change the category to "Encryption and Key Management"

joaoReigota1 (approved):

PR #1178, opened 2020-12-21
Add copy_from_references_current_from_alias query Closes #1172

### LINKED ISSUE #1172: Add copy_from_references_current_from_alias query for Docker
### Platform
Docker

### Description
Add a new "COPY --from References Current FROM Alias" query, that checks if COPY --from references the current FROM alias

rafaela-soares:
For this query, it is necessary to check if COPY --from references the current FROM alias. To do that, we need to check if the cmd copy has an attribute 'flags' like "--from=" and keep the alias. Next, we need to check if this current alias is equal to the current FROM alias ( through attribute Values of cmd from)

Closes #1172


joaoReigota1 (approved):
Approved

PR #1233, opened 2020-12-21
Added New Query RouteTable with default route

### LINKED ISSUE #1231: Add RouteTable with default route query for AWS Cloud Formation
### Platform
AWS Cloud Formation

### Provider
AWS

### Description
The use of NAT gateways is recommended, and not the default route which permits all traffic, in Route Tables.

cataraujo190:
### Description
The use of NAT gateways is recommended, and not the default route which permits all traffic, in Route Tables.

Close #1231


joaoReigota1 (approved):
Approved

PR #1249, opened 2020-12-21
Implemented query AWS Config Rule For Encrypted Volumes Is Disabled #…

### LINKED ISSUE #1247: Add AWS Config Rule For Encrypted Volumes Is Disabled query for Ansible
### Platform
Ansible

### Description
Check if AWS config rules do not identify Encrypted Volumes as a source.

samuel-ferreira:
Implemented query AWS Config Rule For Encrypted Volumes Is Disabled

Closes #1247


assets/queries/ansible/aws/aws_config_rule_for_encrypted_vols_is_disabled/query.rego
```diff
@@ -0,0 +1,28 @@
+package Cx
+
+CxPolicy [ result ] {
+  document := input.document[i]
+  tasks := getTasks(document)
+  task := tasks[t]
+  configRule := task["community.aws.aws_config_rule"]
+  configRuleName := task.name
+
+  upper(configRule.source.identifier) == "ENCRYPTED_VOLUMES"
```
jorge-cruz, line 10: I think what this query should check is if there is no config rule identifying Encrypted Volumes as a source. 
There must at least one config rule identifying Encrypted Volumes as a source.

jorge-cruz (changes requested):

assets/queries/ansible/aws/aws_config_rule_for_encrypted_vols_is_disabled/query.rego
samuel-ferreira: You're right. I updated the query and the test samples

jorge-cruz: The query will return a result for each "community.aws.aws_config_rule" with `configRule.source.identifier != "ENCRYPTED_VOLUMES"`. Try to get all tasks that are "community.aws.aws_config_rule" tasks and check if none of them has `configRule.source.identifier == "ENCRYPTED_VOLUMES"`. That way, the query will return only one result. I think the searchKey could point to the first "community.aws.aws_config_rule" task and, also, the issueType should be "MissingAttribute". Check the Terraform's version of this query if you have any questions

samuel-ferreira: I think it's ok now. Thanks for the explanation :)

jorge-cruz (approved):

PR #1253, opened 2020-12-21
Add missing_flag_from_dnf_install query for Docker

### LINKED ISSUE #1043: Add add_flag_to_dnf_install query for Docker
### Platform
Docker

### Provider
Dockerfile

### Description
The `-y` or `--assumeyes` flag should be added when invoking `dnf install`. If omitted, it can cause the command to fail during the build process, because dnf would expect manual input.

mcarvalhox:
This query checks if the `-y` or `--assume-yes` flag are added when invoking `dnf install`. If omitted, it can cause the command to fail during the build process, because `dnf` would expect manual input. Closes #1043


joaoReigota1 (approved):
Approved

joelsou5a (approved):

PR #1391, opened 2020-12-22
Add port22_in_public_scope Closes #1387

### LINKED ISSUE #1387: Add port22_in_public_scope query for Ansible
### Platform
Ansible

### Provider
AWS

### Description
Add new query "SSH' (TCP:22) in Public Scope" that checks if SSH' (TCP:22) is public since it is recommended that AWS Security Group should not allow Internet traffic to SSH port

rafaela-soares:
For this query, it is necessary to check if from_port and to_port are SSH and if cidr is not private 

Closes #1387

### DIFF
--- a/assets/queries/ansible/aws/port22_in_public_scope/query.rego
+++ b/assets/queries/ansible/aws/port22_in_public_scope/query.rego
@@ -44,7 +44,7 @@ CxPolicy[result] {
     }
 }
 
-getTasks(document) = result {
+getTasks(document) {
     result := [body | playbook := document.playbooks[0]; body := playbook.tasks]
     count(result) != 0
 } else = result {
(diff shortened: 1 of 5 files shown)


assets/queries/ansible/aws/port22_in_public_scope/query.rego
```diff
@@ -0,0 +1,81 @@
...
+} else = result {
+    result := [body | playbook := document.playbooks[_]; body := playbook ]
+    count(result) != 0
+}
+
+isSSH(currentFromPort, currentToPort) = allow {
+     currentFromPort <= 22
+     currentToPort >= 22
+     allow = true
+}
```
joaomartinscx, line 59: If the 'from' or 'to' ports are -1 means that all ports are considered, you should check those cases as well

joaomartinscx: also, if you do not indicate the result from the function, i.e, `= allow`, by default returns boolean and so returns true if all the statements inside are true, so you don't need to specify the return if its boolean

joaomartinscx (changes requested):

joaomartinscx (approved):
Everything looks good to me, good job!

Issue #1430, opened 2020-12-23
Add Service Does Not Target Pod query for Kubernetes


pedro-mimoso:
### Platform
Kubernetes

### Description
Service should Target a Pod

PR #1431, opened 2020-12-23
Add Service Does Not Target Pod query for Kubernetes #1430

pedro-mimoso:
Service should Target a Pod #1430


joaomartinscx (approved):
Everything looks okay, good job!

PR #1441, opened 2020-12-24
Add Azurerm Container Registry With No Locks query

### LINKED ISSUE #1440: Add Azurerm Container Registry With No Locks query for Ansible
### Platform
Ansible

### Provider
Azure

### Description
Azurerm Container Registry Must Contain Associated Locks

fabioGoncalvesCx:
Added new query for Ansible.

Provide a azure_rm_containerregistry resource and a resource to manage the azure_rm_lock

Closes #1440

### DIFF
--- a/assets/queries/ansible/azure/container_registry_with_no_locks/query.rego
+++ b/assets/queries/ansible/azure/container_registry_with_no_locks/query.rego
@@ -1,0 +1,25 @@
+package Cx
+
+CxPolicy [ result ] {
+	document := input.document[i]
+  	taskContainerRegistry := getTasks(document)[t]["azure_rm_containerregistry"]
+  	taskLock := getTasks(document)[t2]["azure_rm_lock"]
+  
+    taskContainerRegistry.resource_group != taskLock.resource_group
+    
+	result := {
+                "documentId": 		input.document[i].id,
+                "searchKey": 	    sprintf("name=%s.{{azure_rm_containerregistry}}.resource_group", [taskContainerRegistry.name]),
+                "issueType":		  "MissingAttribute",
+                "keyExpectedValue": "'azure_rm_containerregistry.resource_group' has lock associated",
+                "keyActualValue": "'azure_rm_containerregistry.resource_group' has no lock associated"
+              }
+}
+
+getTasks(document) = result {
+    result := [body | playbook := document.playbooks[0]; body := playbook.tasks]
+    count(result) != 0
+} else = result {
+    result := [body | playbook := document.playbooks[_]; body := playbook ]
+    count(result) != 0
+}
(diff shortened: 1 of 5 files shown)


assets/queries/ansible/azure/container_registry_with_no_locks/query.rego
```diff
@@ -0,0 +1,25 @@
+package Cx
+
+CxPolicy [ result ] {
+	document := input.document[i]
+  	taskContainerRegistry := getTasks(document)[t]["azure_rm_containerregistry"]
+  	taskLock := getTasks(document)[t2]["azure_rm_lock"]
+  
+    taskContainerRegistry.resource_group != taskLock.resource_group
```
ruigomescx, line 8: If the document has multiple resource groups, will not this always give a different result? because it will always find a different group. For example the container points to lock A and it exists lock A, B and C. This will prove always true, as B and C are different from A. But what we want to check is if A exists, right?

ruigomescx (changes requested):

assets/queries/ansible/azure/container_registry_with_no_locks/query.rego
```diff
@@ -0,0 +1,39 @@
+package Cx
+
+CxPolicy [ result ]  {
+	document := input.document[i]
+    tasks := getTasks(document)
+    task := tasks[t]
+
+  	taskContainerRegistry := task["azure_rm_containerregistry"]
+
+	not xpto(taskContainerRegistry)
```
ruigomescx, line 10: change function to relavant name

assets/queries/ansible/azure/container_registry_with_no_locks/query.rego
```diff
@@ -0,0 +1,39 @@
...
+
+
+		result := {
+                "documentId": 		input.document[i].id,
+                "searchKey": 	    sprintf("name=%s.{{azure_rm_containerregistry}}.resource_group", [task.name]),
+                "issueType":		  "MissingAttribute",
```
ruigomescx, line 16: change issueType to IncorrectValue since we have the a lock, but it's an incorrect one

assets/queries/ansible/azure/container_registry_with_no_locks/query.rego
```diff
@@ -0,0 +1,39 @@
...
+                "keyExpectedValue": "'azure_rm_containerregistry.resource_group' has lock associated",
+                "keyActualValue": "'azure_rm_containerregistry.resource_group' has no lock associated"
```
ruigomescx, line 18: change messages to explain that the lock used is referring to an one that doesn't exist

ruigomescx (changes requested):

ruigomescx (approved):

PR #1502, opened 2020-12-28
Add unprotected_s3_storage query Closes #1501

### LINKED ISSUE #1501: Add unprotected_s3_storage query for Ansible
### Platform
Ansible

### Provider
AWS

### Description
Add new query "unprotected_s3_storage", that check ifs AWS S3 Storage is protected with SSE (Server-Side Encryption)

rafaela-soares:
For this query, it is necessary to check the attribute 'encryption', since it "describes the default server-side encryption to apply to new objects in the bucket. In order to remove the server-side encryption, the encryption needs to be set to 'none' explicitly."

Closes #1501


assets/queries/ansible/aws/unprotected_s3_storage/query.rego
```diff
@@ -0,0 +1,43 @@
...
+
+    
+    result := {
+        "documentId":        document.id,
+        "searchKey":         sprintf("name=%s.{{amazon.aws.s3_bucket}}", [task.name]),
+        "issueType":         "MissingAttribute",
+        "keyExpectedValue":  "amazon.aws.s3_bucket.encryption is set",
+        "keyActualValue": 	 "amazon.aws.s3_bucket.encryption is undefined"
+    }
+}
```
ruigomescx, line 18: According to the documentation:

> Describes the default server-side encryption to apply to new objects in the bucket. In order to remove the server-side encryption, the encryption needs to be set to 'none' explicitly.

So only when we explicitly define as none, is when the encryption is disabled. 
We can cut this part from the query

ruigomescx (changes requested):

ruigomescx (approved):

PR #1539, opened 2020-12-28
Add S3 Bucket Allows Get Action From All Principals query for Ansible

### LINKED ISSUE #1538: Add S3 Bucket Allows Get Action From All Principals query for Ansible
### Platform
Ansible

### Provider
AWS

### Description
S3 Buckets must not allow Get Action From All Principals, as to prevent leaking private information to the entire internet or allow unauthorized data tampering / deletion.

Ruben-Silva:
Adding S3 Bucket Allows Get Action From All Principals query for Ansible, as to prevent leaking private information to the entire internet or allow unauthorized data tampering / deletion.

Closes #1538

### DIFF
--- a/assets/queries/ansible/aws/s3_bucket_allows_get_action_from_all_principals/query.rego
+++ b/assets/queries/ansible/aws/s3_bucket_allows_get_action_from_all_principals/query.rego
@@ -1,0 +1,28 @@
+package Cx
+
+CxPolicy [ result ] {
+  document := input.document[i]
+  tasks := getTasks(document)
+  task := tasks[t]
+  bucket := task["amazon.aws.aws_s3"]
+  bucketName := task.name
+  
+  contains(bucket.mode, "get")
+  contains(bucket.permission, "public")
+
+    result := {
+                "documentId":       input.document[i].id,
+                "searchKey":        sprintf("name={{%s}}.{{amazon.aws.aws_s3}}", [bucketName]),
+                "issueType":        "IncorrectValue",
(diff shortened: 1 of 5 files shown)


assets/queries/ansible/aws/s3_bucket_allows_get_action_from_all_principals/query.rego
```diff
@@ -0,0 +1,28 @@
...
+
+CxPolicy [ result ] {
+  document := input.document[i]
+  tasks := getTasks(document)
+  task := tasks[t]
+  bucket := task["amazon.aws.aws_s3"]
+  bucketName := task.name
+  
+  contains(bucket.mode, "get")
+  contains(bucket.permission, "public")
```
ruigomescx, line 11: We need to refactor this query. Looking at what the query name is suggesting, we are not checking the All Principals, but instead looking at the ACL permissions:
https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html

So we need to look at the policy, like the examples in terraform. I think the issue here, is that the module we're checking doesn't have the policy. Maybe we should be looking for this module for s3 buckets instead: https://docs.ansible.com/ansible/latest/collections/amazon/aws/s3_bucket_module.html#parameter-policy

Check PR for an example on how to check the policy: https://github.com/Checkmarx/kics/pull/1552/files

ruigomescx (changes requested):

ruigomescx (approved):

Issue #1593, opened 2020-12-29
Add Google Storage Bucket Level Access is Enabled query for Ansible


pedro-mimoso:
### Platform
Ansible

### Provider
GCP

### Description
Validates if the Google Storage Bucket Level Access is Enabled

PR #1594, opened 2020-12-29
Add Google Storage Bucket Level Access is Enabled query for Ansible #1593

pedro-mimoso:
Validates if the Google Storage Bucket Level Access is Enabled #1593


pedro-mimoso:
This query may be wrong because there is not a "uniform_bucket_level_access" variable in Ansible

joaoReigota1:
When you enable uniform bucket-level access on a bucket, Access Control Lists (ACLs) are disabled. 
In 'google.cloud.gcp_storage_bucket_access_control' -  The BucketAccessControls resource represents the Access Control Lists (ACLs) for buckets within Google Cloud Storage.
So I believe the query is indeed wrong since you are checking if the bucket has (ACLs) and not if the access is uniform.
My suggestion is to look for another module where you can ensure that  access on a bucket is uniform

PR #1682, opened 2021-01-04
Add RDS Open To Large Scope query for AWS CloudFormation

### LINKED ISSUE #1117: Add RDS Open to Large Scope query for AWS CloudFormation
### Platform
AWS CloudFormation

### Provider
AWS

### Description
Amazon RDS must not be open to a large scope, which means the IP Address in its Security Group must not have more than 256 hosts.

Ruben-Silva:
Adding RDS Open To Large Scope query for AWS CloudFormation, that checks if the IP Address in its Security Group has more than 256 hosts.

Closes #1117

### DIFF
--- a/assets/queries/cloudFormation/rds_open_to_large_scope/query.rego
+++ b/assets/queries/cloudFormation/rds_open_to_large_scope/query.rego
@@ -1,0 +1,66 @@
+package Cx
+
+CxPolicy [ result ] {
+	some i
+        not input.document[i].Resources[name].Type == "AWS::RDS::DBSecurityGroup"
+        resource := input.document[i].Resources[name]
+        resource.Type == "AWS::EC2::SecurityGroup"
+        object.get(resource.Properties.SecurityGroupIngress, "CidrIpv6", "undefined") == "undefined"
+        ipv4 := resource.Properties.SecurityGroupIngress.CidrIp
+        not check_mask_ipv4(ipv4)
+
+        result := {
+                    "documentId": 		input.document[i].id,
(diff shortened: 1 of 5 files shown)


assets/queries/cloudFormation/rds_open_to_large_scope/query.rego
```diff
@@ -0,0 +1,66 @@
+package Cx
+
+CxPolicy [ result ] {
+	some i
+        not input.document[i].Resources[name].Type == "AWS::RDS::DBSecurityGroup"
```
joaoReigota1, line 5: Why are you checking "not input.document[i].Resources[name].Type == "AWS::RDS::DBSecurityGroup" if bellow you check if resource.Type == "AWS::EC2::SecurityGroup"

joaoReigota1 (changes requested):

assets/queries/cloudFormation/rds_open_to_large_scope/query.rego
Ruben-Silva: According to the documentation, when a DBSecurityGroup is defined, there must not be a VPCSecurityGroup, so that was how I was checking. But now that I think about it, checking that is not the goal of the query, so I am going to remove that line.

Ruben-Silva: According to the documentation, when a DBSecurityGroup is defined, there must not be a VPCSecurityGroup, so that was how I was checking. But now that I think about it, checking that is not the goal of the query, so I am going to remove that line.

joaoReigota1 (approved):

PR #1688, opened 2021-01-04
Add lambda_without_xray_tracing query Closes #1686

### LINKED ISSUE #1686: Add lambda_without_xray_tracing query for Ansible
### Platform
Ansible

### Provider
AWS

### Description
Add new query "Lambda Without X-Ray Tracing", that checks if Lambda has X-Ray tracing enabled

rafaela-soares:
According to the documentation about attribute 'tracing_mode', "set mode to 'Active' to sample and trace incoming requests with AWS X-Ray. Turned off (set to 'PassThrough') by default".

So, for this query, it is necessary to check if attribute 'tracing_mode' exists. Furthermore, if when exits, is not set to 'Active' 

Closes #1686


Ruben-Silva (approved):
Looks good. Great work!

PR #1692, opened 2021-01-04
Add S3 Bucket Allows WriteACP Action From All Principals query for Ansible

### LINKED ISSUE #1691: Add S3 Bucket Allows Write_ACP Action From All Principals query for Ansible
### Platform
Ansible

### Provider
AWS

### Description
S3 Buckets must not allow Write_ACP Action From All Principals, as to prevent leaking private information to the entire internet or allow unauthorized data tampering / deletion. This means the 'Effect' must not be 'Allow' when the 'Action' is Write_ACP, for all Principals.

Ruben-Silva:
Adding S3 Bucket Allows WriteACP Action From All Principals query for Ansible, that checks if the 'Effect' is'Allow' when the 'Action' is Write_ACP, for all Principals.

Closes #1691

### DIFF
--- a/assets/queries/ansible/aws/s3_bucket_allows_writeACP_action_from_all_principals/query.rego
+++ b/assets/queries/ansible/aws/s3_bucket_allows_writeACP_action_from_all_principals/query.rego
@@ -1,0 +1,30 @@
+package Cx
+
+CxPolicy [ result ] {
+  document := input.document[i]
+  tasks := getTasks(document)
+  task := tasks[t]
+  bucket := task["amazon.aws.s3_bucket"]
+  bucketName := task.name
+
+  bucket.policy.Statement[_].Effect == "Allow"
+  contains(lower(bucket.policy.Statement[_].Action), "write")
+  contains(lower(bucket.policy.Statement[_].Action), "acp")
+  bucket.policy.Statement[_].Principal == "*"
+
+    result := {
+                "documentId":       input.document[i].id,
+                "searchKey":        sprintf("name={{%s}}.{{amazon.aws.s3_bucket}}.policy.Statement", [bucketName]),
+                "issueType":        "IncorrectValue",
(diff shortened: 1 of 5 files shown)


assets/queries/ansible/aws/s3_bucket_allows_writeACP_action_from_all_principals/query.rego
```diff
@@ -0,0 +1,30 @@
...
+  contains(lower(bucket.policy.Statement[_].Action), "write")
+  contains(lower(bucket.policy.Statement[_].Action), "acp")
+  bucket.policy.Statement[_].Principal == "*"
+
+    result := {
+                "documentId":       input.document[i].id,
+                "searchKey":        sprintf("name={{%s}}.{{amazon.aws.s3_bucket}}.policy.Statement", [bucketName]),
+                "issueType":        "IncorrectValue",
+                "keyExpectedValue": sprintf("amazon.aws.s3_bucket[%s] does not allow WriteACP Action From All Principals", [bucketName]),
+                "keyActualValue":   sprintf("amazon.aws.s3_bucket[%s] allows WriteACP Action From All Principals", [bucketName])
```
joaomartinscx, line 20: the bucketName refers to the task name, not the name of the resource. For example, in the sample negative.yaml, bucketName refers to 'Bucket' and here you are targeting the resource name, which in that case is 'mys3bucket'.

joaomartinscx (changes requested):

assets/queries/ansible/aws/s3_bucket_allows_writeACP_action_from_all_principals/query.rego
Ruben-Silva: Mistake corrected.

joaomartinscx (approved):
Everything looks good to me, good job!

PR #1695, opened 2021-01-04
Add ECS Task Definition W/Generic Name query #1694

### LINKED ISSUE #1694: Add ECS Task Definition With Generic Container Name query for CloudFormation
### Platform
CloudFormation

### Provider
AWS

### Description
Amazon ECS should have Container Name with the name that we need to check

joelcarvalhocheckmarx:
Amazon ECS should have Container Name with the name that we need to check

Closes #1694


ruigomescx:
Query not complying with static analysis

PR #1701, opened 2021-01-04
Add ecr_image_tag_mutable query Closes #1700

### LINKED ISSUE #1700: Add ecr_image_tag_mutable query for Ansible
### Platform
Ansible

### Provider
AWS

### Description
Add new query "ECR Image Tag Mutable", that checks if ECR has an image tag not mutable

rafaela-soares:
For this query, it is necessary to check if the attribute 'image_tag_mutability' is not set to 'immutable'

Closes #1700


Ruben-Silva (approved):
Looks good. Great work!

PR #1763, opened 2021-01-07
added cmd flag generate-id #670

### LINKED ISSUE #670: Queries queryId should be changed to UID
### Description
Currently we are using the query name has it's own ID. We should change to support a UID instead

ruigomescx:
Added flag generate-id (-g for short) in order to make possible to generate a UUID for the query creation.

Closes #670

### DIFF
--- a/cmd/console/main.go
+++ b/cmd/console/main.go
@@ -22,6 +22,7 @@ import (
 	yamlParser "github.com/Checkmarx/kics/pkg/parser/yaml"
 	"github.com/Checkmarx/kics/pkg/source"
 	"github.com/getsentry/sentry-go"
+	"github.com/google/uuid"
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


cmd/console/main.go
```diff
@@ -62,11 +64,15 @@ func main() { // nolint:funlen,gocyclo
 		RunE: func(cmd *cobra.Command, args []string) error {
 			defer sentry.Flush(timeMult * time.Second)
 
-			store := storage.NewMemoryStorage()
 			if verbose {
 				consoleLogger = zerolog.ConsoleWriter{Out: os.Stdout}
 			}
 
+			if generateID {
+				_, err = fmt.Printf(uuid.New().String())
```
joaoReigota1, line 72: change to 'fmt.Println(uuid.New().String())' to prevent '%' at end of line
and add 'os.Exit(0)' so that  stops execution

ruigomescx: added suggestions

joaoReigota1 (approved):

PR #1843, opened 2021-01-25
Add nightly release and remove pre release workflow

elit-cx:
Closes #

**Proposed Changes**

-
-
-


kaplanlior:
Would this create new releases on github? We want to avoid both tagging and releasing (in github) on a nightly basis.
Can we have the binaries saved in another location?

markmishaevcx (approved):

kaplanlior:
Congrats on the first nightly build (:

I would use the short sha for it instead of the full one:
git rev-parse --short <sha>

PR #1878, opened 2021-01-28
Added --config flag to kics #1833

### LINKED ISSUE #1833: Add option to pass .json config file to kics
### Description
Currently we support some flags that the user can use. Since we're planning to expand on the number of flags we provide, it makes sense to create a .json configuration file that could contain all the flags we have. 

We would then pass it to kics as this:
`kics scan --config "path to config file"`

Configuration .json could look like this:
```
{
   "log-file": true, // outputs log file
   "path": "src/folder", // ...
   "queriesPath": "assets/queries" // ...
   ...
}
```

joaoReigota1:
Closes #1833

**Proposed Changes**

- Added viper integration with cobra for kics to be able to load config from file with '--config'  flag

### DIFF
--- a/internal/console/scan.go
+++ b/internal/console/scan.go
@@ -5,6 +5,8 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/Checkmarx/kics/internal/storage"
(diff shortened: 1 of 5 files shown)


internal/console/scan.go
```diff
@@ -22,18 +24,57 @@ import (
...
+	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+		return initializeConfig(cmd)
+	},
 	RunE: func(cmd *cobra.Command, args []string) error {
 		return scan()
 	},
 }
 
+func initializeConfig(cmd *cobra.Command) error {
+	v := viper.New()
```
felipe-avelar, line 43: I think we should check if `cfgFile` has any value to execute initialize and, if has a value, but the path is wrong should return an exiting error.

joaoReigota1: resolved

internal/console/scan.go
```diff
@@ -22,18 +24,63 @@ import (
...
+func initializeConfig(cmd *cobra.Command) error {
+	if _, err := os.Stat(cfgFile); os.IsNotExist(err) {
+		return err
+	}
+	v := viper.New()
+	base := filepath.Base(cfgFile)
+	v.SetConfigName(strings.TrimSuffix(base, "."))
+	v.AddConfigPath(filepath.Dir(cfgFile))
+	if err := v.ReadInConfig(); err != nil {
+		if _, ok := err.(viper.ConfigFileNotFoundError); !ok {
```
felipe-avelar, line 54: This is a double verification if configuration file exists?

rogeriopeixotocx (dismissed):
LGTM

felipe-avelar (approved):
LGTM

PR #1892, opened 2021-01-29
Update VPC Attached With Too Many Gateways query for CloudFormation #1891

### LINKED ISSUE #1891: Update VPC Attached With Too Many Gateways query for CloudFormation
### Platform
CloudFormation

### Description
Was pointing to files that should not be pointing, and to the negative sample of the same query.

pedro-mimoso:
Closes #1891 

**Proposed Changes**

- changed the names of the Resources, since when testing it joins the documents if the Resource has the same name the query would be counting all as one.

### DIFF
--- a/assets/queries/cloudFormation/vpc_attached_with_too_many_gateways/test/negative.yaml
+++ b/assets/queries/cloudFormation/vpc_attached_with_too_many_gateways/test/negative.yaml
@@ -1,6 +1,6 @@
 AWSTemplateFormatVersion: '2010-09-09'
 Resources:
-    myVP_2:
+    myVPC_2:
       Type: AWS::EC2::VPC
       Properties:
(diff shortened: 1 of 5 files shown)


assets/queries/cloudFormation/vpc_attached_with_too_many_gateways/test/negative.yaml
```diff
@@ -1,6 +1,6 @@
 AWSTemplateFormatVersion: '2010-09-09'
 Resources:
-    myVPC:
+    myVP_2:
```
ruigomescx, line 3: fix to myVP**C**_2

ruigomescx (changes requested):

ruigomescx (approved):

PR #1895, opened 2021-01-29
Enable secure cross-repo PR testing with pull_request_target workflow trigger event

### LINKED ISSUE #1894: Enable secure cross-repo PR testing with pull_request_target workflow trigger event
### Description
To enable us to safely run CI jobs in cross-repo PR's we should leverage the new GitHub Actions event type: `pull_request_target`

> https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request_target
> This event is similar to pull_request, except that it runs in the context of the base repository of the pull request, rather than in the merge commit. This means that you can more safely make your secrets available to the workflows triggered by the pull request, because only workflows defined in the commit on the base repository are run. For example, this event allows you to create workflows that label and comment on pull requests, based on the contents of the event payload.

> In order to solve this [security issues with pull_request], we’ve added a new pull_request_target event, which behaves in an almost identical way to the pull_request event with the same set of filters and payload. However, instead of running against the workflow and code from the merge commit, the event runs against the workflow and code from the base of the pull request. This means the workflow is running from a trusted source and is given access to a read/write token as well as secrets enabling the maintainer to safely comment on or label a pull request.

After taking a look at the following securitylab.github [blog post](https://securitylab.github.com/research/github-actions-preventing-pwn-requests) we decided to try the solution in the provided example:

```
    on:
      pull_request_target:
        types: [labeled]

    jobs:
      build:
        name: Build and test
        runs-on: ubuntu-latest
        if: contains(github.event.pull_request.labels.*.name, 'safe to test')
```

rogeriopeixotocx:
Closes #1894

**Proposed Changes**

- replacing `pull_request` with `pull_request_target` workflow trigger.
- adding if conditional checking for label 'safe to test' whenever there are secrets in a job.
- fixing other workflows that were named incorrectly.
- check the linked issue for more details.

### DIFF
--- a/.github/workflows/go-ci-cross-repo.yml
+++ b/.github/workflows/go-ci-cross-repo.yml
@@ -1,12 +1,14 @@
-name: go CI
+name: go CI cross-repo
 
 on:
-  pull_request:
-    branches: [ master ]
+  pull_request_target:
+    branches: [master]
+    types: [labeled]
 
 jobs:
   lint:
     name: lint
+    if: contains(github.event.pull_request.labels.*.name, 'safe to test')
     runs-on: ubuntu-latest
     steps:
     - name: Cache dependencies
@@ -26,6 +28,7 @@ jobs:
         version: v1.29
   unit-tests:
     name: unit-tests
+    if: contains(github.event.pull_request.labels.*.name, 'safe to test')
     strategy:
       matrix:
         go-version: [1.15.x]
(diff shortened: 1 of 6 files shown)


rogeriopeixotocx:
this looks broken I might need a hand to configure this cc: @elit-cx

markmishaevcx (dismissed):

rogeriopeixotocx:
Going to create another PR for this issue. Based on https://github.com/Checkmarx/kics/pull/1920

PR #1900, opened 2021-01-29
Update queries for Kubernetes

### LINKED ISSUE #1890: Update queries for Kubernetes
### Platform
Kubernetes

### Description
- Some queries have typos that can affect the vulnerability detection, such as "Pode" instead of "Pod";
- Queries should be updated to use the new generic functions from the libraries provided;

mcarvalhox:
Closes #1890

**Proposed Changes**

- Typo corrected: `Pode` changed to `Pod` in : 'Privilege Escalation Allowed', 'Allowed Capabilities', 'Allowed Capabilities Sys Admin'
- Added import for generic function `getSpecInfo` 
- Added missing search case for `initContainers` in some queries, which resulted in more vulnerabilities in other samples being detected


assets/queries/k8s/tiller_deployment_is_accessible_within_cluster/query.rego
```diff
@@ -70,18 +73,19 @@ CxPolicy[result] {
 
 	isTillerTemplate(document)
```
joaoReigota1, line 74: the method checkmetadata() is used in the query 'tiller_is_deployed' move it to the libraries and import it to prevent code duplication

assets/queries/k8s/cpu_requests/query.rego
```diff
@@ -1,67 +1,60 @@
...
 
 CxPolicy[result] {
 	document := input.document[i]
-	specInfo := getSpecInfo(document)
+	specInfo := k8sLib.getSpecInfo(document)
 	metadata := document.metadata
 
-	containers := document.spec.containers
+	types := {"initContainers", "containers"}
+	containers := document.spec[types[x]]
```
joaoReigota1, line 30: this is also used in a lot of queries, can pass to the libraries as well

joaoReigota1 (changes requested):

assets/queries/k8s/cpu_requests/query.rego
mcarvalhox: this function would have to receive the document as input. If a document (or sample) has both containers and initContainers,  calling the function would produce the error: **eval_conflict_error: functions must not produce multiple outputs for same inputs**

rogeriopeixotocx:
linter issues fixed @joaoReigota1 @ruigomescx

joaoReigota1 (approved):

PR #1933, opened 2021-02-01
Fixed bugs in contains functions and searchKey in Dockerfile queries

fabioGoncalvesCx:
Closes #1931

**Proposed Changes**

-Missing Dnf Clean All, Missing Flag From Dnf Install, Pip install Keeping Cached Packages, Run Utilities And POSIX Commands, Unpinned Package Version in Pip Install queries fixed. Contains functions and some search Key were wrong.

PR #1951, opened 2021-02-02
Update 'contains' parameters in some Docker queries Closes #1915

### LINKED ISSUE #1915: Update 'contains' parameters in some Docker queries
### Platform
Docker

### Description
The parameters referenced as input in 'contains' function could be a false positive, in some queries. For example, the query "Update Instruction Alone" (9bae49be-0aa3-4de5-bab2-4c3a069e40cd) results in false positives, since verify if "update" exists. In cases as "update-ca-trust", this query will return a false positive.

rafaela-soares:
Closes #1915 

This PR is similar to this one: https://github.com/Checkmarx/kics/pull/1933. In the queries mentioned below, contains function catch false positives since this function searches incorrectly.

**Proposed Changes**

- **Query "Update Instruction Alone" (9bae49be-0aa3-4de5-bab2-4c3a069e40cd)**: Instead of searching for "update", should search for " update " or cases when update is the last word of the command. This way, the contains function does not catch cases like "update-ca-trust" or "https://github.com/Collection/updateLater".

- **Query "Yum Clean All Missing" (00481784-25aa-4a55-8633-3136dfcf4f37)** : Instead of searching for "yum" or "yum install", should search for patterns where yum install could have flags, like "yum -y install". So, a better way is to use a regex function (regex.match("yum (-[a-zA-Z]+ *)*install", command)). This way, this query does not result in false positives like "yum -y install \ yum clean all" or "https://github.com/Collection/yum".


joaoReigota1 (changes requested):
Remove result.json file

rafaela-soares:
> Remove result.json file

Sorry

joaoReigota1 (approved):

PR #1958, opened 2021-02-02
Delete no_password_policy_enabled query for Ansible #1955

### LINKED ISSUE #1955: Update no_password_policy_enabled query for Ansible
### Platform
Ansible

### Provider
AWS

### Description
This query is returning a result for every task if there isn't a `iam_password_policy` task

jorge-cruz:
Closes #1955 

**Proposed Changes**

- Delete no_password_policy_enabled query

The Terraform's version of this query checks the `aws_iam_user_login_profile` resource which has no matching module in Ansible. What this query is doing is testing the presence of a `iam_password_policy` task (if there is none it will return a result for every task) and checking if the `require_lowercase`, `require_uppercase` and `require_numbers` fields are defined with the value `true`. These fields are already checked in the queries:
- `IAM Password Without Lowercase Letter`
- `IAM Password Without Uppercase Letter`
- `IAM Password Without Numbers`


joaomartinscx (commented):
Looks good to me, once query removal is evaluated, I propose the merge of this PR

ruigomescx (approved):

ruigomescx (approved):

PR #1960, opened 2021-02-02
Update samples and rules in Kubernetes queries

### LINKED ISSUE #1957: Update samples and rules in 'StatefulSet Without PodDisruptionBudget', 'Deployment Without PodDisruptionBudget' and 'Service Does Not Target Port'
### Platform
Kubernetes

### Description

- Queries 'StatefulSet Without PodDisruptionBudget', 'Deployment Without PodDisruptionBudget' and 'Service Does Not Target Port' are not reporting a vulnerability in their own positive.yaml file.
Because the parser combines both positive and negative samples into one large file, some positive examples become "fixed" by the negative ones. In this case, a simple fix is to change the labels in the negative samples so that they are unique within those files.

- PodDisruptionBudgets shouldn't be used for single replicas, as the [kubectl drains might get stuck
](https://github.com/kubernetes/kubernetes/issues/48307)

- The searchKey should go deeper

mcarvalhox:
Closes #1957

**Proposed Changes**

StatefulSet Without PodDisruptionBudget' and 'Deployment Without PodDisruptionBudget' : 
- Added verification to ensure the number of replicas is higher than 1
- Added `spec.selector.matchLabels` to the searchKey, as it was not picking up the correct line in some samples
- The query was only comparing the matchLabels with key: app, now any key-value pair can be matched
- Changed the labels in negative.yaml, so that the PodDisruptionBudget only targets the workload from the same file

'Service Does Not Target Pod':
- The query was only comparing the labels under `selector` with key: app, now any key-value pair can be matched
- In positive.yaml, changed the value of `targetPort`, so it doesn't target the Pod sample in negative.yaml


assets/queries/k8s/deployment_without_pod_disruption_budget/query.rego
```diff
@@ -3,26 +3,28 @@ package Cx
...
-		"keyExpectedValue": sprintf("metadata.name=%s are targeted by a PDB", [metadata.name]),
-		"keyActualValue": sprintf("metadata.name=%s are not targeted by a PDB", [metadata.name]),
+		"keyExpectedValue": sprintf("metadata.name=%s is targeted by a PodDisruptionBudget", [metadata.name]),
+		"keyActualValue": sprintf("metadata.name=%s is not targeted by a PodDisruptionBudget", [metadata.name]),
 	}
 }
 
 CheckIFPdbExists(deployments) = result {
 	documents := input.document
 	pdbs := [pdb | documents[index].kind == "PodDisruptionBudget"; pdb = documents[index]]
```
joaoReigota1, line 22: Don't iterate through every document, the query should look for PodDisruptionBudget inside the same document to prevent timeouts

assets/queries/k8s/statefulset_without_pod_disruption_budget/query.rego
```diff
@@ -3,26 +3,28 @@ package Cx
...
-		"keyActualValue": sprintf("metadata.name=%s is not targeted by a PDB", [metadata.name]),
+		"keyExpectedValue": sprintf("metadata.name=%s is targeted by a PodDisruptionBudget", [metadata.name]),
+		"keyActualValue": sprintf("metadata.name=%s is not targeted by a PodDisruptionBudget", [metadata.name]),
 	}
 }
 
-CheckIFPdbExists(statefulsets) = result {
+CheckIFPdbExists(statefulset) = result {
 	documents := input.document
 	pdbs := [pdb | documents[index].kind == "PodDisruptionBudget"; pdb = documents[index]]
```
joaoReigota1, line 22: Don't iterate through every document, the query should look for PodDisruptionBudget inside the same document to prevent timeouts

assets/queries/k8s/deployment_without_pod_disruption_budget/query.rego
```diff
@@ -3,26 +3,28 @@ package Cx
...
-	result := contains(pdbs, deployments.spec.selector.matchLabels.app)
+	result := contains(pdbs, deployments.spec.selector.matchLabels)
+} else = false {
+	true
 }
 
-contains(array, string) {
-	array[a].spec.selector.matchLabels.app == string
+contains(array, label) {
+	array[a].spec.selector.matchLabels[_] == label[_]
```
joaoReigota1, line 29: CheckIFPdbExists is used in both 'deployment_without_pod_disruption_budget' and 'statefulset_without_pod_disruption_budget' to clear duplication pass this function to libraries/k8s/library.rego and import the function in both queries

joaoReigota1 (changes requested):

assets/queries/k8s/service_does_not_target_pod/test/positive.yaml
```diff
@@ -11,7 +11,7 @@ spec:
       nodePort: 30475
       port: 8089
       protocol: TCP
-      targetPort: 8089
+      targetPort: 8088
```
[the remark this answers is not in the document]
mcarvalhox: What's wrong with the indentation?

assets/libraries/k8s/library.rego
```diff
@@ -16,6 +16,18 @@ checkKind(currentKind, listKinds) {
 	currentKind == listKinds[i]
 }
 
+CheckIFPdbExists(statefulset) = result {
+	pdb := input.document[j]
+	pdb.kind == "PodDisruptionBudget"
+	result := containsLabel(pdb, statefulset.spec.selector.matchLabels)
+} else = false {
+	true
+}
```
ruigomescx, line 25: This function should not be in the generic libraries as it's too specific for a query. Also please change the name to a more appropriate name like "isPodDisruptionBudget"

ruigomescx (changes requested):

ruigomescx (approved):

PR #1997, opened 2021-02-08
Update logging_disabled_for_key_vault query Closes #1984

### LINKED ISSUE #1984: Update logging_disabled_for_key_vault query for Terraform
**Platform**
Terraform

**Provider**
Azure

**Description**
This query isn't giving the full list of expected positive results.

NunoAraujoCX:
Closes #1984 

**Proposed Changes**

- Fixed query not giving the full list of the expected positive results because the name of the resource and data was the same in this file positive and negative terraform files so the positive file was using values from the negative file


Ruben-Silva (approved):
Looks good!

PR #2007, opened 2021-02-08
Refactor Service With External Load Balance query #2006

### LINKED ISSUE #2006: Update Service With External Load Balance query for Kubernetes
### Platform
Kubernetes

### Description
Refactor Query
The query has many false positives, this is because the query never detects if 2 or more resources are defined.

joelcarvalhocheckmarx:
Closes #2006 

**Proposed Changes**

- Refactor Query
- The query has many false positives, this is because the query never detects if the Service using an external Load Balancer provider by cloud provider.

### DIFF
--- a/assets/queries/k8s/service_with_external_load_balance/query.rego
+++ b/assets/queries/k8s/service_with_external_load_balance/query.rego
@@ -1,39 +1,52 @@
 package Cx
 
 CxPolicy[result] {
-	service := input.document[i]
-	service.kind == "Service"
-	metadata := service.metadata
+	document := input.document[i]
+	object.get(document, "kind", "undefined") == "Service"
 
-    service.spec.type == "LoadBalancer"
-    service.spec.selector.app
+    metadata = document.metadata
+	document.spec.type == "LoadBalancer"
 
-	documents := input.document
-	pod := [pod | documents[index].spec.template.metadata.labels.app == service.spec.selector.app; pod = documents[index]]
-	count(pod) != 0
-	checkLabels(pod, service.spec.selector.app)
-	res := checkPorts(pod, service.spec.ports[k].targetPort)
-	res == true
+	object.get(metadata, "annotations", "undefined") == "undefined"
 
 	result := {
 		"documentId": input.document[i].id,
-		"searchKey": sprintf("metadata.name={{%s}}.spec", [metadata.name]),
+		"searchKey": sprintf("metadata.name={{%s}}", [metadata.name]),
 		"issueType": "MissingAttribute",
-		"keyExpectedValue": sprintf("metadata.name={{%s}} is not exposing a workload", [metadata.name]),
-		"keyActualValue": sprintf("metadata.name={{%s}} is exposing a workload", [metadata.name]),
+		"keyExpectedValue": "'metadata.annotations' is set",
+		"keyActualValue": "'metadata.annotations' is undefined",
 	}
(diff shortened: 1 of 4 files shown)


assets/queries/k8s/service_with_external_load_balance/query.rego
```diff
@@ -5,30 +5,35 @@ CxPolicy[result] {
...
-	not CheckIfDestinationPortExists(service)
+    service.spec.type == "LoadBalancer"
+    service.spec.selector.app
+
+	documents := input.document
+	pod := [pod | documents[index].spec.template.metadata.labels.app == service.spec.selector.app; pod = documents[index]]
+	count(pod) != 0
+	checkLabels(pod, service.spec.selector.app)
+	res := checkPorts(pod, service.spec.ports[k].targetPort)
+	res == true
```
felipe-avelar, line 16: Could only be `checkPorts(pod, service.spec.ports[k].targetPort)` and refactor to returns true or false

assets/queries/k8s/service_with_external_load_balance/query.rego
```diff
@@ -5,30 +5,35 @@ CxPolicy[result] {
...
-	result := contains(pod, service.spec.selector.app)
-	contains(pod, service.spec.ports[k].targetPort)
-}
 
-contains(array, string) {
+checkLabels(array, string) = labels {
 	array[a].spec.template.metadata.labels.app == string
+    labels := 1
+} else {
+	labels := 0
```
felipe-avelar, line 32: Shouldn't 1 be true and 0, false?

assets/queries/k8s/service_with_external_load_balance/query.rego
```diff
@@ -5,30 +5,35 @@ CxPolicy[result] {
...
 }
-
-contains(array, string) {
+checkPorts(array, string) = port {
 	array[a].spec.template.spec.containers[c].ports[p].containerPort == string
+    port := 1
+} else {
+	port := 0
```
felipe-avelar, line 38: Shouldn't 1 be true and 0, false?

felipe-avelar (changes requested):
Please, check suggested changes

assets/queries/k8s/service_with_external_load_balance/query.rego
```diff
@@ -1,34 +1,52 @@
...
+		"issueType": "IncorrectValue",
+		"keyExpectedValue": sprintf("metadata.name={{%s}} using an external Load Balancer provider by cloud provider", [metadata.name]),
+		"keyActualValue": sprintf("metadata.name={{%s}} is exposing a workload, not using an external Load Balancer provider by cloud provider", [metadata.name]),
+	}
+}
 
-	result := contains(pod, service.spec.selector.app)
-	contains(pod, service.spec.ports[k].targetPort)
+checkLoadBalancer(annotation) {
+    annotation["cloud.google.com/load-balancer-type"] == "Internal"
```
felipe-avelar, line 43: This is a check for older versions for GKE, as you can see [here](https://cloud.google.com/kubernetes-engine/docs/how-to/internal-load-balancing)

I think we should also check this:
`cloud.google.com/load-balancer-type: "Internal"`

felipe-avelar (changes requested):

felipe-avelar (approved):
LGTM

PR #2031, opened 2021-02-09
Fixed queries with wrong evaluations. Closes #2027

### LINKED ISSUE #2027: Update problematic queries for Terraform
### Platform
Terraform

### Description
Some queries have issues in the implementation that should be addressed.

| Flagged Queries | Issue |
|-------------------|-------|
| Cloud SQL DB Is Publicly Accessible | The query should be fixed in two points: remove "dynamic" from the path to access "authorized_networks" and change the public to check if the IP is "0.0.0.0/0" and not just "0.0.0.0". Basically, contains() instead of == |
| SSH' (TCP:22) in Public Scope | Sometimes SSH is allowed but CIDR blocks refer to a var of allowed_networks, seems rush to judge it as a public network. Also,  even though sometimes the results are correct and the lines are too, the search key should point to *cidr_blocks* and not *cidr* |
| Using Default Service Account | The email is not in plain text but stored in a variable, should not be caught in this query. |

joaomartinscx:
Closes #2027 

**Proposed Changes**

| Queries changed | Changes |
|---------------------|---------|
| Cloud SQL DB Is Publicly Accessible | Allow multiple authorized_networks rules and evaluate them separately. Fixed searchKey.  Changed samples and expected results to reflect the changes |
| SSH (TCP:22) in Public Scope | Allow multiple ingress rules and evaluate them separately. Allow values in variables since it's not possible to assess the risk. Fixed searchKey |
| Using Default Service Account |  Allow values in variables since it's not possible to assess the risk |

### DIFF
--- a/assets/queries/terraform/aws/port22_in_public_scope/query.rego
+++ b/assets/queries/terraform/aws/port22_in_public_scope/query.rego
@@ -11,16 +11,20 @@ isPrivate(cidr) = allow {
 
 	cidrLength := count(cidr)
 
-	count({x | cidr[x]; cidr[x] == privateIPs[j]}) == cidrLength
+	privLen := count({x | cidr[x]; cidr[x] == privateIPs[j]})
+    varLen := count({b | cidr[b]; contains(cidr[b],"$")}) # check if all CIDR blocks are stored in variables, which cannot be evaluated in the static scanning
 
+    privLen + varLen == cidrLength
+    
 	allow = true
 }
 
 CxPolicy[result] {
-	resource := input.document[i].resource.aws_security_group[name].ingress
-	currentFromPort := resource.from_port
-	currentToPort := resource.to_port
-	cidr := resource.cidr_blocks
+	resource := input.document[i].resource.aws_security_group[name]
+	ingress := getIngressList(resource.ingress)
+	currentFromPort := ingress[j].from_port
+	currentToPort := ingress[j].to_port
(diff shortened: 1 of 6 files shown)


assets/queries/terraform/aws/port22_in_public_scope/query.rego
```diff
@@ -14,23 +14,37 @@ isPrivate(cidr) = allow {
 	count({x | cidr[x]; cidr[x] == privateIPs[j]}) == cidrLength
 
 	allow = true
+} else = blocksInVars(cidr)
+
+blocksInVars(cidr) { # check if all CIDR blocks are stored in variables, which cannot be evaluated in the static scanning
+    len := count(cidr)
+    count({b | cidr[b]; contains(cidr[b],"$")}) == len
```
jorge-cruz, line 21: I think the query will return a result if the ingress has a list of cidr_blocks with private IP's and variables (e.g. `["192.120.0.0/16", "75.132.0.0/16", "${var}"]`).

A better approach could be checking if some CIDR block is neither a private IP nor a variable.

jorge-cruz (changes requested):

assets/queries/terraform/aws/port22_in_public_scope/query.rego
joaomartinscx: Attended request

jorge-cruz (approved):

ruigomescx (approved):

PR #2094, opened 2021-02-18
Update IAM User With No Group query for AWS CloudFormation 

### LINKED ISSUE #2093: Update IAM User With No Group query for AWS CloudFormation
### Platform
AWS CloudFormation

### Provider
AWS

### Description
Changing query to use "count()" to check if an array is empty and adding query for when "groups" is null.

Ruben-Silva:
Closes #2093

**Proposed Changes**

- Updating query to use "count()" when checking if attribute "Groups" is empty;
- Adding query for when attribute "Groups" is null.


assets/queries/cloudFormation/user_missing_group/query.rego
```diff
@@ -27,3 +27,17 @@ CxPolicy[result] {
 		"keyActualValue": "'Resources.Properties.Groups' is empty",
 	}
 }
+
+CxPolicy[result] {
```
rogeriopeixotocx, line 31: if `Groups` is defined it can never be `null`. Either an empty or filled list of strings. https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-iam-user.html#cfn-iam-user-groups

This policy is not necessary.

assets/queries/cloudFormation/user_missing_group/test/positive.yaml
```diff
@@ -30,6 +30,35 @@ AWSTemplateFormatVersion: "2010-09-09"
 Description: A sample template
```
rogeriopeixotocx, line 30: Third sample is invalid due to Groups null value in line 64:

`[cfn-lint] E0000: Null value at line 7 column 16`

rogeriopeixotocx (changes requested):

assets/queries/cloudFormation/user_missing_group/query.rego
Ruben-Silva: Understood. Removed that validation.

assets/queries/cloudFormation/user_missing_group/test/positive.yaml
Ruben-Silva: Understood. Removed that sample.

rogeriopeixotocx (approved):
LGTM

PR #2097, opened 2021-02-18
fixed case sensitive issues with type flag and fixed bug with dup types

### LINKED ISSUE #2092: Type flag shouldn't be case sensitive
### Expected Behavior

--type ansible works

### Actual Behavior

--type ansible doesn't work

joaoReigota1:
fixed case sensitive issues with type flag and fixed bug with dup types #2092

Closes #2092

**Proposed Changes**

- type flag now ignores case sensitive
-fixed bug where if two extensions add the same type it would appear duplicated invalid arguments


rogeriopeixotocx (approved):
LGTM

PR #2105, opened 2021-02-19
Enhanced config flag #2077

### LINKED ISSUE #2077: KICS - Configuration as Code
### Description
As a user, I'd like to be able to automate my KICS scans without the need for passing flags into the command line, but rather having it in the source code of my project in a specific file

**Use case scenario**
1. I create a JSON file named **kics.config** in the root of my project
2. In this file I write all the configuration needed for scanning my project with KICS (e.g., files to exclude, technology to scan)
3. KICS starts scan without flags (besides the relevant ones like the -p for instance)
4. KICS looks for kics.config file and consider its configurations
5. Scan runs as if we passed this config via cli flags

joaoReigota1:
Closes #2077

**Proposed Changes**

- Kics Can now be called without flags and will get the configurations from the 'kics.config' file inside the root of kics
-Kics will now check the contents of the config file and determine its extension 
-If there is no 'kics.config' file in the root of kics, kicks must be run with flags, or with the '--config' flag with the path to a configuration file 

### DIFF
--- a/internal/console/helpers/helpers.go
+++ b/internal/console/helpers/helpers.go
@@ -2,6 +2,7 @@ package helpers
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -11,6 +12,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/BurntSushi/toml"
 	"github.com/Checkmarx/kics/pkg/model"
 	"github.com/hashicorp/hcl"
 	"github.com/rs/zerolog"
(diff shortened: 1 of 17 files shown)


nunoocx:
There are 3 Low results in SAST scan. It seems that they are from your PR. Please Check https://sast.checkmarx.net/cxwebclient/ViewerMain.aspx?scanId=1052440&ProjectID=1199
image

internal/console/helpers/helpers.go
```diff
@@ -176,3 +178,26 @@ func CustomConsoleWriter(fileLogger *zerolog.ConsoleWriter) zerolog.ConsoleWrite
...
+		return "yaml"
+	}
+
+	if c, err := hcl.Parse(string(rc)); err == nil {
+		if err = hcl.DecodeObject(&temp, c); err == nil {
+			return "hcl"
+		}
+	}
+
+	return "toml"
```
nunoocx, line 202: What if I write something that is neither of the above and is not toml as well?
Maybe in that case we should report an error?

internal/console/helpers/helpers_test.go
```diff
@@ -219,3 +220,64 @@ func TestProgressBar(t *testing.T) {
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := FileAnalyzer(tt.arg); !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("FileAnalyzer() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
```
nunoocx, line 283: Need to add a negative test: what if the content is neither of the supported formats? What is the behaviour?

nunoocx (changes requested):
Check my comments ;)

nunoocx (approved):
LGTM

ruigomescx (approved):

PR #2108, opened 2021-02-19
Update EC2 Network ACL Duplicate Rule query for AWS CloudFormation 

### LINKED ISSUE #2107: Update EC2 Network ACL Duplicate Rule query for AWS CloudFormation
### Platform
AWS CloudFormation

### Provider
AWS

### Description
Updating "getTraffic()" method to consider more formats in which a value can be.

Ruben-Silva:
Closes #2107

**Proposed Changes**

- Updating "getTraffic()" method to consider more formats in which a value can be.

### DIFF
--- a/assets/queries/cloudFormation/ec2_network_acl_duplicate_rule/query.rego
+++ b/assets/queries/cloudFormation/ec2_network_acl_duplicate_rule/query.rego
@@ -34,9 +34,21 @@ getRef(obj) = obj.Ref {
 }
 
 getTraffic(entry) = "egress" {
-	object.get(entry.Properties, "Egress", "undefined") == true
+	entry.Properties.Egress == true
+} else = "egress" {
+	lower(entry.Properties.Egress) == "true"
 } else = "ingress" {
-	true
+	lower(entry.Properties.Egress) == "false"
+} else = "ingress" {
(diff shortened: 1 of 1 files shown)


assets/queries/cloudFormation/ec2_network_acl_duplicate_rule/query.rego
```diff
@@ -34,9 +34,21 @@ getRef(obj) = obj.Ref {
 }
 
 getTraffic(entry) = "egress" {
-	object.get(entry.Properties, "Egress", "undefined") == true
+	entry.Properties.Egress == true
```
felipe-avelar, line 37: Is really necessary to check with and without `lower`? I think only checking with `lower` should be fine, no?

felipe-avelar (changes requested):
Check comments please

assets/queries/cloudFormation/ec2_network_acl_duplicate_rule/query.rego
Ruben-Silva: Using lower only would cause problems when the property came as a boolean, however, Rogério suggested a solution that minimizes the number of elses in that method.

rogeriopeixotocx (approved):
LGTM

PR #2132, opened 2021-02-24
Update getTasks on all Ansible queries

### LINKED ISSUE #2112: Update getTasks for Ansible queries
### Platform
Ansible

### Description
- `getTasks` not working correctly for tasks inside a `tasks` block
- `getTasks` should return the tasks inside `block`, `always` and `rescue` blocks
- `getTasks` from the Ansible library should be used on all queries

jorge-cruz:
Closes #2112 

**Proposed Changes**

- The `getTasks` function now returns all tasks inside `block`, `always` and `rescue` blocks.

To achieve this, the new `getTasksFromBlocks` function checks if a task has a `block` field which means it's a block task. If it's true, the function will try to find all tasks by executing the built-in [`walk`](https://www.openpolicyagent.org/docs/latest/policy-reference/#graphs) function that will traverse all values nested under the block task (Rego doesn't support recursion). This is important because Ansible allows nested `blocks`. Let's check an example:

```
- hosts: all
  tasks:
    - name: "First block"
      block:
        - name: "Task 1" 
          task1: do_something

        - name: "Task 2" 
          task2: do_something

        - name: "Second block"
          block:
            - name: "Task 3"
              task3: do_something

            - name: "Third block"
              block:
              - name: "Task 4"
                task4: do_something
      always:
        - name: "Task 5"
          task5: do_something
```

By executing the `walk(block, [path, value])` function, we will iterate upon all values inside the "First block" task. We need to check that `value` is an object and doesn't have a `block` field (we don't want queries checking block tasks). We also need to validate the `path` by checking if its length is bigger than 1 and if the penultimate element is a valid group ("block", "always", "rescue"). This will assure that `value` is a valid task.

- Created the `TasksPerDocument` rule that returns an object where the fields represent the documents IDs and the values are the list of tasks of the respective document. This object will be saved on the `tasks` variable in the library that will be used by the queries to get the final results. With this, the `getTasks` function is executed only once for each document instead of being executed in each query for each document. This important because the `walk` function is expensive in terms of time.

### DIFF
--- a/assets/libraries/ansible/library.rego
+++ b/assets/libraries/ansible/library.rego
@@ -27,24 +27,22 @@ getTasksFromBlocks(playbook) = result {
 		walk(playbook, [path, task])
 		is_object(task)
 		not task.block
-		valid_path(path)
+		validPath(path)
 	]
 } else = [playbook] {
 	true
 }
 
-valid_path(path) {
-	count(path) == 0
-} else {
-	count(path) % 2 == 0
-	valid_group(path[minus(count(path), 2)])
+validPath(path) {
+	count(path) > 1
+	validGroup(path[minus(count(path), 2)])
 }
 
-valid_group("block") = true
+validGroup("block") = true
 
-valid_group("always") = true
+validGroup("always") = true
 
-valid_group("rescue") = true
+validGroup("rescue") = true
 
 checkState(task) {
 	state := object.get(task, "state", "undefined")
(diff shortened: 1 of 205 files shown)


rogeriopeixotocx:
Tested and working:
example:
image
stdout:
image
payload:
image

this great stuff @jorge-cruz

rogeriopeixotocx (approved):

ruigomescx:
Great work @jorge-cruz , thank you for the improvement!

ruigomescx (approved):

PR #2300, opened 2021-03-18
Update cmk_is_unusable query Closes #2287

### LINKED ISSUE #2287: Update CMK Is Unusable query for Cloudformation
**Platform**
Cloudformation

**Provider**
AWS

**Description**
Don't need to check if the `Enabled` is undefined because it defaults to `true` and we only want to know if `Enabled` is `false`

NunoAraujoCX:
Closes #2287 
**Proposed Changes**

- Remove the query that looks if the attribute `Enabled` is undefined, because we only need to know if it is `false` and the attribute defaults to `true`


Ruben-Silva (approved):
Looks good to me!

ruigomescx (approved):

PR #2301, opened 2021-03-18
Update docs workflows are not running

rogeriopeixotocx:
Fixing bug in update docs workflow by changing the way we provide the personal access token


felipe-avelar (approved):
LGTM

PR #2310, opened 2021-03-18
Add "API Gateway With Open Access" query for Terraform

### LINKED ISSUE #2307: Add "API Gateway With Open Access" query for Terraform
### Platform
Terraform

### Provider
AWS

### Description
Add support in Terraform for a query with the same name already implemented in CloudFormation.
API Gateway Method should restrict the authorization type, except for the HTTP OPTIONS method.

mcarvalhox:
Closes #2307

**Proposed Changes**

- Add support in Terraform for a query with the same name already implemented in CloudFormation.
- The query checks if the authorization type is restricted in API Gateway Method resource (therefore being different from `NONE`). The authorization type can only be `NONE` when the http method is `OPTIONS`.

### DIFF
--- a/assets/queries/terraform/aws/api_gateway_with_open_access/query.rego
+++ b/assets/queries/terraform/aws/api_gateway_with_open_access/query.rego
@@ -1,0 +1,17 @@
+package Cx
+
+CxPolicy[result] {
+	document := input.document[i]
+	resource = document.resource.aws_api_gateway_method[name]
+
+	resource.authorization == "NONE"
+	resource.http_method != "OPTIONS"
+
+	result := {
+		"documentId": document.id,
+		"searchKey": sprintf("aws_api_gateway_method[%s].authorization", [name]),
+		"issueType": "IncorrectValue",
+		"keyExpectedValue": "aws_api_gateway_method.authorization is only 'NONE' if http_method is 'OPTIONS'",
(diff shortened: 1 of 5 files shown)


assets/queries/terraform/aws/api_gateway_with_open_access/query.rego
```diff
@@ -0,0 +1,17 @@
...
+CxPolicy[result] {
+	document := input.document[i]
+	resource = document.resource.aws_api_gateway_method[name]
+
+	resource.authorization == "NONE"
+	resource.http_method != "OPTIONS"
+
+	result := {
+		"documentId": document.id,
+		"searchKey": sprintf("aws_api_gateway_method[%s].authorization", [name]),
```
Ruben-Silva, line 12: Should be "http_method" instead of "authorization"?

"searchKey": sprintf("aws_api_gateway_method[%s].http_method", [name]),

assets/queries/terraform/aws/api_gateway_with_open_access/test/negative.tf
```diff
@@ -0,0 +1,11 @@
+resource "aws_api_gateway_method" "positive1" {
```
Ruben-Silva, line 1: Resource name is "positive1" in negative sample. Use "negative1" instead.

Ruben-Silva (changes requested):

Ruben-Silva (approved):
Looks good to me!

ruigomescx (approved):

PR #2476, opened 2021-03-25
Optimize golangci-lint

rogeriopeixotocx:
**Proposed Changes**

- `golangci-lint` is taking longer than usual, we should try different configurations to optimize its executions.


rogeriopeixotocx:
Linter is now running under a 5 digits milliseconds number.

ruigomescx (approved):

PR #2584, opened 2021-03-29
Update "API Gateway Method Does Not Contains An API Key Or Empty Methods" query for CloudFormation

### LINKED ISSUE #2579: Update "API Gateway Method Does Not Contains An API Key Or Empty Methods" query for CloudFormation
### Platform
CloudFormation

### Description
After consulting the [vulnerability being addressed with this query](https://gsl.dome9.com/D9.AWS.CRY.37.html), it seems that the only aspect that should be verified is if an API key is being required on a method request

mcarvalhox:
Closes #2579

**Proposed Changes**

- removed the last rule, that was verifying if attribute `HttpMethod` was empty
- renamed the query accordingly


Ruben-Silva (approved):
Looks good to me!

ruigomescx (approved):

PR #2645, opened 2021-03-31
Add "Invalid Image" query for Terraform Closes #2643

### LINKED ISSUE #2643: Add "Invalid Image" query for Terraform
### Platform
Terraform

### Provider
Kubernetes

### Description
Image must be defined and not be empty or equal to latest

rafaela-soares:
Closes #2643

**Proposed Changes**

- Support "Invalid Image" query for Terraform (same as "Invalid Image" for Kubernetes). It is necessary to check if the attribute image does not exist or is set to empty or "latest". Since `image =  `  in TF is not considered an attribute set to null, this case was not considered

### DIFF
--- a/assets/queries/k8s/invalid_image/query.rego
+++ b/assets/queries/k8s/invalid_image/query.rego
@@ -36,13 +36,7 @@ CxPolicy[result] {
 }
 
 check_content(images) {
-	images == ""
-}
+	options := {"", "latest", null}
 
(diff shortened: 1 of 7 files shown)


Ruben-Silva (approved):
Looks good to me!

Ruben-Silva (approved):
Still looks good!

ruigomescx (approved):

PR #2675, opened 2021-04-01
Add support to terraform vars

### LINKED ISSUE #2226: Support to Terraform variables
### Description
Currently Terraform variables are not being taken in account when KICS scans a terraform project. The goal is to support the parsing and use of [terraform variables](https://www.terraform.io/docs/language/values/variables.html).

felipe-avelar:
Closes #2226

**Proposed Changes**
- Add support to terraform input variables


rogeriopeixotocx (approved):
~LGTM~ LCTM

pkg/kics/sink.go
```diff
@@ -20,6 +20,11 @@ func (s *Service) sink(ctx context.Context, filename, scanID string, rc io.Reade
 		return errors.Wrapf(err, "failed to get file content: %s", filename)
 	}
 
+	content, err = s.Parser.Resolve(*content, filename)
+	if err != nil {
+		return errors.Wrapf(err, "failed to resolve file content: %s", filename)
+	}
+
```
joaoReigota1, line 27: Since each parser has its own resolver, can't we add this resolve function to pkg/parser/parser.go Parse?
ex: 
func (c *Parser) Parse(filePath string, fileContent []byte) ([]model.Document, model.FileKind, error) {
	ext := filepath.Ext(filePath)
	if ext == "" {
		ext = filepath.Base(filePath)
	}
	if p, ok := c.parsers[ext]; ok {
        resolved, err := p.Resolve(fileContent, filename)
		if err != nil {
			return nil, err
		}
		obj, err := p.Parse(filePath, resolved)
		if err != nil {
			return nil, "", err
		}

		return obj, p.GetKind(), nil
	}

	return nil, "", ErrNotSupportedFile
}

joaoReigota1 (dismissed):
LGTM

rogeriopeixotocx (approved):
LGTM

PR #2678, opened 2021-04-01
Docs changes should not trigger CI #2677

rogeriopeixotocx:
Closes #2677

**Proposed Changes**

- ignore `docs/** `and `README.md`

I submit this contribution under Apache-2.0 license.

Issue #2677, opened 2021-04-01
[CI fine tunning] Do not trigger tests when only docs are changed


rogeriopeixotocx:
### Expected Behavior
Tests workflows should not run when a PR only changes docs

### Actual Behavior
Tests workflows are being executed for all PR's

### Steps to Reproduce the Problem

1.  open a PR

### Specifications

-   Version: n/a
-   Platform: n/a
-   Subsystem: n/A

PR #2681, opened 2021-04-03
Revert "Docs changes should not trigger CI #2677"

rogeriopeixotocx:
Reverts Checkmarx/kics#2678

We cannot use `paths-ignore` feature because both workflows are "required" in our CI pipelines.

The issue is described here: https://github.community/t/feature-request-conditional-required-checks/16761

We might be able to workaround this issue by the using any of the suggestions in the thread above:
- https://github.com/marketplace/actions/skip-duplicate-actions#skip-ignored-paths
- https://github.com/technote-space/get-diff-action

cc: @joaoReigota1 @felipe-avelar @ruigomescx

### DIFF
--- a/.github/workflows/go-ci-integration.yml
+++ b/.github/workflows/go-ci-integration.yml
@@ -3,10 +3,6 @@ name: go-ci-integration
 on:
   pull_request:
     branches: [master]
-    paths-ignore:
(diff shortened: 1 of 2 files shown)


felipe-avelar (approved):
LGTM

PR #2763, opened 2021-04-09
Add "API Gateway Deployment Without Access Log Setting" query for Terraform Closes #2760

### LINKED ISSUE #2760: Add "API Gateway Deployment Without Access Log Setting" query for Terraform
### Platform
Terraform

### Provider
AWS

### Description
API Gateway Deployment should have access log setting defined when connected to an API Gateway Stage

rafaela-soares:
Closes #2760

**Proposed Changes**
- Support "API Gateway Deployment Without Access Log Setting" query for Terraform (same as "API Gateway Deployment Without Access Log Setting" for CloudFormation). It is necessary to check if: 

    1.  when an API Gateway Deployment is set, there is no API Gateway Stage defined
    2.  when an API Gateway Deployment is set, there is no API  Gateway Stage with 'access_log_settings' set associated with it
    3.  an API Gateway Deployment has an API  Gateway Stage with 'access_log_settings' set associated but does not have the attribute 'stage_description' set

### DIFF
--- a/assets/queries/cloudFormation/api_gateway_deployment_without_access_log_setting/query.rego
+++ b/assets/queries/cloudFormation/api_gateway_deployment_without_access_log_setting/query.rego
@@ -5,7 +5,7 @@ CxPolicy[result] {
 	resource = document[i].Resources[name]
 	resource.Type == "AWS::ApiGateway::Deployment"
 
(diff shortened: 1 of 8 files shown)


rogeriopeixotocx (approved):
LGTM

PR #2765, opened 2021-04-09
Add "CMK Rotation Disabled" query for Terraform and Ansible Closes #2764

### LINKED ISSUE #2764: Add "CMK Rotation Disabled" query for Terraform and Ansible
### Platform
Terraform and Ansible

### Provider
AWS

### Description
Customer Master Keys (CMK) must have rotation enabled, which means the attribute 'enable_key_rotation' must be set to 'true' when the key is enabled

rafaela-soares:
Closes #2764

**Proposed Changes**
- Changing query dir from  `\assets\queries\cloudFormation\cmks_rotation_disabled`  to ` \assets\queries\cloudFormation\cmk_rotation_disabled`
-  Support "CMK Rotation Disabled" query for Terraform and Ansible (same as "CMK Rotation Disabled" for CloudFormation). It is necessary to check if the attribute 'enable_key_rotation' is not set or set to false when the key is enabled


rogeriopeixotocx (approved):
LGTM

PR #2835, opened 2021-04-14
Test metrics using zsh script

rogeriopeixotocx:
**Proposed Changes**
- quick and dirty test metrics using zsh script
- JUnit test report should be generated even if tests failed
- we should consider python script with matplot lib in the future to generate graphics and post it back on the PR


felipe-avelar (approved):
LGTM

felipe-avelar (approved):
LGTM

PR #2960, opened 2021-04-22
feat(parser): support ssl certificates and swagger files set as attributes for Ansible and Terraform #2958

### LINKED ISSUE #2958: Add support to certificate and swagger files set as attributes in Ansible and Terraform
### Description
In Ansible, the attributes 'swagger_file' and 'certificate' only refers to the file name/path. The same happens for attribute 'certificate_body' for Terraform. To get access to the information of these attributes, the parser needs to be prepared

cosmicgirl97:
Closes #2958

**Proposed Changes**
- **The algorithm followed to access the content of the attribute 'swagger_file' in an Ansible template was**:
               **1**.  Verify if any attribute in the Ansible template is 'swagger_file' in the function 'playbookParser' (\kics\pkg\parser\yaml\parser.go);
               **2**.  In a positive case, add the content of the swagger file in the attribute 'swagger_file' (through function 'AddSwaggerInfo' used in the function 'playbookParser');

   **Functions description**:

  - **AddSwaggerInfo** (\kics\pkg\parser\additional\swagger_utils.go): Firstly, checks if the swagger path (defined in 'swagger_file') is not a full valid path or is an incomplete path. In case of true, joins the current dir with the incomplete path. Then, read the file and add the content;

- **The algorithm followed to access the content of the attribute 'certificate' in an Ansible template and the attribute 'certificate_body' in Terraform was**:
  
  **1.** For Ansible, verify if any attribute in the Ansible template is 'certificate' in the function 'playbookParser' (\kics\pkg\parser\yaml\parser.go); For Terraform, verify if any attribute in the Terraform template is 'certificate_body' and refers a pem file in the function 'Parse' (\kics\pkg\parser\terraform\terraform.go);
  **2**. In a positive case, add the content of the certificate (through function 'AddCertificateInfo').

  **Functions description**:

  - **AddCertificateInfo** (kics\pkg\parser\additional\certificate_utils.go): Firstly, checks if the certificate path (defined in the attribute related to certificate) is not a full valid path or is an incomplete path. In case of true, joins the current dir with the incomplete path. Then, read the file and add the information about the expiration date and the RSA Key bytes (through function 'getCertificateInfo')
  - **getCertificateInfo** (kics\pkg\parser\additional\certificate_utils.go): reads and parses the certificate PEM and gets the information about the expiration date (cert.NotAfter) and the RSA Key bytes (cert.PublicKey.(*rsa.PublicKey).Size())

**Analysis of consumption metrics** (still work in progress)

The following table is related to the scan of "\kics\assets\queries\ansible\aws\". For that, it needed to add a YAML file in queries ''API Gateway Endpoint Config is Not Private", "API Gateway Without SSL Certificate" and "API Gateway X-Ray Disabled" (the positive and negative files refer to a swagger_file that does not exist in the respective folder). In these two rounds, the results of metrics 'Total CPU usage for inspect', 'Total CPU usage for generate_report' and 'Total MEM usage for get_sources' are inconclusive.

|                                     | Without Parser Changes R1 | With Parser Changes R1 | Difference R1 | Without Parser Changes R2 | With Parser Changes R2 | Difference R2 |
|-------------------------------------|---------------------------|------------------------|---------------|---------------------------|------------------------|---------------|
| Total CPU usage for get_queries     |           13.11s          |         14.79s         | +1.68s        | 13.23s                    | 17.59s                 | +4.36s        |
| Total CPU usage for get_sources     |           3.74s           |          2.71s         | -1.03s        | 3.34s                     | 1.29s                  | -2.05s        |
| Total CPU usage for inspect         |           54.23s          |         56.95s         | +2.72s        | 56.70s                    | 45.96s                 | -10.74s       |
| Total CPU usage for generate_report |          680.00ms         |          1.13s         | +0.45s        | 1.23s                     | 630.00ms               | -0.6s         |
| Total MEM usage for get_queries     |          241.73MB         |        260.29MB        | +18.56MB      | 251.83MB                  | 261.15MB               | +9.32MB       |
| Total MEM usage for get_sources     |          344.13MB         |        311.62MB        | -32.51MB      | 301.76MB                  | 314.21MB               | +12.45MB      |
| Total MEM usage for inspect         |          358.83MB         |        357.80MB        | -1.03MB       | 363.87MB                  | 357.71MB               | -6.16MB       |
| Total MEM usage for generate_report |          360.03MB         |        357.80MB        | -2.23MB       | 363.87MB                  | 357.71MB               | -6.16MB       |

**Queries**:

- Added "Certificate RSA Key Bytes Lower Than 128" for Ansible and Terraform (AWS): It checks if a certificate uses a RSA key with length higher than 128 bytes
- Added "API Gateway Without Configured Authorizer" for Ansible (AWS). It checks if a swagger attribute has a authorizer set


pkg/parser/additional/additional.go
```diff
@@ -0,0 +1,199 @@
...
+		log.Error().Msgf("Failed to execute command 'openssl x509 -in %s -noout -dates'", filePath)
+	}
+
+	date := getExpirationDate(string(out))
+
+	return date
+}
+
+func getBitRSAKeyBit(filePath string) int {
+	out, err := exec.Command("openssl", "x509", "-in", filePath, "-text", "-noout").Output()
```
rogeriopeixotocx, line 88: This is going to add an external dependency to KICS, it will mandate the user to have `openssl` CLI installed which I don't think is a good idea right now. We would have to update our docs, add `openssl` to our docker images and etc...

Instead of using `openssl` to parse the PEM file have you seen these examples using go standard libraries to validate a certificate?

- https://github.com/timewasted/go-check-certs/blob/19b8ef135a9c83d7dd93e6a69c937fd04bdf7044/check-certs.go#L198-L205
- https://github.com/alanorth/check-tls-certs/blob/master/main.go

rogeriopeixotocx: Since, your intention is not to validate the certificate and just to extract the dates. Maybe this example is better:

First I extracted a x509 cert from google.com:

`true | openssl s_client -connect google.com:443 2>/dev/null | openssl x509 > google-ssl.pem`

Then:
```
package main

import (
	"crypto/x509"
	"encoding/pem"
	"fmt"
	"os"
)

func main() {
	certPEM, err := os.ReadFile("google-ssl.pem")
	if err != nil {
		panic("failed to find certificate")
	}

	block, _ := pem.Decode(certPEM)
	if block == nil {
		panic("failed to parse certificate PEM")
	}
	cert, err := x509.ParseCertificate(block.Bytes)
	if err != nil {
		panic("failed to parse certificate: " + err.Error())
	}

	fmt.Println(cert.NotAfter)
}
```

image

In this way you won't even need to parse the abbreviation of the months (ex: "Jan", "Feb", etc..) just split the string

pkg/parser/additional/additional.go
```diff
@@ -0,0 +1,199 @@
...
+
+	if err != nil {
+		log.Error().Msgf("Failed to execute command 'openssl x509 -in %s -text -noout'", filePath)
+		return -1
+	}
+
+	rsaInfo := getRegexMatch(`RSA Public-Key:[ ]\([0-9]{4}[ ]bit\)`, string(out))
```
rogeriopeixotocx, line 95: The same concern as in the previous comment.

Take a look at this example:

```go
package main

import (
	"crypto/x509"
	"encoding/pem"
	"fmt"
	"os"
)

func main() {
	certPEM, err := os.ReadFile("google-ssl.pem")
	if err != nil {
		panic("failed to find certificate")
	}

	block, _ := pem.Decode(certPEM)
	if block == nil {
		panic("failed to parse certificate PEM")
	}
	cert, err := x509.ParseCertificate(block.Bytes)
	if err != nil {
		panic("failed to parse certificate: " + err.Error())
	}
	//fmt.Println(cert.NotAfter)
	publicKeyDer, _ := x509.MarshalPKIXPublicKey(cert.PublicKey)
	publicKeyBlock := pem.Block{
		Type:  "PUBLIC KEY",
		Bytes: publicKeyDer,
	}
	publicKeyPem := string(pem.EncodeToMemory(&publicKeyBlock))
	fmt.Println(publicKeyPem)
}
```

image

pkg/parser/additional/additional.go
```diff
@@ -0,0 +1,199 @@
+package additional
```
rogeriopeixotocx, line 1: Maybe we should think of a better name to put the utility functions.

`certificate_utils.go` could be a candidate.

pkg/parser/additional/additional.go
```diff
@@ -0,0 +1,199 @@
...
+
+	return clean[0]
+}
+
+func getExpirationDate(datetime string) [3]int {
+	// get notAfter datetime
+	expirationInfo := getRegexMatch(`(notAfter=)[a-zA-Z]+[ ]+[0-9]{1,2}[ ]([0-9]{2}:[0-9]{2}:[0-9]{2})[ ][0-9]{4}`, datetime)
+	clean := strings.Split(expirationInfo, "=")
+
+	months := map[string]int{
```
rogeriopeixotocx, line 39: I'm not sure if you have to do this. Maybe there's a function under the golang's time package to parse abbreviated months.

pkg/parser/additional/additional.go
```diff
@@ -0,0 +1,199 @@
...
+		case map[interface{}]interface{}:
+			res[fmt.Sprint(k)] = convert(v2)
+		default:
+			res[fmt.Sprint(k)] = v
+		}
+	}
+	return res
+}
+
+func readFile(filePath string) map[string]interface{} {
```
rogeriopeixotocx, line 153: This should be in a separate file.

pkg/parser/additional/additional.go
```diff
@@ -0,0 +1,199 @@
...
+			log.Error().Msgf("Failed to unmarshal '%s'", fileExtension)
+		}
+		result = convert(resultYaml)
+	}
+
+	return result
+}
+
+// AddSwaggerInfo gets and adds the content of a swagger file
+func AddSwaggerInfo(path, swaggerPath string) map[string]interface{} {
```
rogeriopeixotocx, line 181: Swagger related code should be placed under a different file from the certificate utils.

pkg/parser/terraform/terraform.go
```diff
@@ -47,7 +48,32 @@ func (p *Parser) Parse(path string, content []byte) ([]model.Document, error) {
...
+							certInfo = additional.AddCertificateInfo(path, v3.(string))
+							if certInfo != nil {
+								elements["certificate_body"] = certInfo
+							}
+						}
+					}
+				}
+			}
+		}
+	}
```
rogeriopeixotocx, line 74: This entire block could be a performance bottleneck. We need to investigate resource consumption and compare it to a baseline.

Take a look at @joaoReigota1 's PR for resource consumption metrics #2961

See if you can merge his code to your branch, or wait for this branch to be merged (probably next week) so you can rebase your changes on top of the updated master. Then you need to provide evidence that this feature will not take a critical hit into the engine's performance.

pkg/parser/yaml/parser.go
```diff
@@ -88,5 +89,32 @@ func playbookParser(fileContent []byte) ([]model.Document, error) {
...
+					if elements["swagger_file"] != nil {
+						swaggerInfo = additional.AddSwaggerInfo(filePath, elements["swagger_file"].(string))
+						if swaggerInfo != nil {
+							elements["swagger_file"] = swaggerInfo
+						}
+					}
+				}
+			}
+		}
+	}
```
rogeriopeixotocx, line 117: Again some performance concerns

go.mod
```diff
@@ -37,6 +37,7 @@ require (
 	go.opencensus.io v0.22.4 // indirect
 	golang.org/x/crypto v0.0.0-20210220033148-5ea612d1eb83 // indirect
 	golang.org/x/net v0.0.0-20210316092652-d523dce5a7f4
+	gopkg.in/yaml.v2 v2.4.0 // indirect
```
rogeriopeixotocx, line 40: Why we need an older version of the `yaml` module?

rogeriopeixotocx (changes requested):
This is good stuff. But it still needs work.

rogeriopeixotocx:
Also @cosmicgirl97 (AKA @rafaela-soares) We need unit tests 😄

go.mod
cosmicgirl97: Sorry, my bad. I was using v2 in the file 'additional.go' 🤣. I already removed it.

pkg/parser/additional/additional.go
cosmicgirl97: Thank you so much, Rogério. I appreciate your suggestions. I already adopted them.

pkg/parser/additional/certificate_utils.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	return true, match[0]
+}
+
+func getCertificateInfo(filePath string) (a [3]int, i int) {
+	certPEM, err := os.ReadFile(filePath)
+
+	if err != nil {
+		panic("Failed to read the file " + filePath)
```
joaoReigota1, line 28: We shouldn't be panicking here, maybe create a struct that takes in a and i return values and return the struct and the error 

```
type returnStruct struct {
   a [3]int
   i int
}

func getCertificateInfo(filePath string) (returnStruct, error) {
	certPEM, err := os.ReadFile(filePath)

	if err != nil {
		return returnStruct{}, err
```

pkg/parser/additional/certificate_utils.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	block, _ := pem.Decode(certPEM)
+	if block == nil {
+		panic("Failed to parse the certificate PEM " + filePath)
```
joaoReigota1, line 33: Same as above

pkg/parser/additional/certificate_utils.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	cert, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		panic("Failed to parse the certificate" + filePath)
```
joaoReigota1, line 37: Same as above

pkg/parser/terraform/terraform.go
```diff
@@ -47,7 +48,35 @@ func (p *Parser) Parse(path string, content []byte) ([]model.Document, error) {
...
+								}
+							}
+						}
+					}
+				}
+			}
+		}
+	}
+
+	return json, errors.Wrap(parseErr, "failed terraform parse")
```
joaoReigota1, line 79: Maybe you can refactor this part

pkg/parser/yaml/parser.go
joaoReigota1, line 117: Maybe you can refactor this part

joaoReigota1 (changes requested):

rogeriopeixotocx:
@cosmicgirl97 @rafaela-soares do you have a query to add that uses the certificate info appended to the payload? It would be great to see this new feature in action being used by a query in this same PR

pkg/parser/additional/certificate_utils.go
```diff
@@ -0,0 +1,79 @@
...
+// CheckCertificateBody verifies if the attribute 'certificate_body' refers a file
+func CheckCertificateBody(content string) (b bool, s string) {
+	var re = regexp.MustCompile(`[0-9a-zA-Z-/\\_]+\.pem`)
+	match := re.FindAllString(content, -1)
+
+	if match == nil {
+		return false, ""
+	}
+
+	return true, match[0]
```
felipe-avelar, line 21: If you want only the first match, why no use [FindString](https://golang.org/pkg/regexp/#Regexp.FindString) instead `FindAllString`.

Also, instead of returning `(bool, string)`, I think you could return just string and when string is empty is the same as false... So you could just return the matched string

pkg/parser/additional/certificate_utils.go
```diff
@@ -0,0 +1,79 @@
...
+	var date [3]int
+	date[0] = cert.NotAfter.Year()
+	date[1] = int(cert.NotAfter.Month())
+	date[2] = cert.NotAfter.Day()
+
+	var rsaBytes int
+
+	switch t := cert.PublicKey.(type) {
+	case *rsa.PublicKey:
+		_ = t
```
felipe-avelar, line 49: If you are not using t for anything, I think you can change this to `switch cert.PublicKey.(type) {` and remove `_ = t`

felipe-avelar (changes requested):
Please, check comments

pkg/parser/additional/certificate_utils.go
cosmicgirl97: Thank you so much for your suggestions, Felipe 😄

pkg/parser/additional/certificate_utils.go
cosmicgirl97: It was my first approach, but 'golangci-lint' seems not to like it 😆 If I remove 't', 'golangci-lint' returns " typeSwitchVar: 1 case can benefit from type switch with assignment (gocritic)"

pkg/parser/additional/certificate_utils.go
cosmicgirl97: Thank you so much for your suggestions, João 😄

rogeriopeixotocx:
@cosmicgirl97 @rafaela-soares please add the new terraform samples e.g: assets/queries/terraform/aws/certificate_rsa_key_bytes_lower_than_128/test/negative2.tf to the [ignore list](https://github.com/Checkmarx/kics/blob/master/.github/scripts/samples-linters/ignore-list/terraform) until we fix this.

pkg/parser/terraform/terraform.go
```diff
@@ -36,6 +37,36 @@ func (p *Parser) Resolve(fileContent []byte, filename string) (*[]byte, error) {
...
+								if certInfo != nil {
+									elements["certificate_body"] = certInfo
+								}
+							}
+						}
+					}
+				}
+			}
+		}
+	}
```
rogeriopeixotocx, line 65: Extract nested for loops to separated functions passing down parameters and returning values as needed. I could only find an example of this in Python https://codereview.stackexchange.com/a/191823

pkg/parser/yaml/parser.go
```diff
@@ -65,7 +66,41 @@ func (p *Parser) GetKind() model.FileKind {
...
+					}
+					if elements["swagger_file"] != nil {
+						swaggerInfo = additional.AddSwaggerInfo(filePath, elements["swagger_file"].(string))
+						if swaggerInfo != nil {
+							elements["swagger_file"] = swaggerInfo
+						}
+					}
+				}
+			}
+		}
```
rogeriopeixotocx, line 97: Extract nested for loops to separated functions passing down parameters and returning values as needed. I could only find an example of this in Python https://codereview.stackexchange.com/a/191823

rogeriopeixotocx (changes requested):

pkg/parser/yaml/parser.go
cosmicgirl97: Thank you so much, Rogério!

joaoReigota1 (approved):
LGTM
Won't affect, concurrent Scans PR since the changes are made in the parser

pkg/parser/additional/certificate_utils.go
```diff
@@ -0,0 +1,86 @@
...
+		attributes := make(map[string]interface{})
+		attributes["file"] = filePath
+		attributes["expiration_date"] = date.date
+
+		if date.rsaKeyBytes != -1 {
+			attributes["rsa_key_bytes"] = date.rsaKeyBytes
+		}
+
+		return attributes
+	}
```
joaoReigota1, line 83: We should Log when err != nil
```suggestion
	if err == nil {
		attributes := make(map[string]interface{})
		attributes["file"] = filePath
		attributes["expiration_date"] = date.date

		if date.rsaKeyBytes != -1 {
			attributes["rsa_key_bytes"] = date.rsaKeyBytes
		}

		return attributes
	} 
	log.Trace().Msgf("failed to get certificate path %s: %s", filepath, err)
```

pkg/parser/additional/certificate_utils.go
```diff
@@ -0,0 +1,86 @@
+package additional
```
joaoReigota1, line 1:
```suggestion
package utils
```

pkg/parser/additional/certificate_utils.go
joaoReigota1: filepath variable can lead to a missing file as well

pkg/parser/additional/file_utils.go
```diff
@@ -0,0 +1,54 @@
+package additional
```
joaoReigota1, line 1:
```suggestion
package utils
```

joaoReigota1 (changes requested):

rogeriopeixotocx (approved):
LGTM

PR #2973, opened 2021-04-23
feat(query): "Responses Object Is Empty" for OpenAPI (#2972)

### LINKED ISSUE #2972: Add "Responses Object Is Empty" query for OpenAPI
### Platform
OpenAPI

### Description
Responses Object should not be empty

rafaela-soares:

Closes #2972

**Proposed Changes**
- Added "Responses Object Is Empty" query for OpenAPI. It checks if Responses Object is empty. Since the parser for YAML defines responses as null when empty and the parser for JSON defines as {} these two cases were considered

Considerations:
- The Response Object exists in Component Object and Operation Object


assets/queries/openAPI/responses_object_is_empty/query.rego
```diff
@@ -0,0 +1,35 @@
...
+		"keyExpectedValue": sprintf("paths.{{%s}}.{{%s}}.responses is not empty", [path, op]),
+		"keyActualValue": sprintf("paths.{{%s}}.{{%s}}.responses is empty", [path, op]),
+	}
+}
+
+CxPolicy[result] {
+	doc := input.document[i]
+	openapi_lib.check_openapi(doc) != "undefined"
+
+	doc.components.responses == options[x]
```
joaoReigota1, line 26: Since responses in components a map, shouldn't we add another verification to check if every entry in responses object map are not empty as well ex:
```
components:
  responses:
    Success:
    Error:
      ... 
```
Where in this case the Success entry is empty but the Error is not, I believe the Success entry should give a result as well

joaoReigota1 (changes requested):

assets/queries/openAPI/responses_object_is_empty/query.rego
rafaela-soares: Responses Object is a map of response object in components. Since this query is related to check if 'responses' object is empty, is it relevant to check for the 'response' object as well?

joaoReigota1 (approved):
LGTM

PR #3032, opened 2021-04-27
feat(engine): enhanced query selection and improved performance

### LINKED ISSUE #1798: Improve query selection mechanism
### Description
Currently, we're applying loading and trying to apply every query for each IaC file found. We should improve on this and only apply queries to the relevant counterparts files.

joaoReigota1:
Closes #1798

**Proposed Changes**
- Improved query selection mechanism 
- Added analyzer that goes through paths to determine which types of queries to use
- Fixed malformed queries


pkg/analyzer/analyzer.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/pkg/errors"
+	"github.com/rs/zerolog/log"
+)
+
+// openAPIRegex - Regex that finds OpenAPI defining property
+// cloudRegex - Regex that finds Cloud Formation defining property
+// k8sRegex - Regex that finds Kubernetes defining property
+var (
+	openAPIRegex = regexp.MustCompile("(\\s*\"openapi\":)|(\\s*openapi:)")
+	cloudRegex   = regexp.MustCompile("(\\s*\"Resources\":)|(\\s*Resources:)")
```
felipe-avelar, line 19: This is the only way to identify cloud formation file? I'm asking this, because if, for some reason, we have others resources files containing `Resource` keyword, it would be a problem, no?

felipe-avelar (changes requested):
Please, check comments

felipe-avelar (commented):
I was thinking, is it possible to add a platform field on document? So we could check that on queries if needed...

pkg/analyzer/analyzer.go
rogeriopeixotocx: I think it needs to be uppercase R. But we can also test this: 
https://github.com/aws-cloudformation/aws-cloudformation-template-schema/blob/master/src/main/resources/Schema.template#L371

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resources-section-structure.html

rogeriopeixotocx: It's too bad that `Resources` is the only required field: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-anatomy.html  But we can try find a pattern that matches the any resources inside the `Resources:` key.

pkg/analyzer/analyzer.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	for _, file := range files {
+		wg.Add(1)
+		// analyze the files concurrently
+		go worker(file, results, &wg)
+	}
+
+	go func() {
+		// close channel results when worker has fineshed writing into it
```
rogeriopeixotocx, line 62:
```suggestion
		// close channel results when worker has finished writing into it
```

pkg/analyzer/analyzer.go
```diff
@@ -0,0 +1,140 @@
...
+			availableTypes = append(availableTypes, i)
+		}
+	}
+	// stop metrics for file analyzer
+	metrics.Metric.Stop()
+	return availableTypes, nil
+}
+
+// worker determines the type of the file by ext (dockerfile and terraform)/content and
+// writes the awnser to the results channel
```
rogeriopeixotocx, line 80:
```suggestion
// writes the answer to the results channel
```

pkg/analyzer/analyzer_test.go
```diff
@@ -0,0 +1,59 @@
...
+	tests := []struct {
+		name    string
+		paths   []string
+		want    []string
+		wantErr bool
+	}{
+		{
+			name:    "analyze_test_dir_single_path",
+			paths:   []string{filepath.FromSlash("../../test/fixtures/analyzer_test")},
+			want:    []string{"dockerfile", "cloudformation", "kubernetes", "openapi", "terraform", "ansible"},
```
rogeriopeixotocx, line 21: Can you try to add a scenario where we have the `Resources:` key word inside a terraform file?

rogeriopeixotocx (changes requested):

rogeriopeixotocx:
> I was thinking, is it possible to add a platform field on document? So we could check that on queries if needed...

Looks like a good idea to me

internal/console/scan.go
```diff
@@ -380,6 +381,18 @@ func createInspector(t engine.Tracker, querySource source.QueriesSource) (*engin
 	return inspector, nil
 }
 
+func analyzePaths(paths, types []string) ([]string, error) {
+	var err error
+	if types[0] == "" { // if '--type' flag was given skip file analyzing
```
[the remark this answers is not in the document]
joaoReigota1: types always have one entry, so that we can have the common queries loaded, meaning that it will never return an illegal access error

pkg/analyzer/analyzer_test.go
joaoReigota1: In a Terraform file, it will never go to the check content function, since terraform files are defined by the extension of the file, but I can add to other technologies to check

pkg/analyzer/analyzer.go
joaoReigota1: I will add more cases to unit tests for testing purposes

felipe-avelar (approved):
LGTM

rogeriopeixotocx (approved):
LGTM

rogeriopeixotocx:
@joaoReigota1 this is awesome! "thanks"

PR #3052, opened 2021-04-30
feat(query): "Missing Mandatory Fields for Minimum Working File" for OpenAPI (#3031)

### LINKED ISSUE #3031: Add "Missing Mandatory Fields for Minimum Working File" query for OpenAPI
### Platform
OpenAPI

### Description
Mandatory fields for a minimum working file (such as 'info', 'info.version', 'info.title', and 'paths') should be defined.

rafaela-soares:

Closes #3031

**Proposed Changes**
- Added "Missing Mandatory Fields for Minimum Working File" query for OpenAPI. It checks if mandatory fields for a minimum working file (such as 'info', 'info.version', 'info.title', and 'paths') are undefined.


felipe-avelar (commented):
This query should checks all required fields or just minimum fields to have a working API? If it should checks all mandatory fields, some fields are missing (like name from License or url from Server). If it should checks just mandatory fields for a minimum working file, then, I think, the description should be more explicit about this...

rafaela-soares:
> This query should checks all required fields or just minimum fields to have a working API? If it should checks all mandatory fields, some fields are missing (like name from License or url from Server). If it should checks just mandatory fields for a minimum working file, then, I think, the description should be more explicit about this...

If we check all required fields, it will be complex to implement since there are several objects. Furthermore, objects are set in other ones. A better implementation could be to check just mandatory fields for a minimum working file

assets/queries/openAPI/missing_required_property/query.rego
```diff
@@ -0,0 +1,45 @@
...
+
+import data.generic.openapi as openapi_lib
+
+CxPolicy[result] {
+	doc := input.document[i]
+	openapi_lib.check_openapi(doc) != "undefined"
+
+	objects := {"paths", "info"}
+
+	object.get(doc, objects[x], "undefined") == "undefined"
```
felipe-avelar, line 11: I think, here, we could use `check_field` too, no?

assets/queries/openAPI/missing_required_property/metadata.json
```diff
@@ -0,0 +1,9 @@
+{
+  "id": "cf5ba5d2-5ad5-4582-8200-325fa75f606d",
+  "queryName": "Missing Mandatory Fields for Minimum Working File",
+  "severity": "INFO",
+  "category": "Structure and Semantics",
+  "descriptionText": "Mandatory fields for a minimum working file (such as 'info', 'info.version', 'info.title', and 'paths') should be defined.",
+  "descriptionUrl": "https://swagger.io/specification/",
```
felipe-avelar, line 7: I think https://swagger.io/specification/#openapi-object would be a better reference for this query, since it is the "root object"

felipe-avelar (changes requested):
Check comments, please

assets/queries/openAPI/missing_required_property/query.rego
rafaela-soares: The function `check_field` was implemented to resolve cases when the YAML file has 'info' empty. (YAML parser puts info set to null when it is empty (when set to null, `object.get(field, value, "undefined") == "undefined"` does not return a true positive result, because it can not access the 'field'). In the case of checking doc, I think it is unnecessary to use `check_field` since 'doc' is never null (it has 'id' and 'file' set, for example)

felipe-avelar (approved):
LGTM

PR #3073, opened 2021-05-03
feat(query): "Schema Object Incorrect Ref" for OpenAPI (#3072)

### LINKED ISSUE #3072: Add "Schema Object Incorrect Ref" query for OpenAPI
### Platform
OpenAPI

### Description
Schema Body reference must always point to '#components/schemas'

rafaela-soares:

Closes #3072

Proposed Changes
- Added "Schema Object Incorrect Ref" query for OpenAPI. It checks if schema object reference does not point to '#components/schemas'
- Updated "Response Object With Incorrect Ref": keyExpectedValue and keyActualValue
- Updated "Request Body With Incorrect Ref" : issueType
- The function 'regex.match' was chosen to also cover cases when '$ref' is a remote or a URL reference (function 'startswith' just covers cases when '$ref' is a local reference - https://swagger.io/docs/specification/using-ref/)

### DIFF
--- a/assets/libraries/openapi/library.rego
+++ b/assets/libraries/openapi/library.rego
@@ -24,5 +24,5 @@ incorrect_ref(ref, object) {
 		"requestBody": "#/components/requestBodies",
 	}
 
-	not contains(ref, references[object])
+	regex.match(references[object], ref) == false
 }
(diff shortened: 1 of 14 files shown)


joaoReigota1 (approved):
LGTM

PR #3089, opened 2021-05-04
feat(query): "Certificate RSA Key Bytes Lower Than 256" for Ansible and Terraform (#3088)

### LINKED ISSUE #3088: Update "Certificate RSA Key Bytes Lower Than 128" query for Ansible and Terraform
### Platform
Ansible and Terraform

### Provider
AWS

### Description
NIST recommends a minimum of 2048-bit keys for RSA, which means a minimum of 256 bytes

cosmicgirl97:
Closes #3088

**Proposed Changes**
- Updated "Certificate RSA Key Bytes Lower Than 128" to "Certificate RSA Key Bytes Lower Than 256": NIST recommends a minimum of 2048-bit keys for RSA, which means a minimum of 256 bytes


rogeriopeixotocx (approved):
LGTM

PR #3156, opened 2021-05-07
feat(query): "JSON Object Schema Without Properties" for OpenAPI (#3153)

### LINKED ISSUE #3153: Add "JSON Object Schema Without Properties" query for OpenAPI
### Platform
OpenAPI

### Description
Schema of the JSON object should have properties defined and 'additionalProperties' set to false

rafaela-soares:

Closes #3153

**Proposed Changes**
- Added "JSON Object Schema Without Properties" query for OpenAPI. It checks if  the schema of the JSON object does not have properties defined and/or 'additionalProperties' set to false


felipe-avelar (changes requested):
I think this rule can be simplified by using `walk` function...

```rego
[path, value] := walk(docs)
ref := value.schema["$ref"]
openapi_lib.undefined_field_in_json_object(doc, schema_ref, "properties")
```
or something like this...

Also, I think we don't need `content_allowed` checked, since this is checked on others queries

rogeriopeixotocx (approved):

PR #3288, opened 2021-05-12
feat(query): Added Resource Not Using Tags query for Terraform

joaoReigota1:
**Proposed Changes**
- Added Resource Not Using Tags

AWS services resource tags are an essential part of managing components


rogeriopeixotocx (approved):
LGTM

PR #3317, opened 2021-05-14
feat(query): "Media Type Object Without Schema" for OpenAPI

rafaela-soares:

**Proposed Changes**
- Added "Media Type Object Without Schema" query for OpenAPI. It checks if the Media Type Object does not have 'schema' defined


rogeriopeixotocx (approved):
LGTM

PR #3320, opened 2021-05-14
feat(query): "Parameter Object Without Schema" for OpenAPI

rafaela-soares:

**Proposed Changes**
- Added "Parameter Object Without Schema" query for OpenAPI. It checks if the Parameter Object does not have 'schema' defined


rogeriopeixotocx (approved):
LGTM

PR #3321, opened 2021-05-14
feat(query): Added IAM Role Policy passRole Allows All query for Terraform

joaoReigota1:

**Proposed Changes**
- Added IAM Role Policy passRole Allows All query for Terraform

Using the iam:passrole action with wildcards (*) in the resource can be overly permissive because it allows iam:passrole permissions on multiple resources


rogeriopeixotocx (approved):
LGTM

PR #3355, opened 2021-05-17
feat(query): "Empty Array" for OpenAPI

rafaela-soares:

Proposed Changes
- Added "Empty Array" query for OpenAPI. It checks if an array field is empty


assets/queries/openAPI/empty_array/query.rego
```diff
@@ -0,0 +1,52 @@
...
+import data.generic.openapi as openapi_lib
+
+CxPolicy[result] {
+	doc := input.document[i]
+	openapi_lib.check_openapi(doc) != "undefined"
+
+	[path, value] := walk(doc)
+
+	property := value[x]
+	is_empty_array(property)
```
felipe-avelar, line 12: I think we can use `value` directly with something like this:
```suggestion
	is_empty_array(value)
```

assets/queries/openAPI/empty_array/query.rego
```diff
@@ -0,0 +1,52 @@
...
+}
+
+is_empty_array(property) {
+	is_array(property)
+	count(property) == 0
+}
+
+is_empty_array(property) {
+	property == null
+}
```
felipe-avelar, line 52: What happens if there is something like:
```js
{
  "openapi": "3.0",
//...
  "components": null
}
```
?

assets/queries/openAPI/empty_array/query.rego
```diff
@@ -0,0 +1,52 @@
...
+
+CxPolicy[result] {
+	doc := input.document[i]
+	openapi_lib.check_openapi(doc) != "undefined"
+
+	[path, value] := walk(doc)
+
+	property := value[x]
+	is_empty_array(property)
```
felipe-avelar, line 32: I think we can use `value` directly with something like this:
```suggestion
	is_empty_array(value)
```

assets/queries/openAPI/empty_array/query.rego
```diff
@@ -0,0 +1,52 @@
...
+
+	count(path) != 0
+
+	result := {
+		"documentId": doc.id,
+		"searchKey": sprintf("%s.%s", [openapi_lib.concat_path(path), x]),
```
felipe-avelar, line 18: If changes are accepted, I think `openapi_lib.concat_path(path)` will give full path...

assets/queries/openAPI/empty_array/query.rego
```diff
@@ -0,0 +1,52 @@
...
+CxPolicy[result] {
+	doc := input.document[i]
+	openapi_lib.check_openapi(doc) != "undefined"
+
+	[path, value] := walk(doc)
+
+	property := value[x]
+	is_empty_array(property)
+
+	count(path) == 0
```
felipe-avelar, line 34: This will ever happen? Since path includes `value` key, I think it will never be an empty array, no?

felipe-avelar (changes requested):
Please check comments

assets/queries/openAPI/empty_array/query.rego
rafaela-soares: It will report as an empty array. This case was included because in the YAML sample, cases like `"components":  ` are represented as `"components": null `. However, on second thought, this case does not make sense since we just want to check if the array, when defined, is empty

assets/queries/openAPI/empty_array/query.rego
rafaela-soares: Because of `property := value[x]`, it happens. But with your suggestion, no 🙂

felipe-avelar (approved):
LGTM

PR #3373, opened 2021-05-19
feat(query): Added Schema Object With Circular Ref query for OpenAPI

joaoReigota1:

**Proposed Changes**
- Added Schema Object With Circular Ref query for OpenAPI

Schema Object should not reference it self in 'allOf', 'oneOf', 'anyOf' and 'not' properties


rogeriopeixotocx (approved):
LGTM

PR #3387, opened 2021-05-20
feat(query): Added Schema Object Using Enum With Keyword query to OpenAPI

joaoReigota1:

**Proposed Changes**
- Added Schema Object Using Enum With Keyword query to OpenAPI

Schema Object properties should not contain 'enum' and schema keywords

### DIFF
--- a/assets/queries/openAPI/schema_object_using_enum_with_keyword/query.rego
+++ b/assets/queries/openAPI/schema_object_using_enum_with_keyword/query.rego
@@ -1,0 +1,33 @@
+package Cx
+
+import data.generic.common as common_lib
+import data.generic.openapi as openapi_lib
+
+CxPolicy[result] {
+	keywords := ["multipleOf", "maximum", "minimum", "exclusiveMaximum", "exclusiveMinimum", "pattern", "minLength", "maxLength", "maxItems", "minItems", "uniqueItems", "items", "required", "maxProperties", "minProperties"]
+
+	doc := input.document[i]
+	openapi_lib.check_openapi(doc) != "undefined"
+
+	[path, value] := walk(doc)
+	properties := value.properties[name]
+	object.get(properties, "enum", "undefined") != "undefined"
+	object.get(properties, keywords[x], "undefined") != "undefined"
+
+	path_c := {x | obj := clean_path(path[n]); obj != ""; x := obj}
+
+	result := {
+		"documentId": doc.id,
+		"searchKey": sprintf("%s.%s", [openapi_lib.concat_path(path), name]),
+		"issueType": "IncorrectValue",
+		"keyExpectedValue": sprintf("%s.properties.%s does not contain 'enum' and schema keyword", [concat(".", path_c), name]),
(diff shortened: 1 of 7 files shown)


assets/queries/openAPI/schema_object_using_enum_with_keyword/query.rego
```diff
@@ -0,0 +1,33 @@
+package Cx
+
+import data.generic.common as common_lib
+import data.generic.openapi as openapi_lib
+
+CxPolicy[result] {
+	keywords := ["multipleOf", "maximum", "minimum", "exclusiveMaximum", "exclusiveMinimum", "pattern", "minLength", "maxLength", "maxItems", "minItems", "uniqueItems", "items", "required", "maxProperties", "minProperties"]
```
felipe-avelar, line 7: I think `items` should not be on this list, since it is a Schema Object

assets/queries/openAPI/schema_object_using_enum_with_keyword/query.rego
```diff
@@ -0,0 +1,33 @@
...
+
+	doc := input.document[i]
+	openapi_lib.check_openapi(doc) != "undefined"
+
+	[path, value] := walk(doc)
+	properties := value.properties[name]
```
felipe-avelar, line 13: I think here should be `schema` and `schemas[name]`, since we want to check all schema objects and not only the properties of a schema typed as object...

assets/queries/openAPI/schema_object_using_enum_with_keyword/query.rego
```diff
@@ -0,0 +1,33 @@
...
+		"keyActualValue": sprintf("%s.properties.%s contains 'enum' and schema keyword '%s'", [concat(".", path_c), name, keywords[x]]),
+	}
+}
+
+clean_path(path) = cleaned_path {
+	is_number(path)
+	cleaned_path := ""
+} else = cleaned_path {
+	cleaned_path := path
+}
```
felipe-avelar, line 33: I think `concat_path` already do this check, this is really necessary?

felipe-avelar (dismissed):
Please, check comments

assets/queries/openAPI/schema_object_using_enum_with_keyword/query.rego
joaoReigota1: resolved

assets/queries/openAPI/schema_object_using_enum_with_keyword/query.rego
joaoReigota1: resolved

joaoReigota1: I belive it should be properties since, `enum` can only be used inside schema.properties

assets/queries/openAPI/schema_object_using_enum_with_keyword/query.rego
joaoReigota1: this case is only for the key expected value and actual value since we don't want '{{}}' in these fields

rogeriopeixotocx (approved):
LGTM

PR #3405, opened 2021-05-21
feat(query): add "Property 'allowEmptyValue' Ignored" for OpenAPI

rafaela-soares:

**Proposed Changes**
- Added "Property 'allowEmptyValue' Ignored" query for OpenAPI.
- Since property 'allowEmptyValue' is ignored in the following cases: {"sytle": "simple", "explode": false}, {"sytle": "simple", "explode": true}, {"sytle": "spaceDelimited", "explode": false}, {"sytle": "pipeDelimited", "explode": false}, and {"sytle": "deepObject", "explode": true}, we need to check:
  - If `style` is set to `simple` or if it is undefined when `in` is set to `path` or `header` (by default, when `in` is set to `path` or `header`, `style` is `simple` (It is not necessary to check `explode` since it is ignored in both cases - explode set to false or explode set to true)
  - If `style` is set to `spaceDelimited` or `pipeDelimited` when 'explode' is `false` (in other words, when it is set to false or it is undefined (by default, explode is set to false, except for form style))
  - If `style` is set to `deepObject` when 'explode' is `true`

### DIFF
--- a/assets/queries/openAPI/property_allow_empty_value_ignored/metadata.json
+++ b/assets/queries/openAPI/property_allow_empty_value_ignored/metadata.json
@@ -1,0 +1,9 @@
+{
+  "id": "59c2f769-7cc2-49c8-a3de-4e211135cfab",
+  "queryName": "Property 'allowEmptyValue' Ignored",
+  "severity": "INFO",
+  "category": "Best Practices",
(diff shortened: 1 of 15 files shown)


felipe-avelar (approved):
LGTM

rogeriopeixotocx (approved):
LGTM

PR #3428, opened 2021-05-24
fix(detector): Added the possibility to use /" inside '[]' #2010

### LINKED ISSUE #2010: SearchKey on Terraform queries point to same line with multiple resources of the same type
### Expected Behavior
With multiple resources of the same type, each occurrence should point to the respective lines of the resource.
### Actual Behavior
The results always point to the first occurrence, even with a different searchKey.
Take, for example, the query "S3 bucket without encryption at REST" or this snippet for that query:
```
CxPolicy[result] {
	bucket := input.document[i].resource.aws_s3_bucket[name]
        object.get(bucket,"server_side_encryption_configuration","undefined") == "undefined"

	result := {
		"searchKey": sprintf("aws_s3_bucket[%s]", [name]),
		"issueType": "MissingAttribute",
		"keyExpectedValue": "'aws_s3_bucket.server_side_encryption_configuration' exists",
		"keyActualValue": "'aws_s3_bucket.server_side_encryption_configuration' is missing",
	}
}
```

Then run the query against this sample:
```
resource "aws_s3_bucket" "b" {
        bucket = "my-tf-test-bucket"
        acl    = "private"
}
 
resource "aws_s3_bucket" "a" {
  bucket = "my-another-tf-test-bucket"
  acl    = "private"
}
```

Two occurrences should result by the execution, one that points to the line where *aws_s3_bucket* named **a** is, and another pointed to the resource named **b**. Instead, both lines are pointing to the first resource, **b**.

### Specifications

-   Version: 1.1.2
-   Platform: Windows

joaoReigota1:

Closes #2010

Added the possibility to add \" to '[]' in the search key
ex: 
```
searchKey: sprintf("aws_resource[\"%s\"]", [name])
```

**Proposed Changes**
- Added the possibility to use \" inside '[]' 

### DIFF
--- a/pkg/detector/helper.go
+++ b/pkg/detector/helper.go
@@ -12,7 +12,7 @@ import (
 )
 
 var (
-	nameRegex       = regexp.MustCompile(`^([A-Za-z0-9-_]+)\[([A-Za-z0-9-_{}]+)]$`)
+	nameRegex       = regexp.MustCompile(`^(["A-Za-z0-9-_]+)\[(["A-Za-z0-9-_{}]+)]$`)
 	nameRegexDocker = regexp.MustCompile(`{{(.*?)}}`)
 )
 


felipe-avelar (approved):
LGTM

PR #3430, opened 2021-05-25
refactor(query): containers_run_with_low_uid rewrite

rogeriopeixotocx:


assets/queries/k8s/containers_run_with_low_uid/query.rego
```diff
@@ -1,156 +1,37 @@
...
-#if there are no containers to override low UID setting
-checkContainersOverride(spec, path, metadata, id) = result {
-	object.get(spec, "containers", "undefined") == "undefined"
+	to_number(securityContext.runAsUser) < 10000
 
 	result := {
-		"documentId": id,
-		"searchKey": sprintf("metadata.name={{%s}}.%sspec.securityContext.runAsUser", [metadata.name, path]),
+		"documentId": doc.id,
+		"searchKey": sprintf("%s.securityContext.runAsUser=%d", [common_lib.concat_path(path), securityContext.runAsUser]),
```
joaoReigota1, line 15: Since in line 11 we use to_number, I assume security context can be something other than an integer, maybe use to_number in the search key as well so it doesn't conflict with `%d`

joaoReigota1 (dismissed):
Other than a comment LGTM

assets/queries/k8s/containers_run_with_low_uid/query.rego
joaoReigota1:
```suggestion
		"searchKey": sprintf("%s.securityContext.runAsUser=%d", [common_lib.concat_path(path), to_number(securityContext.runAsUser)]),
```

felipe-avelar: I don't think we need to add `to_number` here, since it will only execute this line if `runAsUser` is a number, otherwise, it will fail on the first `to_number`, no?

felipe-avelar (approved):
LGTM

PR #3431, opened 2021-05-25
feat(cli): include queries filter

### LINKED ISSUE #3413: Add the option to select the queries to be run by KICS (include and not exclude)
**Description**
Add the possibility of providing `--queries-path` or `-p` multiple times, to be able to select queries by inclusion and not only exclusion.

rogeriopeixotocx:
Closes #3413

**Proposed Changes**
- add `--include-queries` or `-i` shorthand to include specific queries from a filesystem query source.
- compose `source.ExcludeQueries` and `source.IncludeQueries` inside the `source.QueriesSelectionFilter`
- adapt mocks and tests (WIP)


joaoReigota1 (approved):
LGTM

mukeshpilaniya:
Instant of include or excluding quires each times, we can do better by creating profiles. suppose a user wanted to exclude/ include only selected 100 queries than each time, user has to do the same thing which can lead to bad user experiences and bad developer practices as well. So instant of doing the same thing multiple times, if we can create a profile for this user and puts all selected quires into that profile than we can provide --profile options, which is a great in-terms of user experiences.

rogeriopeixotocx:
Hi @mukeshpilaniya could this `--profile` behaviour be emulated by having multiple kics.config files versioned somewhere? https://docs.kics.io/latest/configuration-file/

PR #3433, opened 2021-05-25
feat(query): add "Path Template is Empty" for OpenAPI

rafaela-soares:

**Proposed Changes**
- Added "Path Template is Empty" query for OpenAPI. It checks if there is some path template set as empty


rogeriopeixotocx (approved):
LGTM

PR #3447, opened 2021-05-26
feat(query): added "Path Is Ambiguous" query for OpenAPI

felipe-avelar:

**Proposed Changes**
- This query checks if there is a path that, without template, is equal, so it will generate two occurrences when this happens, one for each path


rogeriopeixotocx (approved):
LGTM

PR #3432, opened 2021-05-25
feat(report): added Gitlab SAST report

felipe-avelar:
Closes #1986 

**Proposed Changes**
- Added a new report format `glsast`, which will generate a file to be used with Gitlab SAST integration

I submit this contribution under the Apache-2.0 license.

Issue #1986, opened 2021-02-05
GitLab SAST Reports Integration


adam-moss:
**Is your feature request related to a problem? Please describe.**
It is frustrating when cli tool output has to be further processed or analysed to get up-front meaningful value, or cannot be integrated into existing development lifecycle tooling.

**Describe the solution you'd like**
It would be nice if a feature flag could be added to output the json report in GitLab SAST format, so that it is immediately consumable within the GitLab UI (merge request widget, Security Dashboard etc) through the standard report artifact mechanism in GitLab CI.

**Describe alternatives you've considered**
Processing the output directly with `jq` or equivalent to transform the output.

**Additional context**
Further information on the integrated reports formats included as part of GitLab CI are here: https://docs.gitlab.com/ee/development/integrations/secure.html#report.  SAST json definition is here: https://gitlab.com/gitlab-org/security-products/security-report-schemas/-/raw/v13.1.0/dist/sast-report-format.json

ruigomescx:
Hi @adam-moss. Thank you for your feedback. Currently we're only outputting the results to our internal .json format. 

At the moment we're also working on adding support for SARIF format:
http://docs.oasis-open.org/sarif/sarif/v2.0/csprd01/sarif-v2.0-csprd01.html

We will look into your suggestion, to see we can support it. Feel free to contribute also!
Thank you

adam-moss:
Hi @ruigomescx , there is actually an existing [issue within gitlab](https://gitlab.com/gitlab-org/gitlab/-/issues/118496) to support SARIF, so hopefully there will be a happy confluent 😁

PR #3460, opened 2021-05-27
fix(report): Update gitlab report fields to match proper formatting

theoretick:
**Proposed Changes**

Fixes a few issues in the GitLab SAST output format to conform with v13.1.0 of our SAST schema

- `Severity` should be titlecase
- `Category` is an enum of the report type

Relates to https://github.com/Checkmarx/kics/issues/1986

See discussion in https://github.com/Checkmarx/kics/pull/3432#issuecomment-849764618

### DIFF
--- a/docs/results.md
+++ b/docs/results.md
@@ -332,8 +332,8 @@ Gitlab SAST reports are sorted by severity (from high to info), following [Gitla
 	"vulnerabilities": [
 		{
 			"id": "32e763ac363dfee1ea972d951fb3de00f5f7a8d3f9f57b93e55e2d51957794a6",
-			"category": "Insecure Configurations",
-			"severity": "high",
+			"category": "sast",
+			"severity": "High",
 			"cve": "32e763ac363dfee1ea972d951fb3de00f5f7a8d3f9f57b93e55e2d51957794a6",
 			"scanner": {
 				"id": "keeping_infrastructure_as_code_secure",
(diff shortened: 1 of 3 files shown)


docs/results.md
```diff
@@ -332,8 +332,8 @@ Gitlab SAST reports are sorted by severity (from high to info), following [Gitla
 	"vulnerabilities": [
 		{
 			"id": "32e763ac363dfee1ea972d951fb3de00f5f7a8d3f9f57b93e55e2d51957794a6",
-			"category": "Insecure Configurations",
```
theoretick, line 335: It's a shame to lose this information, I'd suggest either appending it to either the name or message.

felipe-avelar: `Category` is really an enum type? [Looking at the schema](https://gitlab.com/gitlab-org/security-products/security-report-schemas/-/raw/v13.1.0/dist/sast-report-format.json), category is defined as follow:
```json
"category": {
  "type": "string",
  "minLength": 1,
  "description": "Describes where this vulnerability belongs (for example, SAST, Dependency Scanning, and so on)."
}
```

Also, am I using the right scheme?

theoretick: @felipe-avelar I believe it should be an enum and this is a bug, which is why I opened https://gitlab.com/gitlab-org/gitlab/-/issues/332054 to fix this and notify the team. It has been overlooked previously, but this essentially drives our groupings within the UI:

Screen Shot 2021-05-27 at 9 38 04 AM

We talk more about the field [on our integration page](https://docs.gitlab.com/ee/development/integrations/secure.html#category)

> Also, am I using the right schema?

You are! This is just a bug that I noticed on glancing over the diff, so it's on our end.

We should be able to better support different categorizations in the future beyond report types and are currently exploring doing so via CWE mapping instead: https://gitlab.com/gitlab-org/gitlab/-/issues/300978, but for now the grouping relies on the analysis type to differentiate between SAST, Dependency Scanning, DAST, Fuzzing, etc.

felipe-avelar: @theoretick Thanks for your response. ❤️ 
we will think where we should put category information.

Also, thanks for your contribution 😄

felipe-avelar (approved):
LGTM

docs/results.md
theoretick: Happy to help! This is actually my last day before taking a code-free vacation so I won't be around until around June 7th.

I'll poke my colleague to keep an eye on this one however if any follow-ups are needed after EOD

PR #3473, opened 2021-05-28
feat(query): add "Public and Private EC2 Share Role" for Terraform

rafaela-soares:

**Proposed Changes**
- Added "Public and Private EC2 Share Role" query for Terraform. It checks if a public and a private EC2 instances shares the same role through the same `iam_instance_profile`


rogeriopeixotocx (approved):
LGTM

PR #3487, opened 2021-05-31
feat(query): add "Lambda With Vulnerable Policy" for Terraform

rafaela-soares:

**Proposed Changes**
- Added "Lambda With Vulnerable Policy" query for Terraform. It checks if the Lambda policy defines wildcard in the attribute 'action'. It is not necessary to check the attribute 'principal' since Terraform does not support it as you can see [here](https://github.com/hashicorp/terraform-provider-aws/issues/2776)


rogeriopeixotocx (approved):
LGTM

PR #3492, opened 2021-05-31
feat(query): support OpenAPI 2.0/swagger with shared queries

felipe-avelar:

**Proposed Changes**
- Changed structure and tests for open API queries
- Moved circular reference from 3.0 to general


assets/queries/openAPI/3.0/security_undefined/query.rego
```diff
@@ -4,12 +4,17 @@ import data.generic.openapi as openapi_lib
 
 CxPolicy[result] {
 	doc := input.document[i]
-	openapi_lib.check_openapi(doc) != "undefined"
+	version := openapi_lib.check_openapi(doc)
+	version != "undefined"
 	object.get(doc, "security", "undefined") == "undefined"
+	searchKey := {
```
rogeriopeixotocx, line 10: if this starts to repeat itself I wonder if we should move this to the library

felipe-avelar: I think this will happen just in this query, because it needs to point the first line...

rogeriopeixotocx (approved):
LGTM

PR #3499, opened 2021-05-31
feat(query): add "S3 Bucket Public ACL Overridden By Public Access Block" for Terraform

rafaela-soares:

**Proposed Changes**
- Added "S3 Bucket Public ACL Overridden By Public Access Block" query for Terraform. It checks if the S3 bucket public access is overridden by S3 bucket Public Access Block when the following attributes are set to true - 'block_public_acls', 'block_public_policy', 'ignore_public_acls', and 'restrict_public_buckets'"


rogeriopeixotocx (approved):
LGTM

PR #3505, opened 2021-06-01
feat(query): moved path object empty query to general queries

felipe-avelar:
**Proposed Changes**
- Moved path object empty query to general queries for Open API


rogeriopeixotocx (approved):
LGTM

PR #3506, opened 2021-06-01
feat(query): moved minimum greater than maximum query to general

felipe-avelar:
**Proposed Changes**
- Moved minimum greater than maximum query to general for open api


rogeriopeixotocx (approved):
LGTM

PR #3508, opened 2021-06-01
feat(query): add "SNS Topic Encrypted With AWS Managed Key" for Terraform

rafaela-soares:

**Proposed Changes**
- Added "SNS Topic Encrypted With AWS Managed Key" query for Terraform. It checks if the SNS Topic is encrypted with AWS managed key (`alias/aws/sns`)

### DIFF
--- a/assets/libraries/terraform/library.rego
+++ b/assets/libraries/terraform/library.rego
@@ -464,7 +464,6 @@ uses_aws_managed_key(key, awsManagedKey) {
 	key == awsManagedKey
 } else {
 	keyName := split(key, ".")[2]
-	kms := input.document[z].data.aws_kms_key[kmsName]
-	keyName == kmsName
+	kms := input.document[z].data.aws_kms_key[keyName]
 	kms.key_id == awsManagedKey
 }
--- a/assets/libraries/terraform/library.rego
+++ b/assets/libraries/terraform/library.rego
@@ -459,3 +459,12 @@ has_wildcard(statement, typeAction) {
 } else {
 	check_actions(statement, typeAction)
 }
+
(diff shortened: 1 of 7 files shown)


rogeriopeixotocx (approved):
LGTM

aminvielledebatAtBedrock:
Hi @rafaela-soares !

We (with @jycamier :heart:) are hurting with this rule, and several others about encryption, in our scans because we used to use aws kms managed key.

Could you explain us why this rule is implemented, and why it's tagged as `HIGH` please ?

rafaela-soares:
Hello @aminvielledebatAtBedrock!

Glad to see you here! Thank you so much for using KICS and reaching us.
My regards to @jycamier!

From my understanding, KMS Customer Master Keys (CMKs) should be used instead of AWS Managed Keys, because when using AWS Managed Keys, the default keys will be used since there are no Customer Managed Keys (which do not provide sufficient control over encryption and decryption). KMS Customer Master Keys (CMKs) provide more control since you can see the key policy, for example.

In order to give you a more detailed explanation, I will ask our AppSec team to join this thread. Additionally to that, I will request description improvement for this kind of query.

Hope to continue to see you all around😊

rjegoncalves:
Hi @aminvielledebatAtBedrock!

Rafaela's explanation is very much on point. Another relevant note here is that using Customer Managed Keys (CMKs) will provide full control to the customer over the encrypted data - which is needed to meet some regulatory compliance requirements.

jycamier:
Hi @rafaela-soares and @rjegoncalves,

Thanks for your answers !

Indeed ! We find more information about it. Maybe we're concerned.
It's a PEBKAC ! Sorry for the noise !

PR #3513, opened 2021-06-01
feat(query):  moved operation id not unique query to general

felipe-avelar:
**Proposed Changes**
-  Moved operation id not unique query to general for open api 2.0


rogeriopeixotocx (approved):
LGTM

PR #3515, opened 2021-06-01
feat(query): add "SQS VPC Endpoint Without DNS Resolution" for Terraform

rafaela-soares:

**Proposed Changes**
- Added "SQS VPC Endpoint Without DNS Resolution" query for Terraform. It checks if the SQS VPC Endpoint has DNS resolution disabled ('enable_dns_support' set to false). According to the official AWS documentation, the attribute `enableDnsSupport` "indicates whether the DNS resolution is supported", as you can see [here](https://docs.aws.amazon.com/vpc/latest/userguide/vpc-dns.html#vpc-dns-support)


rogeriopeixotocx (approved):
LGTM

PR #3646, opened 2021-06-14
feat(query): Moved Invalid OAuth2 Token URL query to general

joaoReigota1:

**Proposed Changes**
- Moved Invalid OAuth2 Token URL query to general


felipe-avelar (changes requested):
Since query rules are specific for each version, I think this could be two separated query files, one on 3.0 and the other on 2.0

felipe-avelar (approved):
LGTM

PR #3651, opened 2021-06-14
feat(query): updating queries related to "Parameter Object"

rafaela-soares:
Signed-off-by: Rafaela Soares <rafaela.soares@checkmarx.com>

**Proposed Changes**
- Updating queries related to "Parameter Object". (The `concat_path` function returns an empty string for the global "Parameter Object", which makes the search key invalid) 
- Deleted rule related to "definitions" since this object just has Schema Object and not Parameter Object


felipe-avelar (changes requested):
I think it would be nice if some test cases were added to validate the change on queries

rogeriopeixotocx:
Pending analysis on why OPA is yielding a different number of results sometimes,
Seems that after the update to v0.49.4 the output of a query has become non-deterministic.
We have to check if this has something to do with this issue: https://github.com/open-policy-agent/opa/issues/3546
On worst-case scenario we should revert to OPA v0.28.0

rogeriopeixotocx (approved):
LGTM

PR #3664, opened 2021-06-15
feat(query): Added Security Definitions Using Password Flow query to Swagger

joaoReigota1:

**Proposed Changes**
- Added Security Definitions Using Password Flow query to Swagger

### DIFF
--- a/assets/queries/openAPI/2.0/security_definitions_using_password_flow/query.rego
+++ b/assets/queries/openAPI/2.0/security_definitions_using_password_flow/query.rego
@@ -6,8 +6,8 @@ CxPolicy[result] {
 	doc := input.document[i]
 	openapi_lib.check_openapi(doc) == "2.0"
 
-	sec_def := doc.securityDefinitions[key]
-	sec_def.flow == "password"
+	secDef := doc.securityDefinitions[key]
+	secDef.flow == "password"
 
 	result := {
 		"documentId": doc.id,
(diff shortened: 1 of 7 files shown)


assets/queries/openAPI/2.0/security_definitions_using_password_flow/query.rego
```diff
@@ -0,0 +1,19 @@
+package Cx
+
+import data.generic.openapi as openapi_lib
+
+CxPolicy[result] {
+	doc := input.document[i]
+	openapi_lib.check_openapi(doc) == "2.0"
+
+	sec_def := doc.securityDefinitions[key]
+	sec_def.flow == "password"
```
felipe-avelar, line 10: I think it should be `secDef`. Also, I think we should check `secDef.type == "oauth2"`

felipe-avelar (changes requested):
Please, check comments

assets/queries/openAPI/2.0/security_definitions_using_password_flow/query.rego
joaoReigota1: the name 'sec_def' is to follow REGO naming convention.
I think it is not necessary to check if 'type' is "oauth" since the flow field Is only used for type "oauth", meaning that it won't exist in other type making the query fail on line 10

felipe-avelar: I agree with you about `type`, but for variables, we use camel case style in rego files

felipe-avelar (approved):
LGTM

PR #3704, opened 2021-06-18
feat(query): added Multi 'collectionformat' Not Valid For 'in' Parameter query for swagger

felipe-avelar:

**Proposed Changes**
- Added Multi 'collectionformat' Not Valid For 'in' Parameter query for swagger

### DIFF
--- a/assets/queries/openAPI/2.0/multi_collectionformat_not_valid_in_parameter/query.rego
+++ b/assets/queries/openAPI/2.0/multi_collectionformat_not_valid_in_parameter/query.rego
@@ -1,0 +1,24 @@
+package Cx
+
+import data.generic.openapi as openapi_lib
+
+CxPolicy[result] {
+	doc := input.document[i]
+	openapi_lib.check_openapi(doc) == "2.0"
+
+	[path, value] := walk(doc)
+	param := value.parameters[n]
+	param.collectionFormat == "multi"
+	param.in != "query"
+	param.in != "formData"
+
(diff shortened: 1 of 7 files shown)


assets/queries/openAPI/2.0/multi_collectionformat_not_valid_in_parameter/query.rego
```diff
@@ -0,0 +1,24 @@
...
+
+CxPolicy[result] {
+	doc := input.document[i]
+	openapi_lib.check_openapi(doc) == "2.0"
+
+	[path, value] := walk(doc)
+	param := value.parameters[n]
+	param.collectionFormat == "multi"
+	param.in != "query"
+	param.in != "formData"
```
rogeriopeixotocx, line 13: shouldn't it be `param.in != "query"` OR `param.in != "formData"` ? The way it is looks like an `AND` to me

rogeriopeixotocx (changes requested):

assets/queries/openAPI/2.0/multi_collectionformat_not_valid_in_parameter/query.rego
felipe-avelar: No, this query should not be positive if in is query or formData, so we want to dispatch an issue if in is not query AND not formData

rogeriopeixotocx (approved):

PR #3706, opened 2021-06-18
feat(query): added 'Schema with 'additionalProperties' set as Boolean' for swagger

rafaela-soares:
Signed-off-by: Rafaela Soares <rafaela.soares@checkmarx.com>

**Proposed Changes**
- Added 'Schema with 'additionalProperties' set as Boolean' for openAPI 2.0

### DIFF
--- a/assets/queries/openAPI/2.0/schema_with_additional_properties_set_as_boolean/metadata.json
+++ b/assets/queries/openAPI/2.0/schema_with_additional_properties_set_as_boolean/metadata.json
@@ -3,7 +3,7 @@
   "queryName": "Schema with 'additionalProperties' set as Boolean",
   "severity": "INFO",
   "category": "Best Practices",
-  "descriptionText": "The value of 'additionalProperties' should be set as object instead of boolean since openAPI 2.0 does not support boolean value for it",
+  "descriptionText": "The value of 'additionalProperties' should be set as object instead of boolean, since swagger 2.0 does not support boolean value for it",
   "descriptionUrl": "https://swagger.io/specification/v2/#schema-object",
   "platform": "OpenAPI"
 }
--- a/assets/queries/openAPI/2.0/schema_with_additional_properties_set_as_boolean/metadata.json
+++ b/assets/queries/openAPI/2.0/schema_with_additional_properties_set_as_boolean/metadata.json
@@ -1,0 +1,9 @@
+{
+  "id": "3a01790c-ebee-4da6-8fd3-e78657383b75",
+  "queryName": "Schema with 'additionalProperties' set as Boolean",
+  "severity": "INFO",
+  "category": "Best Practices",
+  "descriptionText": "The value of 'additionalProperties' should be set as object instead of boolean since openAPI 2.0 does not support boolean value for it",
+  "descriptionUrl": "https://swagger.io/specification/v2/#schema-object",
+  "platform": "OpenAPI"
+}
--- a/assets/queries/openAPI/2.0/schema_with_additional_properties_set_as_boolean/query.rego
+++ b/assets/queries/openAPI/2.0/schema_with_additional_properties_set_as_boolean/query.rego
@@ -1,0 +1,35 @@
+package Cx
+
+import data.generic.openapi as openapi_lib
+
+CxPolicy[result] {
+	doc := input.document[i]
+	openapi_lib.check_openapi(doc) == "2.0"
+
+	[path, value] := walk(doc)
+	is_boolean(value.schema.additionalProperties)
+
+	result := {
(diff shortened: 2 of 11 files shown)


assets/queries/openAPI/2.0/schema_with_additional_properties_set_as_boolean/metadata.json
```diff
@@ -0,0 +1,9 @@
+{
+  "id": "3a01790c-ebee-4da6-8fd3-e78657383b75",
+  "queryName": "Schema with 'additionalProperties' set as Boolean",
+  "severity": "INFO",
+  "category": "Best Practices",
+  "descriptionText": "The value of 'additionalProperties' should be set as object instead of boolean since openAPI 2.0 does not support boolean value for it",
```
felipe-avelar, line 6:
```suggestion
  "descriptionText": "The value of 'additionalProperties' should be set as schema object instead of boolean, since swagger 2.0 does not support boolean value for it",
```

assets/queries/openAPI/2.0/schema_with_additional_properties_set_as_boolean/query.rego
```diff
@@ -0,0 +1,19 @@
+package Cx
+
+import data.generic.openapi as openapi_lib
+
+CxPolicy[result] {
+	doc := input.document[i]
+	openapi_lib.check_openapi(doc) == "2.0"
+
+	[path, value] := walk(doc)
+	is_boolean(value.schema.additionalProperties)
```
felipe-avelar, line 10: I think here can be just `value.additionalProperties`, no? Since there is `additionalProperty` only for schema objects... Also, if we checks `value.schema.additionalProperties`, we won't check `definitions` object, since it does not have a 'schema' property...

```suggestion
	is_boolean(value.additionalProperties)
```

assets/queries/openAPI/2.0/schema_with_additional_properties_set_as_boolean/query.rego
```diff
@@ -0,0 +1,19 @@
...
+
+	result := {
+		"documentId": doc.id,
+		"searchKey": sprintf("%s.schema.additionalProperties", [openapi_lib.concat_path(path)]),
```
felipe-avelar, line 14: If above sugestion is accepted, I think we should change searchKey too...
```suggestion
		"searchKey": sprintf("%s.additionalProperties", [openapi_lib.concat_path(path)]),
```

assets/queries/openAPI/2.0/schema_with_additional_properties_set_as_boolean/query.rego
```diff
@@ -0,0 +1,19 @@
...
+		"issueType": "IncorrectValue",
+		"keyExpectedValue": "'additionalProperties' is not set as a boolean value",
```
felipe-avelar, line 16: I think it is more clear, if we give expected value as follow:
```suggestion
		"keyExpectedValue": "'additionalProperties' is set as a schema object",
```

assets/queries/openAPI/2.0/schema_with_additional_properties_set_as_boolean/test/positive1.json
```diff
@@ -0,0 +1,35 @@
...
+              "required": [
+                "name"
+              ],
+              "additionalProperties": false
+            }
+          }
+        }
+      }
+    }
+  }
```
felipe-avelar, line 34: I think we could add a test case where `definitions` have a schema with invalid `additionalProperties`, what do you think?

felipe-avelar (changes requested):
Please, check comments

assets/queries/openAPI/2.0/schema_with_additional_properties_set_as_boolean/query.rego
rafaela-soares: I thought it does not make sense to check `definitions object` since it is unusual for a Definitions Object to refer to another Definitions Object. What do you think?

rafaela-soares: In the first instance, I thought that does not make sense to check `definitions object`, since it is unusual for a Definitions Object to refer to another Definitions Object. However, on second thought, it makes sense since `additionalProperty` could be something else than `$ref`:
```
additionalProperties:
  type: string
```

felipe-avelar (approved):
LGTM

rogeriopeixotocx (approved):
LGTM

rogeriopeixotocx (approved):

PR #3712, opened 2021-06-21
feat(query): Added BasePath With Wrong Format query to Swagger

joaoReigota1:

**Proposed Changes**
- Added BasePath With Wrong Format query to Swagger

The 'basePath' value format must match the pattern '^/'

### DIFF
--- a/assets/queries/openAPI/2.0/basepath_with_wrong_format/query.rego
+++ b/assets/queries/openAPI/2.0/basepath_with_wrong_format/query.rego
@@ -1,0 +1,18 @@
+package Cx
+
+import data.generic.openapi as openapi_lib
+
+CxPolicy[result] {
+	doc := input.document[i]
+	openapi_lib.check_openapi(doc) == "2.0"
+
+	not regex.match("^/", doc.basePath)
+
+	result := {
+		"documentId": doc.id,
+		"searchKey": sprintf("basePath={{%s}}", [doc.basePath]),
+		"issueType": "IncorrectValue",
+		"keyExpectedValue": "'basePath' value format matches the pattern '^/'",
+		"keyActualValue": "'basePath' value format doesn't match the pattern '^/'",
+	}
+}
--- a/assets/queries/openAPI/2.0/basepath_with_wrong_format/query.rego
+++ b/assets/queries/openAPI/2.0/basepath_with_wrong_format/query.rego
@@ -12,7 +12,7 @@ CxPolicy[result] {
 		"documentId": doc.id,
 		"searchKey": sprintf("basePath={{%s}}", [doc.basePath]),
 		"issueType": "IncorrectValue",
-		"keyExpectedValue": "'basePath' value format matches the pattern '^/'",
-		"keyActualValue": "'basePath' value format doesn't match the pattern '^/'",
+		"keyExpectedValue": "'basePath' value matches the pattern '^/'",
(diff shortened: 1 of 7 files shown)


assets/queries/openAPI/2.0/basepath_with_wrong_format/query.rego
```diff
@@ -0,0 +1,18 @@
+package Cx
+
+import data.generic.openapi as openapi_lib
+
+CxPolicy[result] {
+	doc := input.document[i]
+	openapi_lib.check_openapi(doc) == "2.0"
+
+	not regex.match("^/", doc.basePath)
```
rogeriopeixotocx, line 9: shouldn't this be:

```suggestion
	not regex.match("^/.*", doc.basePath)
```

/anything/else ??

assets/queries/openAPI/2.0/basepath_with_wrong_format/query.rego
```diff
@@ -0,0 +1,18 @@
...
+
+	result := {
+		"documentId": doc.id,
+		"searchKey": sprintf("basePath={{%s}}", [doc.basePath]),
```
felipe-avelar, line 13: Since `basePath` is just a string, not an array, I think search key can be just `basePath`, as following:
```suggestion
		"searchKey": "basePath",
```

rogeriopeixotocx (changes requested):

assets/queries/openAPI/2.0/basepath_with_wrong_format/query.rego
```diff
@@ -0,0 +1,18 @@
...
+		"issueType": "IncorrectValue",
+		"keyExpectedValue": "'basePath' value format matches the pattern '^/'",
+		"keyActualValue": "'basePath' value format doesn't match the pattern '^/'",
```
felipe-avelar, line 16:
```suggestion
		"keyExpectedValue": "'basePath' value matches the pattern '^/'",
		"keyActualValue": "'basePath' value doesn't match the pattern '^/'",
```

felipe-avelar (changes requested):
Please, check comments

assets/queries/openAPI/2.0/basepath_with_wrong_format/query.rego
felipe-avelar: No, currently regex is correct, as you can see in [specification json's schema](http://23.22.16.221/v2/schema.json)

assets/queries/openAPI/2.0/basepath_with_wrong_format/query.rego
joaoReigota1: This is to avoid cases like comments or names that contain basePath where the search key will point to the wrong lines, for example:
```
1: {
2:  "swagger": "2.0",
3:  "info": {
4:    "title": "basePath API Overview",
5:    "version": "1.0.0"
6:  },
7: "basePath": "/api"
8:}
```
The search key, in this case, will point to line 4 which is not correct.
This is more to avoid further problems

assets/queries/openAPI/2.0/basepath_with_wrong_format/query.rego
joaoReigota1: As Felipe said above I followed the specification JSON's schema

felipe-avelar (approved):
LGTM

rogeriopeixotocx (approved):
LGTM

PR #3715, opened 2021-06-21
feat(query): Added Operation Object Parameters With 'body' And 'formatData' loactions query to Swagger

joaoReigota1:

**Proposed Changes**
- Added Operation Object Parameters With 'body' And 'formatData' locations query to Swagger

Operation object parameters should not have both 'body' and 'formatData' locations


assets/queries/openAPI/2.0/operation_object_parameters_with_body_and_formatdata/query.rego
```diff
@@ -0,0 +1,24 @@
...
+
+CxPolicy[result] {
+	doc := input.document[i]
+	openapi_lib.check_openapi(doc) == "2.0"
+
+	[path, value] := walk(doc)
+	count(path) != 0
+	param := value.parameters
+	body := [x | p := param[l]; p.in == "body"; x := p.in]
+	formatData := [x | p := param[l]; p.in == "formatData"; x := p.in]
```
felipe-avelar, line 13: I think we could use `_` here, since `l` is not used...
```suggestion
	body := [x | p := param[_]; p.in == "body"; x := p.in]
	formatData := [x | p := param[_]; p.in == "formatData"; x := p.in]
```

assets/queries/openAPI/2.0/operation_object_parameters_with_body_and_formatdata/query.rego
```diff
@@ -0,0 +1,24 @@
+package Cx
+
+import data.generic.openapi as openapi_lib
+
+CxPolicy[result] {
+	doc := input.document[i]
+	openapi_lib.check_openapi(doc) == "2.0"
+
+	[path, value] := walk(doc)
+	count(path) != 0
```
felipe-avelar, line 10: I think "greater than" is more clear instead of "not equal"
```suggestion
	count(path) > 0
```

felipe-avelar (changes requested):
Please, check comments

assets/queries/openAPI/2.0/operation_object_parameters_with_body_and_formatdata/test/negative1.json
```diff
@@ -0,0 +1,51 @@
...
+      }
+    }
+  },
+  "parameters": {
+    "limitParam": {
+      "name": "limit",
+      "in": "formatData",
+      "description": "max records to return",
+      "required": true,
+      "schema": {
```
rafaela-soares, line 46: `schema` is not a valid value when `"in": "formatData"`. It is only a valid value when `"in": "body"`. The same happens in the other samples

rafaela-soares (changes requested):

felipe-avelar (approved):
LGTM

PR #3733, opened 2021-06-22
feat(query): add Operation Using Implicit Flow query to Swagger

joaoReigota1:

**Proposed Changes**
- Added Operation Using Implicit Flow query to Swagger

Operation Object should not use implicit flow


rogeriopeixotocx (approved):
LGTM

PR #3737, opened 2021-06-22
feat(engine): extract zip file passed as --path

### LINKED ISSUE #3689: Zip file support
for the Kics as a Service portal - we need support for ZIP file to be given as a parameter instead of a directory path 

It should just unzip it and then use the directory path normally

We can do this on the Lambda but it makes less sense then internal support inside KICS.

**Additional context**

rogeriopeixotocx:
Closes #3689

**Proposed Changes**
- Add helpers file `extract.go` to deal with zip files


internal/console/helpers/extract_handler.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		return filenames, err
+	}
+
+	fi, err := f.Stat()
+	if err != nil {
+		return filenames, err
+	}
+
+	r, err := zip.NewReader(f, fi.Size())
```
felipe-avelar, line 92: I think we could rename `r` to be more clear, like `zipReader` or something like this...

internal/console/helpers/extract_handler.go
```diff
@@ -0,0 +1,150 @@
...
+	MaxZipChunkSize = 1024
+)
+
+// PathExtractionMap - relates temporary dir with original path
+var PathExtractionMap map[string]string
+
+// CheckAndExtractZip - verifies if a given absolute path is a zip file
+// and if so extracts its contents to a temporary file
+func CheckAndExtractZip(absPath string) (string, error) {
+	f, err := os.Open(absPath)
```
felipe-avelar, line 26: I don't see where `f` is closed, also we are opening same path on line 82?

internal/console/helpers/extract_handler.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	r, err := zip.NewReader(f, fi.Size())
+	if err != nil {
+		return filenames, err
+	}
+
+	defer f.Close()
+
+	for _, f := range r.File {
```
felipe-avelar, line 99: I think we should change `f` here, since it is shadowing file declaration... Also, it maybe get confusing when reading

internal/console/helpers/extract_handler.go
```diff
@@ -0,0 +1,150 @@
...
+		}
+
+		outFile, err := os.OpenFile(fpath,
+			os.O_WRONLY|os.O_CREATE|os.O_TRUNC,
+			f.Mode())
+		if err != nil {
+			return filenames, err
+		}
+
+		rc, err := f.Open()
```
felipe-avelar, line 127: I think we could rename `rc` to be more clear...

internal/console/helpers/extract_handler.go
```diff
@@ -0,0 +1,150 @@
...
+				}
+
+				outFile.Close()
+				rc.Close()
+				return filenames, err
+			}
+		}
+
+		outFile.Close()
+		rc.Close()
```
felipe-avelar, line 146: I think we could change this to:
```suggestion
      defer outFile.Close()
		defer rc.Close()
		for {
			_, err = io.CopyN(outFile, rc, MaxZipChunkSize)
			if err != nil {
				if err == io.EOF {
					break
				}
				return filenames, err
			}
		}
```

internal/console/scan.go
```diff
@@ -527,6 +520,22 @@ func createService(inspector *engine.Inspector,
 	return services, nil
 }
 
+func extractPaths(paths []string) (extectedPaths []string, pathExtractionMap map[string]string, err error) {
```
felipe-avelar, line 523: I think this would make more sense to be on a source provider, maybe we should create a source provider for zip files, what do you think?

pkg/model/summary.go
```diff
@@ -69,8 +72,55 @@ type Summary struct {
...
 }
 
+// PathParameters - structure wraps the required fields for temporary path translation
+type PathParameters struct {
+	ScannedPaths      []string
+	PathExtractionMap map[string]string
+}
+
+func getRelativePath(basePath, filePath string) string {
+	var rtn string
```
felipe-avelar, line 82: I think we could rename `rtn` to be more clear... Maybe something like `resultPath`

felipe-avelar (changes requested):
Please, check comments. Also, create changes on report packages. 😊

felipe-avelar (dismissed):
LGTM

felipe-avelar (approved):
LGTM

PR #3750, opened 2021-06-23
build(makefile): enhance make clean task

rogeriopeixotocx:
**Proposed Changes**
- `make clean` should remove all non-versioned output files

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,10 @@ clean: ## remove files created during build
 	rm -rf vendor
 	rm -f coverage.*
 	rm -f cover.out
+	rm -rf *.log
 	rm -rf **/*.log
+	rm -f results.*
+	rm -f e2e/output
 
 .PHONY: mod-tidy
 mod-tidy: ## go mod tidy - download and cleanup modules


felipe-avelar (dismissed):
LGTM

Makefile
```diff
@@ -18,7 +18,10 @@ clean: ## remove files created during build
 	rm -rf vendor
 	rm -f coverage.*
 	rm -f cover.out
+	rm -rf *.log
 	rm -rf **/*.log
+	rm -f results.*
```
felipe-avelar, line 23: If we use make, it will delete all logs and results named files?

rogeriopeixotocx: all logs in all subdirectories, and only results in the root directory

felipe-avelar: We really want this? I'm concerned, because, maybe, we want to preserve those files for future debugging or checking, no?

rogeriopeixotocx: then you should not run make clean

felipe-avelar (approved):
LGTM

PR #3771, opened 2021-06-28
feat(query): Added query Additional Properties Too Permissive for open api

felipe-avelar:

**Proposed Changes**
- Added query Additional Properties Too Permissive for open api


joaoReigota1 (changes requested):
I think that we should add a rule that checks when it is `allOf` that `additionalProperties` is not set to false

felipe-avelar:
> I think that we should add a rule that checks when it is `allOf` that `additionalProperties` is not set to false

This will be resolved by another query

joaoReigota1 (approved):
LGTM

rogeriopeixotocx (approved):
LGTM

PR #3788, opened 2021-06-30
fix(report): Fix path always abs with windows and zip extraction

joaoReigota1:

**Proposed Changes**
- Fixed path always abs with windows and zip extraction


felipe-avelar (approved):
LGTM

PR #3947, opened 2021-07-29
feat(query): Added SQL Alert Policy Without Emails query to ARM

joaoReigota1:

**Proposed Changes**
- Added SQL Alert Policy Without Emails query to ARM

SQL Database Server should contain emails to be notified in the event of a Security Alert


rogeriopeixotocx (approved):
LGTM

PR #3961, opened 2021-07-29
feat(query): add 'Storage Account Allows Unsecure transfer' for ARM

rogeriopeixotocx:


assets/queries/azureResourceManager/storage_account_allows_unsecure_transfer/query.rego
```diff
@@ -0,0 +1,47 @@
...
+		"keyActualValue": "resource with type 'Microsoft.Storage/storageAccounts' doesn't have 'supportsHttpsTrafficOnly' property defined",
+	}
+}
+
+CxPolicy[result] {
+	doc := input.document[i]
+	[path, value] = walk(doc)
+
+	value.type == "Microsoft.Storage/storageAccounts"
+	value.properties.supportsHttpsTrafficOnly == false
```
felipe-avelar, line 30: I think here we could check if `value.properties.supportsHttpsTrafficOnly` is not true, since `supportsHttpsTrafficOnly` is not a required field. Also, maybe it should checks if it is undefined to define the search...
```suggestion
	not value.properties.supportsHttpsTrafficOnly
	skValue := is_undefined(value)
```

felipe-avelar (changes requested):
Please, check comments. Also, maybe a sample with api version > 2019 and missing `supportsHttpsTrafficOnly` would be great

assets/queries/azureResourceManager/storage_account_allows_unsecure_transfer/query.rego
rogeriopeixotocx: I've made some adjustments to be more clear and added one more sample but the only way for it to trigger the query if API version >= 2019 is by defining `properties.supportsHttpsTrafficOnly` to false.. since the default is true

felipe-avelar (approved):
LGTM

PR #3963, opened 2021-07-29
feat(query): Added Storage Blob Service Container With Public Access query for ARM

rafaela-soares:

**Proposed Changes**
- Added Storage Blob Service Container With Public Access query for ARM


felipe-avelar (approved):
LGTM

rogeriopeixotocx (approved):

PR #4001, opened 2021-08-05
refactor(query): 'object.get' function for cloudFormation

tiagocruz20:
Signed-off-by: Tiago Cruz tiago.cruz@checkmarx.com

**Proposed Changes**
- refactoring `object.get` function to `valid.key` function

### DIFF
--- a/assets/queries/cloudFormation/stack_retention_disabled/query.rego
+++ b/assets/queries/cloudFormation/stack_retention_disabled/query.rego
@@ -40,9 +40,8 @@ CxPolicy[result] {
 	resource.Type == "AWS::CloudFormation::StackSet"
 
     autoDeployment := resource.Properties.AutoDeployment
-	common_lib.valid_key(autoDeployment, "Enabled")
 
-    not autoDeployment.Enabled
+    autoDeployment.Enabled == false
 
 	result := {
 		"documentId": input.document[i].id,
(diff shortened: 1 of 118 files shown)


rogeriopeixotocx (dismissed):

assets/queries/cloudFormation/stack_retention_disabled/query.rego
```diff
@@ -37,10 +39,10 @@ CxPolicy[result] {
 	resource := input.document[i].Resources[name]
 	resource.Type == "AWS::CloudFormation::StackSet"
 
-	object.get(resource.Properties, "AutoDeployment", "undefined") != "undefined"
+	common_lib.valid_key(resource.Properties, "AutoDeployment")
```
rafaela-soares, line 42: `common_lib.valid_key(resource.Properties, "AutoDeployment")` is unnecessary. Same for lines 45 and 81

rafaela-soares (changes requested):
Good job, @tiagocruz20! 🚀 

Overall, LGTM. I suggest small changes related to cases when it is unnecessary to check if a key is valid. These cases are related to key value comparison. For example, if we want to compare if `properties.Encrypted == false`, it is unnecessary to check if ` common_lib.valid_key(properties, "Encrypted")` since the comparison itself already assumes that the key is defined and not null

rafaela-soares (approved):
LGTM

PR #4029, opened 2021-08-12
feat(query): updated "IAM Policies With Full Privileges" query for Terraform

rafaela-soares:
**Proposed Changes**
- Updated "IAM Policies With Full Privileges" query for Terraform. The resource `aws_iam_policy_attachment` does not have the attribute `policy` but actually `policy_arn`, which does not fit this query


felipe-avelar (approved):
LGTM

rogeriopeixotocx (approved):
great finding

PR #4106, opened 2021-08-24
feat(query): Updated "S3 Bucket Policy Accepts HTTP Requests" for Terraform

rafaela-soares:
**Proposed Changes**
- Correcting "S3 Bucket Policy Accepts HTTP Requests" query. Should be `statement.Condition.Bool["aws:SecureTransport"] == "false"` instead of `statement.Condition.Bool["aws:SecureTransport"] == false`


rogeriopeixotocx (approved):
LGTM

PR #4138, opened 2021-08-31
fix(query): Updated query query to check if $ref is defined

felipe-avelar:

**Proposed Changes**
- Updated query to check if `$ref` is defined, if that is the case, should be negative


rogeriopeixotocx (approved):
LGTM

PR #4162, opened 2021-09-06
refactor(query): EBS Default Encryption Disabled

tiagocruz20:

### DIFF
--- a/assets/queries/terraform/aws/ebs_default_encryption_disabled/metadata.json
+++ b/assets/queries/terraform/aws/ebs_default_encryption_disabled/metadata.json
@@ -1,7 +1,7 @@
 {
   "id": "3d3f6270-546b-443c-adb4-bb6fb2187ca6",
   "queryName": "EBS Default Encryption Disabled",
-  "severity": "HIGH",
+  "severity": "MEDIUM",
   "category": "Encryption",
   "descriptionText": "EBS Encryption should be enabled",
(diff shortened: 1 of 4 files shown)


assets/queries/terraform/aws/ebs_default_encryption_disabled/metadata.json
```diff
@@ -1,7 +1,7 @@
 {
   "id": "3d3f6270-546b-443c-adb4-bb6fb2187ca6",
   "queryName": "EBS Default Encryption Disabled",
-  "severity": "HIGH",
+  "severity": "MEDIUM",
```
rogeriopeixotocx, line 4:
```suggestion
  "severity": "HIGH",
```
nevermind about my previous suggestion let's keep this at HIGH

rogeriopeixotocx (changes requested):

rogeriopeixotocx (changes requested):
We don't need the OR condition to verify if the enabled is defined because the default is true.. the old query worked just fine https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ebs_encryption_by_default

rogeriopeixotocx:
We don't need the OR condition to verify if the enabled is defined because the default is true.. the old query worked just fine https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ebs_encryption_by_default

PR #4171, opened 2021-09-08
feat(query): added Security Contact Email for Terraform

tiagocruz20:
**Proposed Changes**
- added a query to check if email is set in Security Contact

### DIFF
--- a/assets/queries/terraform/azure/security_contact_email/metadata.json
+++ b/assets/queries/terraform/azure/security_contact_email/metadata.json
@@ -1,0 +1,11 @@
+{
+  "id": "34664094-59e0-4524-b69f-deaa1a68cce3",
+  "queryName": "Security Contact Email",
+  "severity": "MEDIUM",
+  "category": "Best Practises",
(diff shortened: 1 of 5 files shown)


rogeriopeixotocx (changes requested):
Since the email field is required by Terraform.
Don't think it makes sense to have a query to check if this email field is missing.

Opinions? @JoaoDuarteGomes @felipe-avelar @rafaela-soares

felipe-avelar:
> Since the email field is required by Terraform.
> Don't think it makes sense to have a query to check if this email field is missing.
> 
> Opinions? @JoaoDuarteGomes @felipe-avelar @rafaela-soares

I agree with you, for me doesn't makes sense either...

JoaoDuarteGomes:
It's up to you whether you want the query or not.
If this is an issue that can both be correctly detected statically and prevents the plan from being built, this is a legitimate query.

rogeriopeixotocx (approved):

PR #4185, opened 2021-09-10
feat(query): added Azure Front Door WAF Disabled for Terraform

tiagocruz20:
**Proposed Changes**
- added a query to check if WAF is enabled for Azure Front Door


felipe-avelar (approved):
LGTM

PR #4209, opened 2021-09-16
fix: added negative samples + avoiding FP's for Passwords and Secrets Mechanism 

rafaela-soares:
**Proposed Changes**
- Added negative samples + avoiding FP's for Passwords and Secrets Mechanism 


rogeriopeixotocx (approved):
LGTM

PR #4210, opened 2021-09-16
feat(query): support of Terraform module for "SQS VPC Endpoint Without DNS Resolution"

rafaela-soares:
**Proposed Changes**
- Support of Terraform module for "SQS VPC Endpoint Without DNS Resolution"


rogeriopeixotocx (approved):
LGTM

PR #4212, opened 2021-09-16
feat(query): support of Terraform module for "Instance With No VPC"

rafaela-soares:
**Proposed Changes**
- Support of Terraform module for "Instance With No VPC"


rogeriopeixotocx (approved):
LGTM

PR #4225, opened 2021-09-17
feat(query): support of Terraform modules for "RDS Without Logging"

tiagocruz20:
**Proposed Changes**
- support of Terraform modules for "RDS Without Logging"


felipe-avelar (changes requested):
I think we could split `postive.tf` into `positive1.tf` and `positive2.tf`

rogeriopeixotocx (approved):
LGTM

PR #4230, opened 2021-09-20
feat(query): support of Terraform modules for "User Data Shell Script…

tiagocruz20:
… Is Encoded"

**Proposed Changes**
- support of Terraform modules for "User Data Shell Script Is Encoded"


rogeriopeixotocx (approved):
LGTM

PR #4251, opened 2021-09-22
feat(query): support of Terraform modules for "S3 Bucket Allows List Action From All Principals"

rafaela-soares:
**Proposed Changes**
- Support of Terraform modules for "S3 Bucket Allows List Action From All Principals"


rogeriopeixotocx (approved):
LGTM

PR #4255, opened 2021-09-22
feat(query): support of Terraform module for "S3 Bucket With Ignore Public ACL"

joaoReigota1:

**Proposed Changes**
- support of Terraform module for "S3 Bucket With Ignore Public ACL"


assets/queries/terraform/aws/s3_bucket_with_ignore_public_acl/query.rego
```diff
@@ -1,5 +1,7 @@
 package Cx
 
+import data.generic.common as common_lib
+
 CxPolicy[result] {
 	pubACL := input.document[i].resource.aws_s3_bucket_public_access_block[name]
 	pubACL.ignore_public_acls == true
```
rogeriopeixotocx, line 7: I think this query should give results if the opposite happens.. that is ignore_public_acls == false

https://docs.aws.amazon.com/AmazonS3/latest/userguide/access-control-block-public-access.html

rogeriopeixotocx (changes requested):

assets/queries/terraform/aws/s3_bucket_with_ignore_public_acl/query.rego
```diff
@@ -1,5 +1,7 @@
 package Cx
```
rogeriopeixotocx, line 1: Query name should be s3 bucket without ignore public acl

rogeriopeixotocx (changes requested):

rogeriopeixotocx (approved):
LGTM

PR #4258, opened 2021-09-24
feat(Query): added "Pod or Container Without ResourceQuota" for Kubernetes

rafaela-soares:
**Proposed Changes**
- Added "Pod or Container Without ResourceQuota" for Kubernetes
- In draft until severity and category confirm


assets/queries/k8s/pod_or_container_without_resource_quota/query.rego
```diff
@@ -0,0 +1,61 @@
...
+
+is_default(document) {
+	document.metadata.namespace == "default"
+} else {
+	not common_lib.valid_key(document.metadata, "namespace")
+}
+
+pod_or_container(specInfo, document) {
+	specInfo.spec[types[x]]
+	document.kind != "Pod"
```
rogeriopeixotocx, line 58: I believe we need to check for templated types too (Deployments, DaemonSets, etc...)

rogeriopeixotocx (changes requested):

rogeriopeixotocx (approved):

Magier:
Hi, sorry for activating an already merged feature but I don't really understand the goal of this check.
ResourceQuotas (RQ) and LimitRanges (LR) are defined at namespace level. At Pod level you only define resource limits and requests.
So, the semantic of defining a RQ at a pod or container level - as the name of this query suggests - is wrong.

Enforcing the presence of RQ/LR would make only sense, if an entire namespace including it's workloads are configured in a single file. In the more common case where there is a manifest per workload (at least as judging by what I've seen so far) this leads to countless false-positives.

Maybe you could help me understand the reasoning for this check or is it actually an issue?

Thanks!

PR #4298, opened 2021-09-27
feat(query): added support to tf modules for ALB Not Dropping Invalid Headers

joaoReigota1:

**Proposed Changes**
- added support to tf modules for ALB Not Dropping Invalid Headers


felipe-avelar (changes requested):
I think we could add other negative sample, with `load_balancer_type` missing and `drop_invalid_header_fields = true` and other positive sample without both

felipe-avelar (approved):
LGTM

PR #4315, opened 2021-09-29
feat(query): corrected key actual value in "SQS With SSE Disabled" query

rafaela-soares:
**Proposed Changes**
- Corrected key actual value in "SQS With SSE Disabled" query


felipe-avelar (approved):
LGTM

PR #4318, opened 2021-09-29
feat(query): included check for cases when a statement is an object for "S3 Bucket Without SSL In Write Actions"

rafaela-soares:
**Proposed Changes**
- Included check for cases when a statement is an object for "S3 Bucket Without SSL In Write Actions"


rogeriopeixotocx (approved):
LGTM

PR #4362, opened 2021-10-11
feat(parser): added support to parse and scan terraform plans

cx-joao-reigota:

**Proposed Changes**
- added support to parse and scan terraform plans
- modified json parser so it can rebuild tfplan as kics document


NaorFirefly:
Hi, this is a superb feature, but you are parsing only the "planned_values" part, so you are missing the references that are critical for many of the KICS queries. The references are in the "root_module" object under "expressions" for each resource.

PR #4449, opened 2021-10-29
feat(query): added "Function App FTPS Enforce Disabled" for Terraform

rafaela-soares:
**Proposed Changes**
- Added "Function App FTPS Enforce Disabled" for Terraform

### DIFF
--- a/assets/queries/terraform/azure/function_app_ftps_enforce_disabled/metadata.json
+++ b/assets/queries/terraform/azure/function_app_ftps_enforce_disabled/metadata.json
@@ -3,7 +3,7 @@
   "queryName": "Function App FTPS Enforce Disabled",
   "severity": "HIGH",
   "category": "Insecure Configurations",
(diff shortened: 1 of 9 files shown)


JoaoDuarteGomes (commented):
Both category and severity are OK.

TBF, it is probably better to invert the query logic to flag only when this setting is "AllAllowed" or not set at all (since the default value is "AllAllowed"), both "FtpsOnly" and "Disabled" can be considered secure settings.

LMK what you think.

rafaela-soares:
> Both category and severity are OK.
> 
> TBF, it is probably better to invert the query logic to flag only when this setting is "AllAllowed" or not set at all (since the default value is "AllAllowed"), both "FtpsOnly" and "Disabled" can be considered secure settings.
> 
> LMK what you think.

Great observation! Thank you 🙂

JoaoDuarteGomes (approved):
LGTM.

nunoocx (approved):

PR #4451, opened 2021-10-29
feat(query): added "App Service FTPS Enforce Disabled" for Terraform

rafaela-soares:
**Proposed Changes**
- Added "App Service FTPS Enforce Disabled" for Terraform


JoaoDuarteGomes (commented):
I'm not sure I understand the difference to #4449.
It seems we are also flagging when the service is disabled which can be considered a secure configuration.

rafaela-soares:
> I'm not sure I understand the difference to #4449. It seems we are also flagging when the service is disabled which can be considered a secure configuration.

The main difference remains in the resource type. Additionally that, there is a difference in the default value of the 'ftps_state'.
I saw your suggestion in  #4449. Great observation, thank you 🙂

JoaoDuarteGomes (approved):
LGTM.

PR #4457, opened 2021-11-02
fix(cli): report-formats panics with uppercase input closes #4455

### LINKED ISSUE #4455: report-formats flag displays a panic message when receives uppercase input
### Expected Behavior

Some flags that requires an input value, such as `--report-formats` can receive values like: "json", "json,sarif", "html,sarif".
These flags are supposed to be case insensitive, such as `--log-level` that can receive, for example: `"TRACE"`, `"trace"` or even `"TrAcE"`.
If the flag isn't case insensitive, it can return an error or warning message, such as "invalid input", but it's not supposed to throw a panic message if an invalid case exists in the provided input.

### Actual Behavior

If the flag --report-formats receives an input with at least one uppercase character, kics will throw a panic message when before finishing the scan (and no report will be generated)

1.  Open terminal.
2.  Run any scan with a valid sample using the following commands:
- `kics scan -p samples -o output --report-formats ALL` (should display a panic message)
- `kics scan -p samples -o output --report-formats All` (should display a panic message)
- `kics scan -p samples -o output --report-formats all` (should complete the scan successfully)

image

### Specifications

-   Version: 1.4.6
-   Platform: N/A
-   Subsystem:  N/A

rogeriopeixotocx:
Closes #4455

### DIFF
--- a/internal/console/helpers/helpers.go
+++ b/internal/console/helpers/helpers.go
@@ -218,6 +218,7 @@ func GenerateReport(path, filename string, body interface{}, formats []string, p
 	defer progressBar.Close()
 
 	for _, format := range formats {
+		format = strings.ToLower(format)
 		if err = reportGenerators[format](path, filename, body); err != nil {
 			log.Error().Msgf("Failed to generate %s report", format)
 			break
--- a/internal/console/scan.go
+++ b/internal/console/scan.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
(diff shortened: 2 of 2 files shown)


cxlucas:
Hi, @rogeriopeixotocx does this implementation applies to all valid inputs?
Maybe I didn't add enough examples in my issue, but the panic occurs for any valid input with uppercase (such as Json, Sarif, HTML, All)

cxlucas (approved):
LGTM

PR #4478, opened 2021-11-04
feat(passwords and secrets): validation of query ids in custom secrets regexes

### LINKED ISSUE #4349: Additional validation of query-ids in custom secrets regexes file
When we want to use custom queries (like secrets regex), it's good to follow the patterns of kics, such as the query-id: a unique identifier related to a query. But currently, we can create and use custom secret regexes queries with any query-id value, with no need to follow the UUID pattern.

Even the Kics scans work well with a custom query id, the generated reports will have custom IDs associated with these queries, and it will "break" the pattern of query ids in the report.
So I propose an additional validation in the custom secrets regex file, provided by the user. This validation will ensure that the file has a valid query-id for all queries, following the UUID pattern. The user can also use "`kics generate-id`" to provide unique ids for their own queries.

N/A

**Additional context**
In addition, if the user provides invalid query ids (not UUID), he can not use "exclude-queries" flag properly, as this flag works with only valid UUIDs.

Example of result.json generated with a custom query-id

image

rafaela-soares:
Closes #4349

**Proposed Changes**
- Validation of query ids in custom secrets regexes

### DIFF
--- a/pkg/engine/secrets/inspector.go
+++ b/pkg/engine/secrets/inspector.go
@@ -553,7 +553,7 @@ func validateCustomSecretsQueriesID(allRegexQueries []RegexQuery) error {
 	for i := range allRegexQueries {
 		re := regexp.MustCompile(`^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$`)
 		if !(re.MatchString(allRegexQueries[i].ID)) {
-			return fmt.Errorf("the query %s defines a invalid query ID (%s)", allRegexQueries[i].Name, allRegexQueries[i].ID)
+			return fmt.Errorf("the query %s defines an invalid query ID (%s)", allRegexQueries[i].Name, allRegexQueries[i].ID)
 		}
 	}
 	return nil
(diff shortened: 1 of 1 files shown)


pkg/engine/secrets/inspector.go
```diff
@@ -541,3 +548,13 @@ func getPasswordsAndSecretsQueryID() (string, error) {
 	}
 	return metadata["id"], nil
 }
+
+func validateCustomSecretsQueriesID(allRegexQueries []RegexQuery) error {
+	for i := range allRegexQueries {
+		re := regexp.MustCompile(`^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$`)
+		if !(re.MatchString(allRegexQueries[i].ID)) {
+			return fmt.Errorf("the query %s defines a invalid query ID (%s)", allRegexQueries[i].Name, allRegexQueries[i].ID)
```
nunoocx, line 556: ...defines **an** invalid

nunoocx (changes requested):

felipe-avelar (approved):
LGTM

joaoReigota1 (approved):
LGTM

nunoocx (approved):
LGTM

PR #4510, opened 2021-11-14
feat(query): added "Google Compute Subnetwork with Private Google Access Disabled" for Ansible and Terraform

cosmicgirl97:
**Proposed Changes**
- Added "Google Compute Subnetwork with Private Google Access Disabled" for Ansible and Terraform

🚨 SEVERITY AND CATEGORY NEED REVIEW 🚨

### DIFF
--- a/assets/queries/ansible/gcp/google_compute_subnetwork_with_private_google_access_disabled/query.rego
+++ b/assets/queries/ansible/gcp/google_compute_subnetwork_with_private_google_access_disabled/query.rego
@@ -32,7 +32,7 @@ CxPolicy[result] {
 		"documentId": id,
 		"searchKey": sprintf("name={{%s}}.{{%s}}.private_ip_google_access", [task.name, modules[m]]),
 		"issueType": "IncorrectValue",
-		"keyExpectedValue": "%s.private_ip_google_access is set to true",
-		"keyActualValue": "%s.private_ip_google_access is set to false",
+		"keyExpectedValue":  sprintf("%s.private_ip_google_access is set to yes", [modules[m]]),
+		"keyActualValue": sprintf("%s.private_ip_google_access is set to no", [modules[m]]),
 	}
 }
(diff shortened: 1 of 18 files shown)


assets/queries/ansible/gcp/google_compute_subnetwork_with_private_google_access_disabled/query.rego
```diff
@@ -0,0 +1,40 @@
...
+	ans_lib.checkState(subnetwork)
+
+	subnetwork.private_ip_google_access != "yes"
+
+	result := {
+		"documentId": id,
+		"searchKey": sprintf("name={{%s}}.{{%s}}.private_ip_google_access", [task.name, modules[m]]),
+		"issueType": "IncorrectValue",
+		"keyExpectedValue": "%s.private_ip_google_access is set to true",
+		"keyActualValue": "%s.private_ip_google_access is set to false",
```
joaoReigota1, line 37: This needs to be an sprintf and pass modules[m]

joaoReigota1 (changes requested):

JoaoDuarteGomes (changes requested):
Category looks good.
Severity should be lowered to LOW.

joaoReigota1 (approved):
LGTM

JoaoDuarteGomes (approved):
LGTM.

puttehi:
Why is this considered a security issue? If it is not set/false ("positive result"), private instances cannot access Google services external IPs. If it is set/true ("negative result"), private instances can access Google services external IPs.

I don't think this has anything to do with security?

From [module documentation](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_subnetwork#private_ip_google_access):
> When enabled, VMs in this subnetwork without external IP addresses can access Google APIs and services by using Private Google Access.

From [Google documentation](https://cloud.google.com/vpc/docs/private-google-access):
> Private Google Access
> 
> VM instances that only have internal IP addresses (no external IP addresses) can use Private Google Access. They can reach the external IP addresses of Google APIs and services. The source IP address of the packet can be the primary internal IP address of the network interface or an address in an alias IP range that is assigned to the interface. If you disable Private Google Access, the VM instances can no longer reach Google APIs and services; they can only send traffic within the VPC network.
> 
> Private Google Access has no effect on instances that have external IP addresses. Instances with external IP addresses can access the internet, according to the [internet access requirements](https://cloud.google.com/vpc/docs/vpc#internet_access_reqs). They don't need any special configuration to send requests to the external IP addresses of Google APIs and services.
> 
> You enable Private Google Access on a subnet by subnet basis; it's a setting for subnets in a VPC network. To enable a subnet for Private Google Access and to view the requirements, see [Configure Private Google Access](https://cloud.google.com/vpc/docs/configure-private-google-access).

PR #4535, opened 2021-11-18
refactor(query): policies for Terraform

rafaela-soares:
**Proposed Changes**
- Refactored queries that do not consider correctly Statement as an array of objects or object or Effect set to allow


felipe-avelar (dismissed):
LGTM

felipe-avelar (approved):
LGTM

PR #4568, opened 2021-11-26
feat(report): Added JUnit report

### LINKED ISSUE #3902: Support JUnit report format
Most CI systems support JUnit format

We would like to have the report format as JUnit (XML) format

Another common report format (glsast isn't common as JUnit)

felipe-avelar:

Closes #3902 

**Proposed Changes**
- Now KICS can exports results in JUnit XML format
- The makefile was changed to increase test timeout, since e2e tests was panicking with 10 minutes timeout


docs/integrations_jenkins.md
```diff
@@ -128,3 +128,47 @@ Jenkins will exhibit the following warning:
...
+If it is not installed, you can install it by clicking on **Plugins** on the left menu bar, then click on **Install Plugin** and select **JUnit**.
+
+After that, you can add the following line to your pipeline script:
+```groovy
+    junit testResults: 'results/junit-results.xml', skipPublishingChecks: true
+```
+
+The skipPublishingChecks option is important, otherwise the JUnit plugin will try to publish the results to GitHub in Jenkins. If you want to do this, you should check the [JUnit plugin documentation](https://plugins.jenkins.io/junit/) for more information.
+
+If your using our example script, it will looks like the following:
```
rafaela-soares, line 145: If you are using our example script, it will look like the following:

docs/results.md
```diff
@@ -432,6 +434,51 @@ Gitlab SAST reports are sorted by severity (from high to info), following [Gitla
 
 ```
 
+## JUnit
+You can export JUnit report by using `--report-formats "junit"`.
+JUnit reports follows [JUnit XML specification by junit-team](https://github.com/junit-team/junit5/blob/main/platform-tests/src/test/resources/jenkins-junit.xsd), also, the generated file will have a prefix `junit-` and looks like:
```
rafaela-soares, line 439: JUnit reports follow ...

pkg/report/model/junit.go
```diff
@@ -0,0 +1,107 @@
...
+	failedTestCase := junitTestCase{
+		Name:     fmt.Sprintf("[%s]: %s", query.QueryName, queryDescription),
+		Failures: make([]junitFailure, len(query.Files)),
+	}
+
+	for idx := range query.Files {
+		failedTest := junitFailure{
+			Type: query.QueryName,
+			Message: fmt.Sprintf(
+				"A problem was found on '%s' file in line %d, %s, but %s.",
```
rafaela-soares, line 71: I suggest something like: `"A problem was found on '%s' file, in line %d. The actual value is %s. However, the expected value is %s.,` or `[file: %s, line: %d] The actual value is %s. However, the expected value is %s`.

docs/dockerhub.md
```diff
@@ -96,7 +96,7 @@ Flags:
   -d, --payload-path string           path to store internal representation JSON file
       --preview-lines int             number of lines to be display in CLI results (min: 1, max: 30) (default 3)
   -q, --queries-path string           path to directory with queries (default "./assets/queries")
-      --report-formats strings        formats in which the results will be exported (all, glsast, html, json, pdf, sarif) (default [json])
+      --report-formats strings        formats in which the results will be exported (all, glsast, html, json, junit, pdf, sarif, sonarqube) (default [json])
```
rafaela-soares, line 99: I think the same update can be done in `commands.md`

rafaela-soares (changes requested):
LGTM 🚀 
I have only minor suggestions:

pkg/report/model/junit.go
felipe-avelar: KICS already gives a descriptive message of what is  expected and what is the actual value, so with some "prefix" on message can make it strange, since it will look something like: "The actual value is 'The attribute <some attribute here> is <wrong value>'. However, the expected value is 'The attribute <some attribute here> is expected to be <correct value>'...
This is why I think the pattern `<expected>, but <actual>` works fine here.

rafaela-soares (approved):
LGTM

PR #4589, opened 2021-12-10
fix(ci): Fixed docs version regex of prepare-release workflow

cxlucas:
**Proposed Changes**
- Fix the regex to identify any valid version in kics docs (the current regex didn't identify version 1.4.6.1).


joaoReigota1 (approved):
LGTM

PR #4609, opened 2021-12-20
fix(query): refactored query Resource Not Using Tags to lower execution time

joaoReigota1:

**Proposed Changes**
-  refactored query Resource Not Using Tags to lower execution time


rafaela-soares (approved):
LGTM

PR #4618, opened 2021-12-20
fix(query): deleting searchLine in "Resource Not Using Tags" for Terraform

rafaela-soares:
**Proposed Changes**
- Deleting searchLine in "Resource Not Using Tags" for Terraform


joaoReigota1 (approved):
LGTM

PR #4627, opened 2021-12-21
add allow rule for mysql password hashes

rndmh3ro:
**Proposed Changes**
MySQL-password are twiced hashed.
They are always 41 bytes long with the fist byte being an asterisk.


rndmh3ro:
Any opinions about this one? I'd be fine if it doesn't get merged as I have excluded the passwords with `# kics-scan ignore-block`

joaoReigota1 (approved):
LGTM

joaoReigota1:
Hi @rndmh3ro, sorry for the late response the Pull request was accepted and should be present in the next release.
Thank you for using KICS!

PR #4650, opened 2021-12-27
feat(query): added "Serverless Function Without Unique IAM Role" for CloudFormation

rafaela-soares:
**Proposed Changes**
- Added "Serverless Function Without Unique IAM Role" for CloudFormation


joaoReigota1 (changes requested):
The Role can also be passed as a Reference using !Ref, I think we should add this case as well

rafaela-soares:
> The Role can also be passed as a Reference using !Ref, I think we should add this case as well

The YAML parser interprets the !Ref. I added samples for those cases.

joaoReigota1 (approved):
LGTM

PR #4640, opened 2021-12-27
feat(query): support AWS SAM - split CloudFormation into aws and aws_sam folders

rafaela-soares:
**Proposed Changes**
- Support AWS SAM - split CloudFormation into aws and aws_sam folders

I submit this contribution under the Apache-2.0 license.

PR #4652, opened 2021-12-28
Fix queries counter after aws / aws sam split

kaplanlior:
Follow up to (#4640)

-


joaoReigota1 (approved):
LGTM

PR #4658, opened 2022-01-03
fix(parser): fixed KICS panic in empty fifo value access

rafaela-soares:
**Proposed Changes**
- Fixed KICS panic in empty fifo value access: added length check


joaoReigota1 (approved):
LGTM

PR #4669, opened 2022-01-06
feat(cli): added contribution appeal when the user includes external queries

### LINKED ISSUE #2224: Include external queries option
KICS comes with a wide set of queries, which is great.
I want to request a feature that would ease people to create more queries.

ATM, for someone to add query, files must be added to KICS directory hierarchy.
If the user want to upgrade KICS he needs to remember to copy/move all the files
he'd added to the new copy of KICS.

If we have an option for KICS to include external queries it would simply the process.
Allowing a person or an organization to keep their customization between upgrades/changes, or to manage them on their GIT.

Please notice that in case the user uses this option, we also want to suggest he'd contribute those queries to KICS, so everyone can enjoy them. Linking to the contribution page.

rafaela-soares:
Closes #2224

**Proposed Changes**
- Added contribution appeal when the user includes external queries (the queries path do not follow assets/queries structure)
- Changed `PrintVersionCheck` location and colour

### DIFF
--- a/internal/console/helpers/helpers.go
+++ b/internal/console/helpers/helpers.go
@@ -46,13 +46,15 @@ var reportGenerators = map[string]func(path, filename string, body interface{}) 
 // Line is the color to print the line with the vulnerability
 // minVersion is a bool that if true will print the results output in a minimum version
 type Printer struct {
-	Medium  color.RGBColor
-	High    color.RGBColor
-	Low     color.RGBColor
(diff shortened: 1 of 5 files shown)


joaoReigota1 (changes requested):
This should be done using the Printer package KICS currently has

joaoReigota1 (approved):
LGTM

PR #4727, opened 2022-01-20
feat(query): added "API Gateway Without Configured Authorizer" for CloudFormation and Terraform

cosmicgirl97:
**Proposed Changes**
- Added "API Gateway Without Configured Authorizer" for CloudFormation and Terraform


joaoReigota1 (approved):
LGTM

PR #4740, opened 2022-01-24
feat(buildah): added support to Buildah

### LINKED ISSUE #3135: Add Support to Buildah
https://buildah.io/
> A tool that facilitates building OCI container images.

can be used via script to build container by separate steps not through Containerfile/Dockerfile (when `buildah bud` not used), for example:
https://www.redhat.com/sysadmin/building-buildah#creating-and-running-the-bash-script

There lots of results on github for searching `buildah run`, the issue about adding support for such scripts (or invocation of `buildah` in other ways):
https://github.com/chapeaux/cncpx/blob/f7311f81eafedb5a5780518e667e7c566ce1021f/containers/images/godot_ubi8.sh

rafaela-soares:
Closes #3135

**Proposed Changes**
- Added support to Buildah
- Added buildah parser
- Added 2 Buildah queries: "Run Using apt" and "Run Using Dnf Update"

P.S. For now, there is no file test for `comments.go` because package `mvdan.cc/sh/v3` does not provide the creation of struct `Pos` in the current version but will provide it in the future


docs/platforms.md
```diff
@@ -10,6 +10,10 @@ KICS supports scanning Ansible files with `.yaml` extension.
 
 KICS supports scanning Azure Resource Manager (ARM) templates with `.json` extension. To build ARM JSON templates from Bicep code check the [official ARM documentation](https://docs.microsoft.com/en-us/azure/azure-resource-manager/bicep/bicep-cli#build) and [here](https://docs.microsoft.com/en-us/azure/azure-resource-manager/bicep/compare-template-syntax) to understand the differences between ARM JSON templates and Bicep
 
+## Buidah
+KICS supports scanning Buidah files with `.sh` extension.
```
joaoReigota1, line 14:
```suggestion
## Buildah
KICS supports scanning Buildah files with `.sh` extension.
```

go.mod
```diff
@@ -38,7 +38,8 @@ require (
 	github.com/zclconf/go-cty v1.10.0
 	golang.org/x/net v0.0.0-20211008194852-3b03d305991f
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
-	helm.sh/helm/v3 v3.7.2
+	helm.sh/helm/v3 v3.7.1
```
joaoReigota1, line 41: Helm package version shouldn't be rolled back

pkg/analyzer/analyzer.go
```diff
@@ -42,6 +42,7 @@ var (
 	blueprintArtifactsRegexProperties = regexp.MustCompile("(\\s*\"properties\":)|(\\s*properties:)")
 	blueprintRegexTargetScope         = regexp.MustCompile("(\\s*\"targetScope\":)|(\\s*targetScope:)")
 	blueprintRegexProperties          = regexp.MustCompile("(\\s*\"properties\":)|(\\s*properties:)")
+	buildahRegex                      = regexp.MustCompile("buildah")
```
joaoReigota1, line 45: Only `buildah` is too generic and may conflict with other files, I think we should use a different approach like:
```
 buildahRegex = regexp.MustCompile("\$\(\s*buildah\s+from\s+\w+\)")
```
that looks for the from command for buildah which every file is required to have

pkg/parser/buildah/parser.go
```diff
@@ -0,0 +1,206 @@
...
+		}
+	}
+	return command
+}
+
+func getWordValue(wd *syntax.Word) string {
+	printer := syntax.NewPrinter()
+	var buf bytes.Buffer
+
+	printer.Print(&buf, wd)
```
joaoReigota1, line 159: We should handle this error also, if we are only using the printer to place the wd inside a buffer can't we use the `fmt` package with the `Frintf()` method?

Same for line 171

joaoReigota1 (changes requested):

joaoReigota1 (approved):
LGTM

PR #4762, opened 2022-01-27
fix(query): service should match containerPort using targetPort

Churro:
**Problem**

The Kubernetes rule "_Service Does Not Target Pod_" matches a Service with `containerPort` based on the `port` rather than the `targetPort` attribute. This leads to false positive findings in situations where `port` and `targetPort` have different values.

This seems to be a regression introduced with commit 3eb9ff4a615b274bc52d16e01c1b248b3464c27c. It was not detected by existing tests, as `port` and `targetPort` always correspond to the same value.

**Proposed Changes**
- Change `port` to `targetPort` to reflect the fact that a Service can map any incoming port to a target container port
- Add another true negative testcase to cover the scenario of different ports in a Service
- Rename `negative.yaml` to `negative1.yaml`


joaoReigota1 (approved):
LGTM

PR #4788, opened 2022-02-02
fix(query): containers-run-with-low-uid k8s query should consider statement precedence

Churro:
**Problem**

The K8s rule _containers-run-with-low-uid_ traverses documents recursively, searching for `runAsUser` defintions. By doing so, the implementation disregards hierarchical inheritance, in which a `runAsUser` statement can occur. If defined in the SecurityContext of a pod, it is inherited by all containers, unless redefined in the SecurityContext of a container. More information about precedence is provided [here](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container).

The current rule shows false positive alerts for each SecurityContext definition, which does not include `runAsUser`.

**Proposed Changes**

- Change the current implementation to check for possible definitions at multiple locations:
  - Does a container explicitly define `runAsUser`?
  - Does a container inherit `runAsUser` from the pod SecurityContext?
  - Do neither pod nor container define `runAsUser`?

- Additional tests
  * _negative2.yaml_: `runAsUser` defined in `spec.template.spec.securityContext`
    *  Existing rule: 1 false positive match
  * _negative3.yaml_: `runAsUser` defined in `spec.template.spec.securityContext` and SecurityContext of one container
    *  Existing rule: 1 false positive match
  * _positive4.yaml_: IncorrectValue in `spec.template.spec.securityContext.runAsUser`
    * Existing rule: 3 false positive matches
  * _positive5.yaml_: MissingAttribute in SecurityContext of containers
  * _positive6.yaml_: IncorrectValue in SecurityContext of containers
  * _positive7.yaml_: IncorrectValue and MissingAttribute in SecurityContext of containers
    * Existing rule:
      * Does not recognize missing `runAsUser` statement in second container
      * Missing `runAsUser` attribute for second container is not recognized

### DIFF
--- a/assets/queries/k8s/containers_run_with_low_uid/query.rego
+++ b/assets/queries/k8s/containers_run_with_low_uid/query.rego
@@ -1,39 +1,63 @@
 package Cx
 
 import data.generic.common as common_lib
+import data.generic.k8s as k8sLib
 
-CxPolicy[result] {
-	doc := input.document[i]
+types := {"initContainers", "containers"}
 
-	[path, value] = walk(doc)
-	securityContext := value.securityContext
-
-	to_number(securityContext.runAsUser) < 10000
+# container defines runAsUser
+checkUser(specInfo, container, containerType, document, metadata) = result {
+	uid := container.securityContext.runAsUser
+	to_number(uid) < 10000
 
 	result := {
-		"documentId": doc.id,
-		"searchKey": sprintf("%s.securityContext.runAsUser=%d", [common_lib.concat_path(path), securityContext.runAsUser]),
+		"documentId": document.id,
+		"searchKey": sprintf("metadata.name={{%s}}.%s.%s.name={{%s}}.securityContext.runAsUser=%d", [metadata.name, specInfo.path, containerType, container.name, uid]),
 		"issueType": "IncorrectValue",
-		"keyExpectedValue": sprintf("%s.securityContext.runAsUser should not be a low UID", [common_lib.concat_path(path)]),
-		"keyActualValue": sprintf("%s.securityContext.runAsUser is a low UID", [common_lib.concat_path(path)]),
-		"searchLine": common_lib.build_search_line(path, ["securityContext", "runAsUser"]),
+		"keyExpectedValue": sprintf("metadata.name={{%s}}.%s.%s.name={{%s}}.securityContext.runAsUser should be set to a UID >= 10000", [metadata.name, specInfo.path, containerType, container.name]),
+		"keyActualValue": sprintf("metadata.name={{%s}}.%s.%s.name={{%s}}.securityContext.runAsUser is set to a low UID", [metadata.name, specInfo.path, containerType, container.name]),
+	}
+}
+
+# pod defines runAsUser and container inherits this setting
+checkUser(specInfo, container, containerType, document, metadata) = result {
+	containerCtx := object.get(container, "securityContext", {})
+	not common_lib.valid_key(containerCtx, "runAsUser")
+
+	uid := specInfo.spec.securityContext.runAsUser
+	to_number(uid) < 10000
+
+	result := {
+		"documentId": document.id,
+		"searchKey": sprintf("metadata.name={{%s}}.%s.securityContext.runAsUser=%d", [metadata.name, specInfo.path, uid]),
+		"issueType": "IncorrectValue",
+		"keyExpectedValue": sprintf("metadata.name={{%s}}.%s.securityContext.runAsUser should be set to a UID >= 10000", [metadata.name, specInfo.path]),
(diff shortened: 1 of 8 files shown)


assets/queries/k8s/containers_run_with_low_uid/query.rego
```diff
@@ -1,39 +1,63 @@
...
+		"keyActualValue": sprintf("metadata.name={{%s}}.%s.securityContext.runAsUser is set to a low UID", [metadata.name, specInfo.path]),
+	}
+}
 
-	[path, value] = walk(doc)
-	securityContext := value.securityContext
+# neither pod nor container define runAsUser
+checkUser(specInfo, container, containerType, document, metadata) = result {
+	specCtx := object.get(specInfo.spec, "securityContext", {})
+	not common_lib.valid_key(specCtx, "runAsUser")
```
rafaela-soares, line 42:
```suggestion
	not common_lib.valid_key(specInfo.spec.securityContext, "runAsUser")
```

assets/queries/k8s/containers_run_with_low_uid/query.rego
```diff
@@ -1,39 +1,63 @@
...
+		"keyActualValue": sprintf("metadata.name={{%s}}.%s.%s.name={{%s}}.securityContext.runAsUser is set to a low UID", [metadata.name, specInfo.path, containerType, container.name]),
 	}
 }
 
-CxPolicy[result] {
-	doc := input.document[i]
+# pod defines runAsUser and container inherits this setting
+checkUser(specInfo, container, containerType, document, metadata) = result {
+	containerCtx := object.get(container, "securityContext", {})
+	not common_lib.valid_key(containerCtx, "runAsUser")
```
rafaela-soares, line 25:
```suggestion
	not common_lib.valid_key(container.securityContext, "runAsUser")
```

assets/queries/k8s/containers_run_with_low_uid/query.rego
```diff
@@ -1,39 +1,63 @@
...
-	[path, value] = walk(doc)
-	securityContext := value.securityContext
+# neither pod nor container define runAsUser
+checkUser(specInfo, container, containerType, document, metadata) = result {
+	specCtx := object.get(specInfo.spec, "securityContext", {})
+	not common_lib.valid_key(specCtx, "runAsUser")
 
-	not common_lib.valid_key(securityContext, "runAsUser")
+	containerCtx := object.get(container, "securityContext", {})
+	not common_lib.valid_key(containerCtx, "runAsUser")
```
rafaela-soares, line 45:
```suggestion
	not common_lib.valid_key(container.securityContext, "runAsUser")
```

rafaela-soares (changes requested):
Another great contribution! 🚀 Thank you, @Churro!

Let me know what you think about the following suggestions, please.

assets/queries/k8s/containers_run_with_low_uid/query.rego
Churro: The desired condition here is: `runAsUser` is not defined in a _potentially_ existing `specInfo.spec.securityContext` object. 

The `object.get(.., {})` check ensures that the securityContext can exist but doesn't necessarily have to. With the check you suggested the rule would evaluate to _false_, in case no securityContext is defined on pod level. However, this can be perfectly legit, since `runAsUser` may instead be defined on container level.

assets/queries/k8s/containers_run_with_low_uid/query.rego
Churro: The desired condition here is: `runAsUser` is not defined in a _potentially_ existing `container.securityContext` object. 

The `object.get(.., {})` check ensures that the securityContext can exist but doesn't necessarily have to. With the check you suggested the rule would evaluate to _false_, in case no securityContext is defined on container level. However, this can be perfectly legit, since `runAsUser` may instead be defined on pod level.

assets/queries/k8s/containers_run_with_low_uid/query.rego
Churro: The `object.get(.., {})` check ensures that the securityContext can exist but doesn't necessarily have to. This check is crucial because otherwise, the rule would evaluate to _false_, in case no securityContext is yet defined on container level.

Churro:
Thank you for the review @rafaela-soares! :smiley:

I've tried to address your concerns inline with your suggestions. Please let me know, in case there are further issues you would like to see addressed.

rafaela-soares:
> Thank you for the review @rafaela-soares! 😃
> 
> I've tried to address your concerns inline with your suggestions. Please let me know, in case there are further issues you would like to see addressed.

I understood your approach. Very impressive! Thank you so much for your time explaining it 🙂 
I have no further issues. Thank you so much! 🚀

rafaela-soares (approved):
LGTM

PR #4817, opened 2022-02-09
fix(query): update service_does_not_target_pod k8s rule with new logic to match labels and ports

Churro:
**Problem**

PR #4762 corrected port matching to targetPort. However, in the current rule more issues deserve to be addressed:

* When matching service selectors with pod labels:
  * `matchLabel()` only compares only label values, not their keys
  * `count(ret) != 0` does not ensure all service selector labels are also defined in pod labels. 
  One label is enough to cause a potential false positive alert.
  * With some kinds the labels to match must be specified in `spec.selector.matchLabels` instead of `metadata.labels`
    * E.g., see structs with `Selector *metav1.LabelSelector` [here](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/apps/v1/types.go)
* When matching ports:
  * `targetPort` could also be a string. See [here](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#serviceport-v1-core):
    > If this is a string, it will be looked up as a named port in the target Pod's container ports. If this is not specified, the value of the 'port' field is used (an identity map)
  * `targetPort` could be omitted overall. See [here](https://kubernetes.io/docs/concepts/services-networking/service/#defining-a-service):
    > Note: A Service can map any incoming port to a targetPort. By default and for convenience, the targetPort is set to the same value as the port field. 
* `spec.ports.name` (in case of kind: _Service_) may be omitted overall. In that case, CxPolicy evaluates to false in result lines `keyExpectedValue` and `keyActualValue`, effectively preventing any result from being returned

**Proposed Changes**

* When matching service selectors with pod labels, remediate the problems described above
* When matching ports: cover all possibilities, following the same logic that is also implemented in K8s (see [here](https://github.com/kubernetes/kubectl/blob/master/pkg/util/service_port.go))
* Include port instead of (potentially missing) name in result
* Additional test cases
  * _negative3.yaml_: targetPort is string, containerPort with name
  * _negative4.yaml_:
    * Contains labels in `selector.matchLabels` instead of `metadata.labels`
    * service port without name attribute
  * _negative5.yaml_: targetPort omitted


rafaela-soares (dismissed):
You rock, @Churro 🚀 Thank you so much for another great contribution!

LGTM!

Churro:
Thank you for the feedback @rafaela-soares!

And apologies for a small mistake that I just corrected in a follow-up commit. The test files were not added to the correct folder. I hope that now everything is fine 🙏

Churro:
Thank you for the feedback @rafaela-soares!

And apologies for a small mistake that I just corrected in a follow-up commit. The test files were not added to the correct folder. I hope that now everything is fine 🙏

Churro:
The most recent commits now also address the previously failing CI tests and add an additional safeguard to prevent potential false positives with services that specify no selectors, e.g., with `type: ExternalName`.

Due to the fact that KICS integration tests combine YAML files from all queries, there were also "duplicates" that led to unexpected multiple results when only one was expected. What the commit changed is the way other resources are searched, i.e., more resiliency by using a comprehension instead of expecting only one result a priori.

I have now even more carefully reviewed and tested everything again and see no further issues. Would be cool to see this approved once again and get it merged 😊

assets/queries/k8s/service_does_not_target_pod/query.rego
```diff
@@ -1,56 +1,84 @@
...
-	label := service.spec.selector[_]
-	ret := match_label(service.spec.selector)
-	count(ret) != 0
-	not confirmPorts(ret, servicePorts)
+
+	resources := [x | x := input.document[_]; matchResource(x, service.spec.selector)]
+	count(resources) > 0
+
+	servicePorts := service.spec.ports[_]
+	not confirmPorts(resources[0], servicePorts)
```
rafaela-soares, line 15: Hello again, @Churro! 🙂 Thank you so much for your changes and great work improving this query!


I reviewed your last changes and I think `confirmPorts` should look in all resources, instead of the first one. Let me know what you think about it, please.

The following scenario led me to think about what I suggest:
- file1.yaml
```
apiVersion: v1
kind: Service
metadata:
  name: helloworld
spec:
  type: NodePort
  selector:
    app: helloworld
  ports:
    - name: http
      nodePort: 30475
      port: 8089
      protocol: TCP
      targetPort: 8089
```
- file2.yaml
```
apiVersion: v1
kind: Pod
metadata:
  name: nginx3
  labels:
    app: helloworld
spec:
  containers:
  - name: nginx
    image: nginx
    ports:
      - containerPort: 808
```
- file3.yaml
```
apiVersion: v1
kind: Pod
metadata:
  name: nginx
  labels:
    app: helloworld
spec:
  containers:
  - name: nginx
    image: nginx
    ports:
      - containerPort: 8089

```

`resources[0]` will possibly contain the pod defined in `file2.yaml` and the query will return a result because `confirmPorts` will fail. However, there is a pod that is targeted by the service in `file3.yaml`.

rafaela-soares (changes requested):

assets/queries/k8s/service_does_not_target_pod/query.rego
Churro: Thank you for the comprehensive review, @rafaela-soares! 😊 

The use-case you describe is perfectly possible. I have now updated the rule to check all resources and hopefully you don't mind that I just took your example and added it as another negative testcase.

rafaela-soares: Of course not 😊 Thank you so much!

PR #4829, opened 2022-02-13
fix(query): extended readinessProbe k8s rule to cover additional resources

Churro:
**Problem**

- `document.spec[types[x]][j]` only applies to Pods but not Deployment, DaemonSet, etc. 
- Related to that, `searchKey` points to the wrong location with resources other than Pods
- `readinessProbe`, by design, is not working with `initContainers` (see [here](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/)):
  > Also, init containers do not support lifecycle, livenessProbe, readinessProbe, or startupProbe because they must run to completion before the Pod can be ready.

**Proposed Changes**
- A rule implementation analogous to the livenessProbe rule which remediates the described issues

### DIFF
--- a/assets/queries/k8s/readiness_probe_is_not_configured/query.rego
+++ b/assets/queries/k8s/readiness_probe_is_not_configured/query.rego
@@ -1,28 +1,24 @@
 package Cx
 
-import data.generic.k8s as k8s
+import data.generic.k8s as k8sLib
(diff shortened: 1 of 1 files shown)


rafaela-soares (approved):
Another great observation 🚀 Thank you!!

LGTM!

PR #4834, opened 2022-02-14
fix(query): changed severity, description text and URL in "Liveness Probe Is Not Defined" query

### LINKED ISSUE #4755: Liveness probe rule for k8s
I'm getting a medium vulnerability because I don't have a default liveness probe defined in my helm chart:

```json
    {
      "id": "127b677571fe7efaabd12b4ad27505051cb321a7d459dc12d07417421999e621",
      "category": "sast",
      "message": "Liveness Probe must be defined.",
      "description": "metadata.name={{kics-helm}}.spec.containers.name={{kics-helm}}.livenessProbe is undefined",
      "cve": "",
      "severity": "Medium",
      "scanner": {
        "id": "kics",
        "name": "kics"
      },
      "location": {
        "file": "pipeline/templates/deployment.yaml",
        "start_line": 50
      },
      "identifiers": [
        {
          "type": "kics_id",
          "name": "Liveness Probe Is Not Defined",
          "value": "ade74944-a674-4e00-859e-c6eab5bde441",
          "url": "https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-tcp-liveness-probe"
        }
      ]
    },
```

Why is this an issue? K8s' own documentation doesn't recommend having a liveness probe out of the box:

   > If the process in your container is able to crash on its own whenever it encounters an issue or becomes unhealthy, you do not necessarily need a liveness probe; the kubelet will automatically perform the correct action in accordance with the Pod's restartPolicy.

   (https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#when-should-you-use-a-liveness-probe)

I'm also unsure why it would be "medium." I'd put it at "info", maybe "low" at the highest.

I would have expected a warning about a missing readiness probe, however; at least in my experience, a missing readiness probe can result in helm deploying a failing deployment which would have otherwise been successfully detected as a failing deployment and rolled back with a readiness probe, probably due to this:

> The default state of readiness before the initial delay is `Failure`. If a container does not provide a readiness probe, the default state is `Success`.

(https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#types-of-probe)

I'm not a k8s expert so maybe I have something to learn here! If so, at the very least it would be nice if the link was updated. The default link for this test (https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-tcp-liveness-probe) is quite unhelpful.

Thanks!

rafaela-soares:
Closes #4755

**Proposed Changes**
- Changed severity, description text and URL in "Liveness Probe Is Not Defined" query

### DIFF
--- a/assets/queries/k8s/liveness_probe_is_not_defined/metadata.json
+++ b/assets/queries/k8s/liveness_probe_is_not_defined/metadata.json
@@ -1,10 +1,10 @@
 {
   "id": "ade74944-a674-4e00-859e-c6eab5bde441",
   "queryName": "Liveness Probe Is Not Defined",
-  "severity": "MEDIUM",
+  "severity": "INFO",
   "category": "Availability",
-  "descriptionText": "Liveness Probe must be defined.",
-  "descriptionUrl": "https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-tcp-liveness-probe",
+  "descriptionText": "In case of an unresponsive container, a Liveness Probe can help your application become more available since it restarts the container. However, it can lead to cascading failures. Define one if you really need it",
+  "descriptionUrl": "https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#when-should-you-use-a-liveness-probe",
   "platform": "Kubernetes",
   "descriptionID": "f724fa60"
 }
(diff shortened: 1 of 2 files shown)


JoaoDuarteGomes (approved):
Severity LGTM.

joaoReigota1 (approved):

PR #4869, opened 2022-02-23
feat(test): Added more tests

joaorufi:
**Proposed Changes**
- Added more tests


pkg/report/asff_test.go
```diff
@@ -0,0 +1,39 @@
...
+		t.Run(test.name, func(t *testing.T) {
+			if err := os.MkdirAll(test.caseTest.path, os.ModePerm); err != nil {
+				t.Fatal(err)
+			}
+
+			err := PrintASFFReport(test.caseTest.path, test.caseTest.filename, test.caseTest.summary)
+			require.NoError(t, err)
+
+			require.FileExists(t, filepath.Join(test.caseTest.path, "asff-"+test.caseTest.filename+".json"))
+		})
```
rafaela-soares, line 37: Maybe we should remove the file

```suggestion
			require.FileExists(t, filepath.Join(test.caseTest.path, "asff-"+test.caseTest.filename+".json"))
			os.RemoveAll(test.caseTest.path)
		})
```

rafaela-soares (changes requested):

pkg/report/asff_test.go
joaorufi: I'm using `/tmp` dir to save files, so I think there is no need to delete them, since they are deleted automatically

rafaela-soares: Are you sure? At least, in Windows, my files in `/tmp` seem to not be deleted automatically

joaoReigota1 (approved):

rafaela-soares (approved):
LGTM

PR #4882, opened 2022-02-27
fix(query): Extend container_is_privileged k8s rule to cover additional resource kinds

Churro:
**Proposed Changes**

- New positive test case that covers `privileged: true` in a Deployment
- More comprehensive description about security implications


rafaela-soares (approved):
LGTM!

Such a great observation and refactor, @Churro 🚀 Thank you so much!

PR #4884, opened 2022-02-27
fix(query): Extend net_raw_capabilities_not_being_dropped k8s rule to cover further resource kinds

Churro:
**Proposed Changes**

- Fold the check whether `securityContext.capabilities.drop` is defined into one rule
  - Even if `securityContext` or `securityContext.capabilities` are undefined, it is now immediately highlighted that `securityContext.capabilities.drop` is missing


rafaela-soares (changes requested):
Hi, @Churro!

First of all, sorry for the delay in reviewing some of your PRs. And thank you so much for your contributions to KICS!

Your approach (check whether securityContext.capabilities.drop is defined into one rule) is very impressive! 🚀 

However, due to how the KICS `searchKey` is defined, the `searchKey` points wrongly with your approach. For example, take a look at the `positive1.yaml`.

image

In the master, there is a result in line 13. In this PR, the result points to line 18. That happens because KICS interprets "." in the [searchKey](https://github.com/Checkmarx/kics/pull/4884/files#diff-c30628de2e3da519fe65c8b0a7761282e0222bbf680355e0b82ec7c91be04773R40) as a "breaking line" and it will search for the key(s) in the file. 

In this case, although there is no `securityContext` defined in the container named as payment2 (line 13) the `searchKey` is pointing to line 18 because the searchKey tells KICS to search for `securityContext.capabilities`.

As a solution for that, we built the function `get_nested_values_info` that receives two inputs: object and array_vals. This function verifies if the array_vals are nested in the object. It will return an object with two fields: **valid** (returns if the array_vals are nested in the object (array_vals should be sorted) and **searchKey** returns the searchKey possible.

For example:

```
object := {"elem1": {"elem2": "elem3"}}
array_vals := ["elem2", "elem3", "elem4"]
return_value := {"valid": false, "searchKey": "elem2.elem3"}
```

Can you apply the function `get_nested_values_info`, please? 😊 You can find more examples of the application [here](https://github.com/Checkmarx/kics/blob/master/assets/queries/k8s/containers_run_with_low_uid/query.rego#L42-L50).  (Please merge the branch with the master)

P.S. The other PRs with the same approach should also apply the function `get_nested_values_info`

Churro:
Hi @rafaela-soares, I have now applied your suggestions and merged with master.

From an engineering perspective, however, I don't appreciate the introduction of the `get_nested_values_info` function and would strongly encourage you to reconsider this approach. It basically adds significant complexity to queries that is unrelated to the purpose of the rules themselves and only needed for kics output. 

My suggestion would be to refine the `searchKey` implementation to deliver the expectable output, rather than trying to workaround a deficiency in the matching strategy with adaptations to countless rules. Probably, it helps if illustrate what I mean with two concrete examples:

1. **searchKey matching**
For the sample snippet you posted, the `searchKey` output already included the container name, i.e.:
```
"searchKey": "metadata.name={{example}}.spec.containers.name={{payment2}}.securityContext.capabilities"
...
"searchKey": "metadata.name={{example}}.spec.containers.name={{payment3}}.securityContext.capabilities
...
"searchKey": "metadata.name={{example}}.spec.containers.name={{payment4}}.securityContext.capabilities"
...
"searchKey": "metadata.name={{example}}.spec.containers.name={{payment}}.securityContext.capabilities"
```
I would expect the `searchKey` implementation to understand that a search key which includes `containers.name={{payment2}}` cannot be located in container `payment4`. Instead, I would assume that `searchKey` always points to that line in the YAML with the greatest common denominator, i.e., if there is a container named `payment2` but no `securityContext.capabilities`, it should still point to the line with `name: payment2`.

2. **Runtime complexity**
By replacing `object.get(...)` with `get_nested_values_info` you switch from a runtime complexity of O(1) to O(n). This is because `walk()` traverses all branches of the document and doesn't stop when it finds `array_vals`.

The effect becomes dramatically obvious if we take the sample snippet of our screenshot and profile it with my original query (eval time: 0,371931ms) and the one you suggested with `get_nested_values_info` (eval time: 1,191899ms).

```bash
$ opa eval -d query_me.rego -d ../../../libraries/common.rego -d ../../../libraries/k8s.rego -i sample.json \
--profile --format=pretty 'data.Cx'

+----------+----------+----------+---------------+
|   TIME   | NUM EVAL | NUM REDO |   LOCATION    |
+----------+----------+----------+---------------+
| 36.658µs | 20       | 20       | query.rego:43 |
| 33.66µs  | 20       | 20       | query.rego:40 |
| 32.229µs | 20       | 20       | query.rego:42 |
| 27.17µs  | 1        | 1        | query.rego:28 |
| 21.29µs  | 1        | 1        | data.Cx       |
| 20.599µs | 8        | 0        | query.rego:36 |
| 15.84µs  | 8        | 8        | query.rego:34 |
| 12.19µs  | 4        | 4        | query.rego:38 |
| 11.94µs  | 8        | 8        | query.rego:35 |
| 11.32µs  | 3        | 6        | query.rego:13 |
+----------+----------+----------+---------------+
```

```bash
$ opa eval -d query_nested.rego -d ../../../libraries/common.rego -d ../../../libraries/k8s.rego -i sample.json \
--profile --format=pretty 'data.Cx'

+-----------+----------+----------+------------------------------------+
|   TIME    | NUM EVAL | NUM REDO |              LOCATION              |
+-----------+----------+----------+------------------------------------+
| 442.537µs | 264      | 203      | ../../../libraries/common.rego:730 |
| 129.307µs | 78       | 132      | ../../../libraries/common.rego:729 |
| 109.327µs | 20       | 20       | query_nested.rego:42               |
| 64.598µs  | 28       | 28       | query_nested.rego:39               |
| 34.43µs   | 20       | 20       | query_nested.rego:41               |
| 32.68µs   | 17       | 18       | ../../../libraries/common.rego:29  |
| 29.488µs  | 8        | 8        | query_nested.rego:34               |
| 22.989µs  | 1        | 1        | data.Cx                            |
| 22.569µs  | 4        | 4        | ../../../libraries/common.rego:737 |
| 20.97µs   | 7        | 6        | ../../../libraries/common.rego:714 |
+-----------+----------+----------+------------------------------------+
```

This may be seem negligible with this tiny sample snippet but with large manifests and when run in combination with all other queries, it will make a difference.

Churro:
Hi @joaoReigota1, thank you for the hint on `searchLine`. I wasn't aware of this property 👍 
I've now revised the rule and added this. As a result, the visual search lines correspond to the reported ones.

Please let me know in case something is still missing to get this ready for v1.5.4.

joaoReigota1 (approved):
LGTM!

PR #4885, opened 2022-02-27
fix(query): Show privilege_escalation_allowed k8s alert also in case no securityContext is defined

Churro:
**Problem**

The `privilege_escalation_allowed` rule is rated high and, indeed, represents a very security-relevant setting. Currently, the rule does not trigger an alert in case no `securityContext` is defined with a container. There is a separate rule to check whether `securityContext` exists but the severity rating is low. Hence, alerts for a rule with high severity are only shown if a user acts upon a low severity rule. 

Therefore, I suggest to change the rule to always show the alert, regardless of the existence of `securityContext`. This is important since a report otherwise does not provide a full picture of issues that demand for action.

**Proposed Changes**

- Make the definition of `securityContext` optional, such that an alert is also shown in case `container.securityContext` is undefined


assets/queries/k8s/privilege_escalation_allowed/query.rego
```diff
@@ -1,41 +1,43 @@
...
 	result := {
 		"documentId": input.document[i].id,
-		"searchKey": sprintf("metadata.name={{%s}}.%s.%s.name={{%s}}.securityContext.allowPrivilegeEscalation", [metadata.name, specInfo.path, types[x], containers[index].name]),
-		"issueType": "IncorrectValue",
-		"keyExpectedValue": sprintf("%s.%s[%s].securityContext.allowPrivilegeEscalation is false", [specInfo.path, types[x], containers[index].name]),
-		"keyActualValue": sprintf("%s.%s[%s].securityContext.allowPrivilegeEscalation is true", [specInfo.path, types[x], containers[index].name]),
+		"searchKey": sprintf("metadata.name={{%s}}.%s.%s.name={{%s}}.securityContext", [metadata.name, specInfo.path, types[x], container.name]),
+		"issueType": "MissingAttribute",
+		"keyExpectedValue": sprintf("metadata.name={{%s}}.%s.%s.name={{%s}}.securityContext.allowPrivilegeEscalation is set and is true", [metadata.name, specInfo.path, types[x], container.name]),
+		"keyActualValue": sprintf("metadata.name={{%s}}.%s.%s.name={{%s}}.securityContext.allowPrivilegeEscalation is undefined", [metadata.name, specInfo.path, types[x], container.name]),
```
rafaela-soares, line 41:
```suggestion
		"keyExpectedValue": sprintf("metadata.name={{%s}}.%s.%s.name={{%s}}.securityContext.allowPrivilegeEscalation is set and is false", [metadata.name, specInfo.path, types[x], container.name]),
		"keyActualValue": sprintf("metadata.name={{%s}}.%s.%s.name={{%s}}.securityContext.allowPrivilegeEscalation is undefined", [metadata.name, specInfo.path, types[x], container.name]),
```

rafaela-soares (changes requested):
Another great observation and refactor, @Churro 🚀 Thank you!

I have two request changes:

Churro:
Suggestions applied, thank you for the thorough review @rafaela-soares! 👍

joaoReigota1 (changes requested):

Churro:
Hi @joaoReigota1, thank you for the review. I've now added `searchLine` and another positive testcase related to that

joaoReigota1 (approved):
LGTM 🚀

rafaela-soares (approved):
LGTM 🚀

PR #4887, opened 2022-02-27
fix(query): k8s rule service_account_token_automount_not_disabled should also consider automount option in ServiceAccount

Churro:
**Problem**

* Since K8s v1.6, `automountServiceAccountToken` can also be defined at Service Account level. Pods may override this setting. The current rule inspects `automountServiceAccountToken` only on pod level and yields false positive alerts in case this setting is defined together with the Service Account.

**Proposed Changes**

* Match `specInfo` with SA definition and check whether the SA defines `automountServiceAccountToken`


assets/queries/k8s/service_account_token_automount_not_disabled/test/negative2.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: redistest-sa
+automountServiceAccountToken: false
+---
+metadata:
```
joaoReigota1, line 7: We need to add a kind here so that the k8s validator passes

joaoReigota1 (changes requested):
Hi @Churro, thank you for contributing to KICS and helping us improve our Kubernetes queries.
Indeed you are correct, other than a change that is required for our k8s files validator, everything seems good to me and I will be happy to merge it

assets/queries/k8s/service_account_token_automount_not_disabled/test/negative2.yaml
Churro: Done, thank you for the hint!

joaoReigota1 (approved):
LGTM

PR #4888, opened 2022-02-27
fix(query): Define NET_BIND_SERVICE as exception in containers_with_added_capabilities k8s rule

Churro:
**Proposed Changes**

- Allow `NET_BIND_SERVICE` as a capability that may be added
  - This is necessary, e.g., to the run nginx ingress controller pods on port 80 and 443 as non-root users, see [here](https://www.nginx.com/blog/announcing-nginx-ingress-controller-for-kubernetes-release-1-6-0/#Other-Improvements)
  - It is also inline with the "Restricted policy" implemented in the upcoming Pod Security Admission Controller, see [here](https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted)


rafaela-soares (approved):
LGTM 🚀

Thank you so much for the amazing observation and contribution, @Churro! We are very grateful!

PR #4891, opened 2022-02-27
fix(query): Extend image_pull_policy_of_container_is_not_always k8s rule to cover additional resource kinds

Churro:
The rule intends to address the problem that image tags are mutable and it is not guaranteed that the version of an image that is cached on a node equals the one with the same tag in a registry. This potential security issue can be prevented by any of the following conditions:

- Explicitly setting `imagePullPolicy` to "Always" to ensure the kubelet downloads the image anew every time it launches the container
- By not specifying `imagePullPolicy` but by providing an image in one the following ways:
  - With tag "latest"
  - By not specifying a tag at all. From the [docs](https://kubernetes.io/docs/concepts/containers/images/#image-names):
    > If you don't specify a tag, Kubernetes assumes you mean the tag latest
  - By specifying a digest with the image

Except for the last sub-item, Kubernetes will apply `imagePullPolicy: Always`. When an image is provided together with a digest, caching by a node is legit and does not trigger the potential security risk this rule strives to detect. Also, this is analogous to what is done by the kubelet with policy "Always" (see [here](https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy). In other words, since the digest associated with an image is immutable, it unambiguously refers to a specific image locally and/or in the registry.

**Problem**

- The rule only covers "Pod" and works with the assumption that `imagePullPolicy` is always defined. Otherwise, the rule aborts early as a statement evalutes to false. This is an issue as the default pull policy is `IfNotPresent` and not `Always`
- The rule does not cover the case that no tag is used or an image is specified with a digest

**Proposed Changes**

- Cover the possibilites described above
- New positive test cases:
  - `imagePullPolicy != "Always"` in a Deployment
  - `imagePullPolicy` not specified but a tag

### DIFF
--- a/assets/queries/k8s/image_pull_policy_of_container_is_not_always/query.rego
+++ b/assets/queries/k8s/image_pull_policy_of_container_is_not_always/query.rego
@@ -1,20 +1,28 @@
 package Cx
 
-CxPolicy[result] {
-	document := input.document
-	metadata := document[i].metadata
-	spec := document[i].spec
-	types := {"initContainers", "containers"}
-	containers := spec[types[x]]
-	containers[c].imagePullPolicy != "Always"
+import data.generic.k8s as k8sLib
 
-	not contains(containers[c].image, ":latest")
+types := {"initContainers", "containers"}
+
+CxPolicy[result] {
+	document := input.document[i]
+	metadata := document.metadata
+
+	specInfo := k8sLib.getSpecInfo(document)
+	container := specInfo.spec[types[x]][_]
+
+	imagePullPolicy := object.get(container, "imagePullPolicy", {})
+	imagePullPolicy != "Always"
+
+	not contains(container.image, "@") # digest
+	not contains(container.image, ":latest") # image with latest tag
+	contains(container.image, ":") # any tag
 
 	result := {
-		"documentId": input.document[i].id,
-		"searchKey": sprintf("metadata.name={{%s}}.spec.%s.name={{%s}}.imagePullPolicy", [metadata.name, types[x], containers[c].name]),
+		"documentId": document.id,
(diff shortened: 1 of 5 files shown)


rafaela-soares:
Hi, @Churro! Thank you so much for your impressive work and contributions to KICS!

I think we can cover the cases where the image is set to the latest, has digest or is undefined but the `imagePullPolicy` is defined and not set to "Always". What do you think?

For example:
```
apiVersion: v1
kind: Pod
metadata:
  name: private-image-test-always
spec:
  containers:
    - name: uses-private-image
      image: gcr.io/google-samples/cassandra@sha256:123456
      imagePullPolicy: Never
```

Churro:
Hi @rafaela-soares,

Probably my explanation in the first post was a bit misleading:
In case a digest is specified together with the image, specifying `imagePullPolicy: Always` may be set to `IfNotPresent` or overall omitted. This is fine since the digest unambiguously identifies an image and unlike a tag, it is not "mutable". 

The likelihood that a locally cached image with digest X, computed from its content, is the same as the one at a remote registry with digest X, computed from its content, is warranted by modern hash functions that are resistant to preimage and collision attacks. 

In other words, there is no match for the sample you posted but that is on purpose because the digest refers to an unambiguous image and the situation this rule intends to detect cannot occur.

If you would specify `Always` instead of `IfNotPresent` but an image **without digest**, the kubelet would compute the hash digest of a locally cached image, compare it with the digest of an image at a remote location and in case of digest mismatch, re-pull the image. Hence, it would use the digest (as the tag could be overwritten) to check if the image contents are the same.

rafaela-soares:
Hi @Churro!

I understood now! Thank you so much for the kind explanation!

rafaela-soares (approved):
LGTM 🚀 Thank you so much for another great contribution!

PR #4892, opened 2022-02-27
fix(query): Extend image_without_digest k8s rule to cover further resource kinds

Churro:
**Proposed Changes**

- Remove `MissingAttribute` check for `image` + adapt the testcase
  - For a container to be deployable, specifying an image is indispensible (see `type Container struct` [here](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go)).
  - Trying to deploy a pod without an image attribute yields an error: `The Pod "private-image-test-1.2" is invalid: spec.containers[0].image: Required value`
  - Note: `image` can be empty (= null) though, e.g., if `command` is also used
- More comprehensive metadata description

### DIFF
--- a/assets/queries/k8s/image_without_digest/metadata.json
+++ b/assets/queries/k8s/image_without_digest/metadata.json
@@ -3,7 +3,7 @@
   "queryName": "Image Without Digest",
   "severity": "LOW",
   "category": "Insecure Configurations",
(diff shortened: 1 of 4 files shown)


rafaela-soares (approved):
LGTM!

Thank you so much for improving the k8s queries, @Churro 🚀 Impressive work!

PR #4893, opened 2022-02-27
fix(query): Fix path to spec in root_container_not_mounted_as_read_only k8s rule

Churro:
**Proposed Changes**

- Fold the check for the existence of `container.securityContext` with the check for `readOnlyRootFilesystem`
- Correct the spec reference `searchKey` to the actual path

### DIFF
--- a/assets/queries/k8s/root_container_not_mounted_as_read_only/metadata.json
+++ b/assets/queries/k8s/root_container_not_mounted_as_read_only/metadata.json
@@ -1,9 +1,9 @@
 {
   "id": "a9c2f49d-0671-4fc9-9ece-f4e261e128d0",
-  "queryName": "Root Container Not Mounted As Read-only",
+  "queryName": "Root Container Not Mounted Read-only",
   "severity": "LOW",
   "category": "Build Process",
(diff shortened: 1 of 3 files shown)


assets/queries/k8s/root_container_not_mounted_as_read_only/metadata.json
```diff
@@ -1,9 +1,9 @@
 {
   "id": "a9c2f49d-0671-4fc9-9ece-f4e261e128d0",
-  "queryName": "Root Container Not Mounted As Read-only",
+  "queryName": "Root Container Not Mounted Read-only",
```
rafaela-soares, line 3: Can you also update the queryName in the `positive_expected_result.json`, please?

rafaela-soares (changes requested):
Hi, @Churro! Thank you so much for another great contribution!

I only have one request:

joaoReigota1 (changes requested):

Churro:
Hi @joaoReigota1, thank you for the review. As requested, I've now added the `searchLine` attribute

joaoReigota1 (approved):

rafaela-soares (approved):
LGTM 🚀

PR #4894, opened 2022-02-27
fix(query): Corrected tag flagging invalid_image k8s rule

Churro:
**Problem**

- The logic to check whether a tag is provided seems flawed. Specifying an empty image is fine, e.g., with `Cronjob` where an empty image tag is provided but shell commands are provided in `command`.
  What this rule most likely wanted to target in the first place is whether the **image tag** is empty (or latest), rather than the image itself.

**Proposed Changes**

- Remove `MissingAttribute` check for `image`
  - For a container to be deployable, specifying an image is indispensible (see `type Container struct` [here](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go)).
  - Trying to deploy a pod without an image attribute yields an error: `The Pod "private-image-test-1.2" is invalid: spec.containers[0].image: Required value`
  - `image` can be empty (= null) though, e.g., if `command` is also used

### DIFF
--- a/assets/queries/k8s/invalid_image/metadata.json
+++ b/assets/queries/k8s/invalid_image/metadata.json
@@ -1,9 +1,9 @@
 {
   "id": "583053b7-e632-46f0-b989-f81ff8045385",
-  "queryName": "Invalid Image",
+  "queryName": "Invalid Image Tag",
   "severity": "LOW",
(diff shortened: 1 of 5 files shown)


joaoReigota1 (approved):
LGTM

PR #4940, opened 2022-03-08
feat(query): Anonymous Auth Is Not Set To False

joaorufi:
Signed-off-by: joaorufi <joao.rufino@checkmarx.com>

**Proposed Changes**
- Added query Anonymous Auth Is Not Set To False


assets/queries/k8s/anonymous_auth_is_not_set_to_false/query.rego
```diff
@@ -0,0 +1,55 @@
+package Cx
+
+import data.generic.common as common_lib
+
+CxPolicy[result] {
+	resource := input.document[i]
+	container := resource.spec.containers[j]
```
Churro, line 7: Please note that `resource.spec.containers` works with Pods but not with `Deployment`, `DaemonSet`, `Cronjob`, etc. as containers are specified in a different hierarchy level. I'd suggest to use `k8sLib.getSpecInfo` instead.

assets/queries/k8s/anonymous_auth_is_not_set_to_false/query.rego
```diff
@@ -0,0 +1,55 @@
...
+		"keyActualValue": "--anonymous-auth flag is set to true",
+	}
+}
+
+CxPolicy[result] {
+	resource := input.document[i]
+	container := resource.spec.containers[_]
+
+	inArray(container.command, "kube-apiserver")
+	not inArray(container.args, "--anonymous-auth=false")
```
Churro, line 42: Just a hint: Changing this to something like
```
containerArgs := object.get(container, "args", {})
not inArray(containerArgs, "--anonymous-auth=false")
```
would enable you to omit the other `MissingAttribute`. It would also ensure that only actionable info is shown (meaning that "should contain args" still doesn't hint what exactly should be specified as args)

joaorufi:
> Hi @joaorufi, I just stumbled upon this PR and hope it's ok for you that I left some suggestions for potential refinement ☺️

Hi @Churro, thanks for the orientations, i'm glad that you are helping us so much!

assets/queries/k8s/anonymous_auth_is_not_set_to_false/query.rego
```diff
@@ -39,6 +39,42 @@ CxPolicy[result] {
...
+	specInfo := k8sLib.getSpecInfo(resource)
+	container := specInfo.spec.containers[j]
+
+	inArray(container.command, "kubelet")
+	containerArgs := object.get(container, "args", {})
+	inArray(containerArgs, "--anonymous-auth=true")
+
+	result := {
+		"documentId": input.document[i].id,
+		"searchKey": sprintf("spec.container[%d].args", [j]),
```
Churro: If you only specify `spec.container[...]`, without `"metadata.name={{%s}}` in front, searchKey will potentially point to the wrong `document`, if a manifest includes multiple resources with containers that have the same name (or index in your case)

assets/queries/k8s/anonymous_auth_is_not_set_to_false/query.rego
```diff
@@ -0,0 +1,45 @@
...
+import data.generic.k8s as k8sLib
+
+CxPolicy[result] {
+	resource := input.document[i]
+	specInfo := k8sLib.getSpecInfo(resource)
+	container := specInfo.spec.containers[j]
+	commands := ["kube-apiserver", "kubelet"]
+
+	inArray(container.command, commands[_])
+	containerArgs := object.get(container, "args", {})
```
rafaela-soares, line 12:
```suggestion

```

I think this line does not make sense here since this rule will return a result if `--anonymous-auth=true` is defined in the `container.args`. So, this rule considers that args is defined in the container 


Additionally that, I think `--anonymous-auth=true` can be defined inside `container.command`. For example:
```
apiVersion: v1
kind: Pod
metadata:
  name: command-demo
  labels:
    purpose: demonstrate-command
spec:
   containers:
      - command:
        - kube-apiserver
        - --anonymous-auth=false
        name: kube-apiserver
```

assets/queries/k8s/anonymous_auth_is_not_set_to_false/query.rego
```diff
@@ -0,0 +1,45 @@
...
+}
+
+CxPolicy[result] {
+	resource := input.document[i]
+	specInfo := k8sLib.getSpecInfo(resource)
+	container := specInfo.spec.containers[j]
+	commands := ["kube-apiserver", "kubelet"]
+
+	inArray(container.command, commands[_])
+	containerArgs := object.get(container, "args", {})
```
rafaela-soares, line 31: Please, use the function `get_nested_values_info` in line 31. It receives two inputs: object and array_vals. This function verifies if the array_vals are nested in the object. It will return an object with two fields: valid (returns if the array_vals are nested in the object (array_vals should be sorted) and searchKey returns the searchKey possible.

For example:

```
object := {"elem1": {"elem2": "elem3"}}
array_vals := ["elem2", "elem3", "elem4"]
return_value := {"valid": false, "searchKey": "elem2.elem3"}
```
You can find an example of the application [here](https://github.com/Checkmarx/kics/blob/master/assets/queries/k8s/containers_run_with_low_uid/query.rego#L42-L50)

rafaela-soares (changes requested):

assets/queries/k8s/anonymous_auth_is_not_set_to_false/query.rego
```diff
@@ -0,0 +1,45 @@
...
+	inArray(containerArgs, "--anonymous-auth=true")
+
+	result := {
+		"documentId": input.document[i].id,
+		"searchKey": sprintf("spec.container[%d].args", [j]),
```
rafaela-soares, line 17:
```suggestion
        "searchKey": sprintf("metadata.name={{%s}}.%s.%s.name={{%s}}.args", [metadata.name, specInfo.path, types[x], container.name]
```

rafaela-soares (changes requested):

rafaela-soares (approved):
LGTM 🚀

PR #4949, opened 2022-03-09
Fix(query): S3 Bucket Policy Accepts Http Requests

cxAndreFelicidade:
**Proposed Changes**
- Added Allow true to deny_http_requests function


rafaela-soares (approved):
LGTM 🚀

PR #4963, opened 2022-03-10
feat(Query): Added Alicloud Public Security Group Rule Sensitive Port for Terraform

cxMiguelSilva:
Closes #

**Proposed Changes**
- Added Alicloud Public Security Group Rule Sensitive Port for Terraform
- Added port 4333: MySQL in tcpPortsMap on common.rego library
- Fixed agregation and positive_expected_result in querys that use  tcpPortsMap on common.rego library


rafaela-soares (approved):
LGTM 🚀

PR #4974, opened 2022-03-13
fix(query): Extend container_requests_not_equal_to_its_limits k8s rule to cover further resource kinds and remove redundant checks

Churro:
**Proposed Changes**

- Remove redundant `MissingAttribute` checks that are covered by other rules. While this rule has a LOW severity rating, checks for missing cpu and memory attributes are rated MEDIUM. 
  - The concerning rules are:
    - [cpu_requests_not_set](https://github.com/Checkmarx/kics/blob/master/assets/queries/k8s/cpu_requests_not_set/query.rego)
    - [cpu_limits_not_set](https://github.com/Checkmarx/kics/blob/master/assets/queries/k8s/cpu_limits_not_set/query.rego)
    - [memory_requests_not_defined](https://github.com/Checkmarx/kics/blob/master/assets/queries/k8s/memory_requests_not_defined/query.rego)
    - [memory_limits_not_defined](https://github.com/Checkmarx/kics/blob/master/assets/queries/k8s/memory_limits_not_defined/query.rego)
  - When kics is run with the full rule set (or better said: together with these 4 rules), it will produce search hits on the same result lines as the ones I removed from `positive_expected_result.json` here.

### DIFF
--- a/assets/queries/k8s/container_requests_not_equal_to_its_limits/metadata.json
+++ b/assets/queries/k8s/container_requests_not_equal_to_its_limits/metadata.json
@@ -3,7 +3,7 @@
   "queryName": "Container Requests Not Equal To It's Limits",
   "severity": "LOW",
   "category": "Resource Management",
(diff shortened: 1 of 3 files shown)


rafaela-soares (approved):
LGTM 🚀 Thank you so much for another contribution!

PR #4976, opened 2022-03-14
feat(analyzer & parser): decrypt Ansible Vault file on the fly

### LINKED ISSUE #4417: Decrypt ansible vault files on the fly

When running Ansible anlaysis, it could be very practical to analyse encrypted files by decrypt these files on the fly and analyse their values. For now, if we want to use a password **admin** we just have to encrypt our password. May be a new configuration like `KICS_ANSIBLE_VAULT_PASSWORD` or just use the `ANSIBLE_VAULT_PASSWORD_FILE` could do the job.

Kics would then decrypt vault files on the fly, analyse their values and add the potentiel security analysis in the report

rafaela-soares:
Closes #4417

**Proposed Changes**
- Decrypt Ansible Vault file on the fly
- Now, KICS can decrypt Ansible Vault files on the fly. For that, the user needs to define the environment variable `ANSIBLE_VAULT_PASSWORD_FILE`.


pkg/utils/ansible_vault.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"regexp"
+
+	vault "github.com/sosedoff/ansible-vault-go"
+)
+
+// DecryptAnsibleVault verifies if the fileContent is encrypted by ansible-vault. If yes, the function decrypts it
+func DecryptAnsibleVault(fileContent []byte, secret string) []byte {
+	match, _ := regexp.MatchString(`^\s*\$ANSIBLE_VAULT.*`, string(fileContent))
```
joaoReigota1, line 11: We should check this error

pkg/utils/ansible_vault.go
```diff
@@ -0,0 +1,20 @@
+package utils
+
+import (
+	"regexp"
+
+	vault "github.com/sosedoff/ansible-vault-go"
+)
+
+// DecryptAnsibleVault verifies if the fileContent is encrypted by ansible-vault. If yes, the function decrypts it
+func DecryptAnsibleVault(fileContent []byte, secret string) []byte {
```
joaoReigota1, line 10: We should log that we are decrypting the Ansible file

joaoReigota1 (changes requested):
Other than some changes looks good to me

joaoReigota1 (approved):
LGTM

PR #4998, opened 2022-03-16
update(docs-generator): added trace to severities

rafaela-soares:
**Proposed Changes**
- Added trace to severities


joaoReigota1 (approved):
LGTM

PR #5008, opened 2022-03-17
feat(query): Added Protect Kernel Defaults Is Not Set To False for Kubernetes

cxMiguelSilva:
**Proposed Changes**
- Added Protect Kernel Defaults Is Not Set To False for Kubernetes

### DIFF
--- a/assets/queries/k8s/protect_kernel_defaults_is_not_set_to_false/metadata.json
+++ b/assets/queries/k8s/protect_kernel_defaults_is_not_set_to_false/metadata.json
@@ -1,0 +1,11 @@
+{
+    "id": "6cf42c97-facd-4fda-b8af-ea4529123355",
+    "queryName": "Protect Kernel Defaults Is Not Set To False",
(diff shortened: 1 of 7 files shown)


assets/queries/k8s/protect_kernel_defaults_is_not_set_to_false/metadata.json
```diff
@@ -0,0 +1,11 @@
+{
+    "id": "6cf42c97-facd-4fda-b8af-ea4529123355",
+    "queryName": "Protect Kernel Defaults Is Not Set To False",
```
rafaela-soares, line 3:
```suggestion
       "queryName": "Kubelet Protect Kernel Defaults Set To False",
```

Please, also update the query folder name and the `positive_expected_result.json`

rafaela-soares (changes requested):

rafaela-soares (changes requested):

PR #5017, opened 2022-03-18
fix(query): Added kubelet config to Authorization Mode Set To Always Allow query

joaorufi:
**Proposed Changes**
- Added kubelet config to Authorization Mode Set To Always Allow query


rafaela-soares (dismissed):
LGTM 🚀

PR #5029, opened 2022-03-22
Feat(Query): Privileged Containers Enabled for Docker Compose

cxAndreFelicidade:
**Proposed Changes**
- Query: Privileged Containers Enabled for Docker Compose


assets/queries/dockerCompose/privileged_containers_enabled/query.rego
```diff
@@ -0,0 +1,18 @@
+package Cx
+
+import data.generic.common as common_lib
+
+CxPolicy[result] {
+	resource := input.document[i]
+	service_parameters := resource.services[name]
+    privileged := service_parameters[name2].privileged
```
rafaela-soares, line 8: Please, do deeper research about where the field `privileged` can be defined. I think this approach is not correct.

Observe the output of the docker-compose, please:

image


Probably, the best approach is the following (Please, confirm it):

```suggestion
    privileged := service_parameters.privileged
```

image

rafaela-soares (changes requested):

rafaela-soares (approved):
LGTM 🚀

joaorufi (approved):

PR #5066, opened 2022-03-28
update(printer): move printer package from internal to use KICS as a module

joaoReigota1:

**Proposed Changes**
- move printer to package instead of internal so that KICS can be used as a module


rafaela-soares (approved):
LGTM 🚀

PR #5074, opened 2022-03-28
update(BoM): update BoM queries and BoM docs

rafaela-soares:
**Proposed Changes**
- split `policy` from `resource_accessibility` (It was decided that KICS does not have the responsibility to interpret the policy. KICS should only display it, when it exists)
- unified some `bom_output` fields
- updated BoM docs
- commented `user_name` and `is_default_policy` (for now, these fields are not used for CE and maybe they will not be used in the future. Anyway, it considers only one user and it can be more than one)


joaoReigota1 (approved):
LGTM

PR #5078, opened 2022-03-28
delete(query): Remove redundant default_service_account_in_use k8s rule

Churro:
**Proposed Changes**

- Remove the rule entirely as it is duplicated by https://github.com/Checkmarx/kics/blob/master/assets/queries/k8s/service_account_token_automount_not_disabled/query.rego
  - In particular, note _default_ fallback in lines that define the variable `serviceAccountName`
- The rule does not take into account that `automountServiceAccountToken` can also be set on pod level (and would thereby override the SA definition), resulting in false positive alerts


rafaela-soares (approved):
LGTM 🚀 

Thank you so much for noticing and reporting, @Churro!

PR #5107, opened 2022-03-31
feat(query): Added Service Account Lookup Set To False for Kubernetes

cxMiguelSilva:
**Proposed Changes**
- Added Service Account Lookup Set To False for Kubernetes


joaorufi (approved):

rafaela-soares (approved):
LGTM 🚀

PR #5195, opened 2022-04-15
fix(query): updated ecr_repositories_not_encrypted TF rule to match KMS encryption

Churro:
**Problem**

The TF rule _ECR Repository Not Encrypted_ issues an alert in case `encryption_configuration` is not specified, assuming there is no encryption at all. This is not correct since encryption-at-rest is enabled out-of-the-box for ECR repositories using AWS managed keys. In the AWS TF provider, this is denoted as `AES256` (see [here](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ecr_repository#encryption_configuration) and [here](https://docs.aws.amazon.com/AmazonECR/latest/userguide/encryption-at-rest.html)).

Practically, this means that the check this rule performs does not match the rule's purpose and it should actually check whether customer-managed keys are used (see [here](https://aws.amazon.com/about-aws/whats-new/2020/07/ecr-supports-encryption-images-aws-kms-keys/?nc1=h_ls)) instead of the default encryption. 

**Proposed Changes**
- Rename the rule to _ECR Repository Not Encrypted With CMK_
- Extend the description
- Check whether `encryption_type = 'KMS'` and a KMS key ARN is specified in `kms_key`
  -  `kms_key` is optional but, if unspecified, an AWS-managed key is used, effectively resulting in the same situation as if no `KMS` but `AES256` was used
- Change severity from MEDIUM to LOW
  - Using a CMK is "best practice" and the security implications of not doing it are not comparable to "not encrypted at all" but mostly needed for compliance (PCI-DSS and friends)

### DIFF
--- a/assets/queries/terraform/aws/ecr_repository_not_encrypted/metadata.json
+++ b/assets/queries/terraform/aws/ecr_repository_not_encrypted/metadata.json
@@ -1,9 +1,9 @@
 {
   "id": "0e32d561-4b5a-4664-a6e3-a3fa85649157",
-  "queryName": "ECR Repository Not Encrypted",
-  "severity": "MEDIUM",
+  "queryName": "ECR Repository Not Encrypted With CMK",
+  "severity": "LOW",
   "category": "Encryption",
-  "descriptionText": "ECR (Elastic Container Registry) Repository encryption should be set",
+  "descriptionText": "ECR repositories should be encrypted with customer-managed keys to meet stricter security and compliance requirements on access control, monitoring, and key rotation",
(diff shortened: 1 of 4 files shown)


rafaela-soares (commented):
Hello, @Churro!

Thank you so much for noticing, reporting, and contributing! We are very glad about your contributions 😊 

LGTM! 🚀

P.S. We will only wait for the AppSec approval related to the severity.

rafaela-soares (approved):
Severity approved!


LGTM 🚀

joaorufi (approved):

PR #5196, opened 2022-04-15
fix(query): added support for aws_iam_policy_document.Principals to policy_without_principal tf rule

Churro:
**Problem**

[iam_policy_document](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document) is an alternative policy specification. When unmarshalled, `Principals` remains as-is and is thus not understood by the `policy_without_principal` rule, which checks only for `Principal`, as in traditional policies.

**Proposed Changes**
- Add a check for `Statement.Principals`
- Add another negative testcase


rafaela-soares (approved):
Another great observation 🚀 

Thank you so much for improving KICS queries!

LGTM!

joaorufi (approved):

PR #5197, opened 2022-04-15
fix(query): adjust severity of iam_access_analyzer_undefined rule to LOW

Churro:
**Proposed Changes**
- Rename `iam_access_analyzer_undefined` rule to `iam_access_analyzer_not_enabled`, as _undefined_ implies a missing definition at a spot, where it would have been necessary. It seems also more inline with the naming of other queries
- Align description to actual purpose of IAM access analyzer
- Adjust rating from MEDIUM to LOW for the following reasons:
  - The main functionality it fulfills is to provide the user a tool to compare permissions granted in IAM policies with those actually used. This is very helpful in identifying and reducing over-privileged policies to a smaller subset with the bare minimum of what is necessary. However, this needs to be done interactively (also see [here](https://docs.aws.amazon.com/IAM/latest/UserGuide/what-is-access-analyzer.html) and [here](https://aws.amazon.com/iam/features/analyze-access/)).
  - After enabling Access Analyzer (as checked by this rule) it doesn't do what is promised in the description of this rule: _identify unintentional access_. Out-of-the box it also doesn't alert about access attempts that were not granted in the first place. To get this, one would have to setup quite a few things around that, e.g., AWS Security Hub and/or EventBridge for SNS notifications (see example [here](https://aws.amazon.com/blogs/security/automate-resolution-for-iam-access-analyzer-cross-account-access-findings-on-iam-roles/)).
  - Unlike other MEDIUM rules, e.g., _Api Gateway Access Logging Disabled_ or _VPC FlowLogs Disabled_, there is no direct security impact if Access Analyzer is not enabled. If access policies are already least privilege, the benefit of using Access Analyzer boils down to monitoring permission sets for changes over time. Even if there are changes then, they may be fully on purpose and only a manual comparison would reveal suspicious things.


cxlucas:
Hi Churro,
This PR found a bug in our E2E Test script.
I'm working on it to identify and solve this bug, as soon as we have this fix on the master, I will let you know.
Thanks very much for your valuable contributions, and sorry for this issue. It's a problem related to the E2E tests, not related to your changes.

rafaela-soares:
Hello, @Churro!

Your suggestions make totally sense! Thank you so much for improving KICS queries 🚀 

I would like to ask you a few things to merge this PR:

- Can you also update the query `kics\assets\queries\cloudFormation\aws\iam_access_analyzer_undefined\metadata.json`, please?
- Can you merge this branch with the master and run the E2E tests? You can find the documentation [here](https://github.com/Checkmarx/kics/blob/master/e2e/README.md). Let us know if you need any help 😊

Churro:
Hi @rafaela-soares, thank you for the hint regarding the CF rule! Seems I overlooked that 😊

I just did what you requested but the e2e tests are failing. That's not really surprising, considering the renaming and severity change. However, this affects, e.g., also E2E-CLI-018, which excludes "Access Control", whereas this rule is now in the "Best Practices" category. Would you please tweak the tests the way you think it will make most sense?

rafaela-soares:
> Hi @rafaela-soares, thank you for the hint regarding the CF rule! Seems I overlooked that 😊
> 
> I just did what you requested but the e2e tests are failing. That's not really surprising, considering the renaming and severity change. However, this affects, e.g., also E2E-CLI-018, which excludes "Access Control", whereas this rule is now in the "Best Practices" category. Would you please tweak the tests the way you think it will make most sense?
> 
> P.S.: There seems to some small glitches on [E2E tests](https://github.com/Checkmarx/kics/blob/master/e2e/README.md):
> 
> * `kics_repository_folder/github/scripts/server-mock` -> `kics_repository_folder/.github/scripts/server-mock` (the dot)
> * `docker build -t kics:e2e-tests -f .\Dockerfile .` -> `\.`  works only on Windows

Hello again, @Churro!

Yes, of course. I tweaked them 😊 (I removed the category "Access Control" without replacing it with "Best Practices" because if I do it, the scan will return no results and the exit code will be 0. The query that was reporting a result in the scan has also "Best Practices" as a category)

Thank you so much for noticing and reporting. We will improve the documentation.

rafaela-soares (approved):
LGTM 🚀

PR #5199, opened 2022-04-15
fix(query): fixed searchKey and resource kind in pod_or_container_without_resource_quota k8s rule

Churro:
**Proposed Changes**

- Change `searchKey` to point to the namespace, where possible, otherwise the document name
- Use `checkKind` and also include `PersistentVolumeClaim`
- Merge the two `MissingAttribute` rules. They still deliver the same matches
- Extend the description text


rafaela-soares (approved):
LGTM 🚀 

Great refactor and improvement!

joaorufi (approved):

PR #5201, opened 2022-04-16
fix(query): extended scope of MissingAttribute rule in seccomp k8s rule

Churro:
**Problem**

#4789 updated the seccomp k8s rule to support `seccompProfile`. For backward compatibility with deprecated annotations, the `MissingAttribute` check only proceeded when a `securityContext` already existed. This introduced an unwanted side effect: if containers didn't specify a `securityContext`, no alert was shown.

There is another rule to check the existence of a `securityContext` but it's only rated LOW, meaning that only after fixing this LOW severity issue, the seccomp issue with MEDIUM severity show up.

**Proposed Changes**
- Add a check to the `MissingAttribute` query to also work without `securityContext`. A new check ensures that `MissingAttribute` is not shown, if annotations are used
- Extended the description text and updated the URL to point to the modern way of using seccomp with K8s
- `MissingAttribute` findings in `positive1.yaml` test: they fell out in #4789 but should actually be reported

### DIFF
--- a/assets/queries/k8s/seccomp_profile_is_not_configured/metadata.json
+++ b/assets/queries/k8s/seccomp_profile_is_not_configured/metadata.json
@@ -3,8 +3,8 @@
   "queryName": "Seccomp Profile Is Not Configured",
   "severity": "MEDIUM",
   "category": "Insecure Configurations",
-  "descriptionText": "Check if any resource does not configure Seccomp default profile properly",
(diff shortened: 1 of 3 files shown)


rafaela-soares (approved):
LGTM 🚀 

Another great observation, as always! Thank you!!

joaorufi (approved):

PR #5209, opened 2022-04-19
Fix(e2e): Results json compare (index out of range)

cxlucas:
Closes #

**Proposed Changes**
- This PR introduces a bug fix in the E2E Tests application that was throwing a panic message in a specific case when changing a query that is used by an E2E test.
- Instead of a panic message, the E2E now breaks the test and displays an error message that indicates where are the differences found in the tests.


joaorufi (approved):

rafaela-soares (approved):
LGTM 🚀

PR #5214, opened 2022-04-19
Feat(e2e): Allow E2E Tests to run locally (dev) and dockerized (CI)

cxlucas:
Closes #

**Proposed Changes**
- The E2E Tests are focused on testing KICS Docker image, but for development purposes, it's very hard to create a docker build every time that a developer needs to run the E2E Tests after any change in the code. To solve this problem, this PR introduces a "switch" from E2E execution where:
- In development mode (locally): it runs the tests using a kics binary (which is much faster than building the docker image).
- In github (on each PR): it runs the tests using the kics docker image generated from the source code.


rafaela-soares (approved):
LGTM 🚀

joaorufi (approved):

PR #5264, opened 2022-04-27
fix(query): adjusted severity rating and added searchLine in rbac_wildcard_in_rule k8s rule

Churro:
**Proposed Changes**

- Add `searchLine` to point to specific rule attributes
- Combine the CxPolicy rules into one
- Extend description
- Adjust rating from LOW to MEDIUM (or HIGH?) for the following reasons:
  - The rule [rbac_roles_with_read_secrets_permissions](https://github.com/Checkmarx/kics/blob/master/assets/queries/k8s/rbac_roles_with_read_secrets_permissions/query.rego) checks for a specific "permission subset" of this rule and  is rated MEDIUM. This rule here implicitly also covers the same case as the other rule by checking for wildcards, which allow even wider access rights
  - Wildcard permissions on resources (pods, depoyments, secrets, ...) also include create, update, patch, delete, bind, escalate, impersonate verbs. Attackers may abuse this to perform mutating changes, e.g., create malicious pods, delete secrets, change role bindings, impersonate users/groups/service accounts. Overall, this can put the availability of deployments at risk, in addition to violating confidentiality and integrity
  - The worst-case setting would be * on all `apiGroups`, `resources`, and `verbs`, essentially meaning cluster admin rights and this would even justify a HIGH severity rating. Depending on what is set to *, the potential blast radius varies. If you think it would be reasonable to assume the worst-case scenario here, HIGH would be appropriate - otherwise at least MEDIUM, I'd say 😊

### DIFF
--- a/assets/queries/k8s/rbac_wildcard_in_rule/metadata.json
+++ b/assets/queries/k8s/rbac_wildcard_in_rule/metadata.json
@@ -1,9 +1,9 @@
 {
   "id": "6b896afb-ca07-467a-b256-1a0077a1c08e",
   "queryName": "RBAC Wildcard In Rule",
-  "severity": "LOW",
+  "severity": "MEDIUM",
   "category": "Access Control",
-  "descriptionText": "Kubernetes Roles and ClusterRoles should not use wildcards in rules (objects or actions)",
+  "descriptionText": "Roles and ClusterRoles with wildcard RBAC permissions provide excessive rights to the Kubernetes API and should be avoided. The principle of least privilege recommends to specify only the set of needed objects and actions",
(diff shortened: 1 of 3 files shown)


rafaela-soares (commented):
Amazing refactor and improvement, @Churro! Thank you so much!

LGTM! I will only wait for the severity approval.

joaorufi (dismissed):

rafaela-soares:
Hello, @Churro! Can you update the severity to HIGH, please?

Churro:
Hi @rafaela-soares, done :smiley:

rafaela-soares (approved):
LGTM! 🚀 

Thank you so much, @Churro!

PR #5265, opened 2022-04-27
fix(query): fixed searchKey rbac_roles_with_read_secrets_permissions …

Churro:
**Proposed Changes**

- Add `searchLine` to point to problematic `verbs` instead of `rules`
- Combine the two CxPolicy rules into one
- Extend the description


rafaela-soares (approved):
LGTM 🚀

PR #5266, opened 2022-04-27
feat(query): add new k8s rule to detect port-forwarding into containers (RBAC)

Churro:
**Proposed Changes**

- Add new rule to check whether RBAC roles allow port-forwarding into pods. This is sometimes useful for troubleshooting but dangerous in production. In case of pod compromise, attackers may be able to bypass network security restrictions or probe container ports directly


rjegoncalves (commented):
LGTM

rafaela-soares (approved):
LGTM 🚀 

Thank you so much for adding new queries, @Churro!

joaorufi (approved):

PR #5267, opened 2022-04-27
feat(query): add new k8s rule to detect account impersonation (RBAC)

Churro:
**Proposed Changes**

- Add new rule to check whether RBAC roles make use of the 'impersonate' verb to allow impersonation of others users, groups, serviceaccounts. This sudo-like functionality is sometimes useful for troubleshooting but very dangerous in production. In case of pod compromise, attackers may be able to impersonate system:admin or other highly privileged users and thereby escalate their privileges to the whole cluster
- This rule covers CIS Kubernetes Benchmark 1.23, 5.1.8


rjegoncalves (commented):
LGTM. 
Severity and Category are appropriate, and additional CIS Kubernetes Benchmark (1.23) coverage added: 5.1.8.

joaorufi (approved):

PR #5268, opened 2022-04-27
feat(query): add new k8s rule to detect bind or escalate permissions (RBAC)

Churro:
**Proposed Changes**

- Add new rule to check whether RBAC roles are allowed to create new Roles or RoleBindings. This is dangerous, as the created roles may include resources and permissions that exceed the privileges of the role that is used to create them.
- This rule covers CIS Kubernetes Benchmark 1.23, 5.1.8

A longer explanation of why this is a problem, can also be found [here](https://raesene.github.io/blog/2021/01/16/Getting-Into-A-Bind-with-Kubernetes/).


rjegoncalves (commented):
LGTM. 
Severity and Category are appropriate and additional CIS Kubernetes Benchmark (1.23) coverage added: 5.1.8.

rafaela-soares (approved):
LGTM 🚀

joaorufi (approved):

PR #5286, opened 2022-05-01
feat(query): add new k8s rule to detect exec permissions (RBAC)

Churro:
**Proposed Changes**

- Add new rule to check whether RBAC roles allow to run 'kubectl exec' to execute commands in containers. This is sometimes useful for troubleshooting but dangerous in production. In case of compromise, attackers may abuse this to subject to execute malicious code

### DIFF
--- a/assets/queries/k8s/rbac_roles_with_exec_permission/metadata.json
+++ b/assets/queries/k8s/rbac_roles_with_exec_permission/metadata.json
@@ -1,0 +1,10 @@
+{
+  "id": "c589f42c-7924-4871-aee2-1cede9bc7cbc",
+  "queryName": "RBAC Roles with Exec Permission",
+  "severity": "MEDIUM",
+  "category": "Access Control",
+  "descriptionText": "Roles or ClusterRoles with RBAC permissions to run commands in containers via 'kubectl exec' could be abused by attackers to execute malicious code in case of compromise. To prevent this, the 'pods/exec' verb should not be used in production environments",
+  "descriptionUrl": "https://kubernetes.io/docs/reference/access-authn-authz/rbac/",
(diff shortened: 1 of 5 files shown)


rjegoncalves (commented):
LGTM

rafaela-soares (approved):
LGTM 🚀

joaorufi (approved):

Dentrax:
Shouldn't we also check for `pods/attach`?

Churro:
Good question, @Dentrax. Do you know of any attack that is particularly related to `pods/attach`?

To my knowledge, the only action it allows is to attach a terminal to a running container. That's, of course, very useful in combination with `pods/exec` (already flagged by this rule) but unless there are further, potentially risky resources/verbs, I cannot think of an immediate security issue due to the attach functionality itself.

Since `pods/exec` also works without tty, I don't see a gain in detecting that `pods/attach` may or may not be specified as resource. Please let me know if you think I'm missing something here.

Dentrax:
Thanks for clarifications, @Churro! Unfortunately, I don't have much knowledge to validate your opinions. AFAICS [^1], we _can't execute_ any command but able to get outputs from processes. We can also send stdin from our terminal to the program. What if a process prints sensitive logs that no one should see? Wouldn't someone do fuzz to process by spamming random stdin inputs?

[^1]: https://stackoverflow.com/a/50031131/5685796

Churro:
Hmm, fuzzing processes via stdin would only work if a process would explicitly wait for input. Also, sensitive information should never be printed to stdout or stderr. Nonetheless, you're right that reality allows for such things to happen and this should be detected. I'll file a PR for `pods/attach` in a moment. Thanks again!

PR #5295, opened 2022-05-02
fix(queries): fixed Docker queries related to issues 5115, 5116, and 5118

### LINKED ISSUE #5115: Update "COPY '--from' Without FROM Alias Defined Previously" query for Docker, allow external images
### Platform
Docker

### Query
`68a51e22-ae5a-4d48-8e87-b01a323605c9`

### Description
Currently, this query detects an issue when a `COPY` instruction references another docker image using `--from` and this image is not defined in the same Dockerfile as part of a multi-step build. The `--from` flag of the `COPY` instruction can also be used to reference external Docker images (https://docs.docker.com/develop/develop-images/multistage-build/#use-an-external-image-as-a-stage). This query should not raise an issue in this case.

### Source
Some Dockerfiles for the official [nats](https://hub.docker.com/_/nats/) image uses this to copy files between the different versions of their docker images and kics reports that this query fails: https://github.com/nats-io/nats-docker/blob/9095670eefc7c5af2ba6400a42ff88097b018c70/2.7.4/scratch/Dockerfile

### LINKED ISSUE #5116: Update "RUN Instruction Using 'cd' Instead of WORKDIR" query for Docker, allow cd when not at the beginning
### Platform
Docker

### Query
`f4a6bcd3-e231-4acf-993c-aa027be50d2e`

### Description
In long `RUN` commands there sometimes exists `cd`s in the middle of the Instruction to change a directory. It is not helpful to replace such a `cd` with a `WORKDIR` instruction as this would create an additional layer. Therefore only `cd`s that are at the beginning of a `RUN` instruction should be reported.

### Example
The official docker image for [fluentd](https://hub.docker.com/_/fluentd/) uses such a construct in the Dockerfile for its debian image: https://github.com/fluent/fluentd-docker-image/blob/a9e4cee765c7aaf7876d6fe3282aa565dbcdc2a4/v1.14/debian/Dockerfile

### LINKED ISSUE #5118: Update "Changing Default Shell Using SHELL Command" query for Docker, query and reasoning are in conflict
### Platform
Docker

### Query
`8a301064-c291-4b20-adcb-403fe7fd95fd`

### Description
The query currently checks that the `SHELL` instruction is not used at all in a Dockerfile. Instead, it suggests changing the shell within every `RUN` instruction. This conflicts with the Dockerfile reference that is mentioned as the reasoning. Here it is instead suggested that the shell is changed using the `SHELL` instruction (https://docs.docker.com/engine/reference/builder/#shell) and **not** in the `RUN` instructions. When using a command to change the shell (eg. `RUN powershell -command`) this leads to inefficiencies. The JSON syntax for the `RUN` instruction could be used to stop this. But the Dockerfile reference instead proposes to use the `SHELL` instruction as then the `RUN` instructions are simpler to read. I was not able to find any reasoning for why the `SHELL` instruction should be avoided.

I would therefore suggest either removing this query, changing it to check that the shell is not changed using the `RUN` instruction or updating the reasoning if there is one.

cxAndreFelicidade:
Closes #5115, closes #5116, closes #5118

**Proposed Changes**
- On issue 5115: Query was changed and files refactored.
- On issue 5116: Query description was changed.
- On issue 5118: Query was changed and files refactored.


assets/queries/dockerfile/changing_default_shell_using_shell_command/metadata.json
```diff
@@ -1,9 +1,9 @@
 {
   "id": "8a301064-c291-4b20-adcb-403fe7fd95fd",
-  "queryName": "Changing Default Shell Using SHELL Command",
+  "queryName": "Changing Default Shell Using RUN Command",
   "severity": "MEDIUM",
   "category": "Insecure Defaults",
-  "descriptionText": "Using the command SHELL to override the default shell instead of the RUN command",
+  "descriptionText": "Using the command RUN to override the default shell instead of the SHELL command leads to ineficiencies. It also does not make sense since Docker provides the SHELL command for this exact purpose.",
```
rafaela-soares, line 6:
```suggestion
  "descriptionText": "Using the command RUN to override the default shell instead of the SHELL command leads to inefficiencies. It also does not make sense since Docker provides the SHELL command for this exact purpose.",
```

assets/queries/dockerfile/changing_default_shell_using_shell_command/query.rego
```diff
@@ -1,14 +1,33 @@
 package Cx
 
+import data.generic.common as common_lib
+
+CxPolicy[result] {
+	resource := input.document[i].command[name][_]
+	resource.Cmd == "run"
+    value := resource.Value
+	contains(value[v], "/bin/bash")
```
rafaela-soares, line 9: Maybe we could refactor this line to something less broad

assets/queries/dockerfile/changing_default_shell_using_shell_command/query.rego
```diff
@@ -1,14 +1,33 @@
...
+		"keyActualValue": sprintf("FROM={{%s}}.{{%s}} uses the RUN command to change the default shell", [name, resource.Original]),
+	}
+}
+
 CxPolicy[result] {
 	resource := input.document[i].command[name][_]
-	resource.Cmd == "shell"
+	resource.Cmd == "run"
+    value := resource.Value
+	contains(value[v], "powershell")
```
rafaela-soares, line 24: Not sure if this is the best approach

assets/queries/dockerfile/copy_from_without_from_alias_defined_previously/metadata.json
```diff
@@ -3,7 +3,7 @@
   "queryName": "COPY '--from' Without FROM Alias Defined Previously",
   "severity": "MEDIUM",
   "category": "Build Process",
-  "descriptionText": "COPY command with the flag '--from' should mention a previously defined FROM alias",
+  "descriptionText": " This query is used to ensure that build stages are named. This way even if the Dockerfile is re-ordered, the COPY instruction doesn’t break.",
```
rafaela-soares, line 6: I am not sure if we should keep this query. Seems more lint than security, imho

rafaela-soares (changes requested):

assets/queries/dockerfile/copy_from_without_from_alias_defined_previously/query.rego
```diff
@@ -1,31 +1,21 @@
...
+    flag_split := split(flags[f], "=")
+    to_number(flag_split[1]) > -1
+	
 
 	result := {
 		"documentId": input.document[i].id,
-		"searchKey": sprintf("FROM={{%s}}.{{%s}}", [name, resource.Original]),
+		"searchKey": sprintf("FROM={{%s}}.{{%s}}", [name, commands.Original]),
 		"issueType": "IncorrectValue",
 		"keyExpectedValue": "COPY '--from' references a previous defined FROM alias",
```
joaorufi, line 18:
```suggestion
		"keyExpectedValue": "COPY '--from' should reference a previous defined FROM alias",
```

assets/queries/dockerfile/changing_default_shell_using_shell_command/query.rego
```diff
@@ -1,14 +1,33 @@
...
+	resource := input.document[i].command[name][_]
+	resource.Cmd == "run"
+    value := resource.Value
+	contains(value[v], "/bin/bash")
+
+	result := {
+		"documentId": input.document[i].id,
+		"searchKey": sprintf("FROM={{%s}}.{{%s}}", [name, resource.Original]),
+		"issueType": "IncorrectValue",
+		"keyExpectedValue": sprintf("FROM={{%s}}.{{%s}} uses the SHELL command to change the default shell", [name, resource.Original]),
```
joaorufi, line 15:
```suggestion
		"keyExpectedValue": sprintf("FROM={{%s}}.{{%s}} should use the SHELL command to change the default shell", [name, resource.Original]),
```

assets/queries/dockerfile/changing_default_shell_using_shell_command/query.rego
```diff
@@ -1,14 +1,33 @@
...
+    value := resource.Value
+	contains(value[v], "powershell")
 
 	result := {
 		"documentId": input.document[i].id,
 		"searchKey": sprintf("FROM={{%s}}.{{%s}}", [name, resource.Original]),
 		"issueType": "IncorrectValue",
-		"keyExpectedValue": sprintf("FROM={{%s}}.{{%s}} uses the RUN command to change the default shell", [name, resource.Original]),
-		"keyActualValue": sprintf("FROM={{%s}}.{{%s}} uses the SHELL command to change the default shell", [name, resource.Original]),
+		"keyExpectedValue": sprintf("FROM={{%s}}.{{%s}} uses the SHELL command to change the default shell", [name, resource.Original]),
```
joaorufi, line 30:
```suggestion
		"keyExpectedValue": sprintf("FROM={{%s}}.{{%s}} should use the SHELL command to change the default shell", [name, resource.Original]),
```

joaorufi (changes requested):

assets/queries/dockerfile/changing_default_shell_using_shell_command/metadata.json
rafaela-soares: This requested change is marked as resolved but the change was not done, @cxAndreFelicidade

joaorufi (approved):

joaorufi (dismissed):

rafaela-soares (changes requested):

rafaela-soares (approved):
LGTM 🚀

PR #5334, opened 2022-05-11
fix(password and secrets): improve performance

rafaela-soares:
**Proposed Changes**
- Improved Password and Secrets performance: since Password and Secrets run the queries against the original file, there is no point to consider all the `fileMetadatas.` We can only consider one `fileMetadata` per `filePath`.


joaoReigota1 (approved):
LGTM

PR #5368, opened 2022-05-18
feat(query): Ensure Administrative Boundaries Between Resources for Kubernetes

cxMiguelSilva:
**Proposed Changes**
- Ensure Administrative Boundaries Between Resources


rjegoncalves (commented):
The Severity is OK.
The Category should be Access Control.

assets/queries/k8s/ensure_administrative_boundaries_between_resources/query.rego
```diff
@@ -0,0 +1,26 @@
...
+
+CxPolicy[result] {
+	resource := input.document[i]
+	metadata := resource.metadata
+
+    resource.kind == listKinds[_]
+    namespace := metadata.namespace
+
+    c := count({j | res = input.document[j]; res.kind == listKinds[_]; res.metadata.namespace == namespace})
+    c > 1
```
rafaela-soares, line 16: Not sure if this approach is the best. 

When a namespace is not defined, by default, the namespace is "default", which does not fit the best practice of limiting the scope. Maybe we should advise the users to not use a default namespace.

Additionally, I think that using the same namespace more than once is not a vulnerability.  As mentioned in the CIS WorkBench Kubernetes documentation, "Ensure that these namespaces are the ones you need and are adequately administered as per your requirements".  I think we should advise the user to review its namespaces, in case the user has many and in case the user has a few.


Anyway, please be careful with the typos in `keyExpectedValue` and `keyActualValue` (probably it will be changed)

assets/queries/k8s/ensure_administrative_boundaries_between_resources/metadata.json
```diff
@@ -0,0 +1,11 @@
+{
+    "id": "e84eaf4d-2f45-47b2-abe8-e581b06deb66",
+    "queryName": "Ensure Administrative Boundaries Between Resources",
```
rafaela-soares, line 3: We need to improve the `queryName`

rafaela-soares (changes requested):

assets/queries/k8s/ensure_administrative_boundaries_between_resources/query.rego
```diff
@@ -0,0 +1,26 @@
...
+
+	result := {
+		"documentId": input.document[i].id,
+		"searchKey": sprintf("metadata.namespace={{%s}}", [metadata.namespace]),
+		"issueType": "IncorrectValue",
+		"keyExpectedValue": "consider using diffrent namespace",
```
joaorufi, line 22:
```suggestion
		"keyExpectedValue": "consider using different namespace",
```

assets/queries/k8s/ensure_administrative_boundaries_between_resources/query.rego
```diff
@@ -0,0 +1,26 @@
...
+		"keyActualValue": "same namaspace is used used multiple times",
```
joaorufi, line 23:
```suggestion
		"keyActualValue": "same namaspace is used multiple times",
```

joaorufi (changes requested):

joaorufi (approved):

PR #5371, opened 2022-05-18
fix(cpu): fix number cpus macos

### LINKED ISSUE #5364: Getting CPU info fails on macOS Docker

- **Please make sure to:**
  - When attaching files to the issue make sure they are properly formatted

### Expected Behavior

No error on getting CPU info.

### Actual Behavior

```
Scanning with Keeping Infrastructure as Code Secure v1.5.8


9:01PM INF Scanning with Keeping Infrastructure as Code Secure v1.5.8
9:01PM INF Operating system: linux
9:01PM INF Total memory: 3.8G
9:01PM ERR failed to get CPU error="open /sys/fs/cgroup/cpu/cpu.cfs_quota_us: no such file or directory"
```

1.  `docker run -it -w /path -v $PWD:/path checkmarx/kics:v1.5.8 scan -p /path -v`

### Specifications
(N/A if not applicable)

-   Version: 1.5.8
-   Platform: macOS 12.4
-   Subsystem: Docker Desktop 4.7.0 (77141)

Related to #5321

joaorufi:
Signed-off-by: joaorufi <joao.rufino@checkmarx.com>

Closes #5364

**Proposed Changes**
- Look at `/sys/fs/cgroup/cpu.max` file if `/sys/fs/cgroup/cpu/cpu.cfs_quota_us` file does not exists

### DIFF
--- a/internal/console/pre_scan.go
+++ b/internal/console/pre_scan.go
@@ -148,12 +148,8 @@ func (console *console) preScan() {
 		log.Info().Msgf("Total memory: %s", bytefmt.ByteSize(mem.Total))
 	}
 
-	cpu, err := consoleHelpers.GetNumCPU()
-	if err != nil {
-		log.Err(err).Msg("failed to get CPU")
-	} else {
-		log.Info().Msgf("CPU: %.1f", cpu)
-	}
+	cpu := consoleHelpers.GetNumCPU()
+	log.Info().Msgf("CPU: %.1f", cpu)
 
 	noProgress := flags.GetBoolFlag(flags.NoProgressFlag)
 	if strings.EqualFold(flags.GetStrFlag(flags.LogLevelFlag), "debug") {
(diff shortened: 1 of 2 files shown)


rafaela-soares:
@joaorufi, please, take a look at the two extra SAST vulnerabilities

internal/console/pre_scan.go
```diff
@@ -148,12 +148,8 @@ func (console *console) preScan() {
...
 	}
 
-	cpu, err := consoleHelpers.GetNumCPU()
-	if err != nil {
-		log.Err(err).Msg("failed to get CPU")
-	} else {
-		log.Info().Msgf("CPU: %.1f", cpu)
-	}
+	cpu := consoleHelpers.GetNumCPU()
+	log.Info().Msgf("CPU: %.1f", cpu)
```
rafaela-soares, line 152: Should we print the CPU even if it is set to -1?

rafaela-soares (changes requested):

internal/console/pre_scan.go
joaorufi: It will never happen that the value is -1. When I return -1 on the function, it also return an error and if there is an error, it tries to get the cpus number in other ways until the last case, returning `float32(runtime.NumCPU())`. This will never be -1

PR #5491, opened 2022-06-18
feat(query): add new k8s rule to detect attach permission (RBAC)

Churro:
**Proposed Changes**

- Add new rule to check whether RBAC roles allow to attach to contains via 'kubectl attach'. This is useful for debugging but in case of compromise, an attacker could misuse the permission to read stdout, stderr and write stdin. In a worst-case scenario, log output may include sensitive information and the control flow of a running process could be tainted.

More information [here](https://stackoverflow.com/a/50031131/5685796). Thanks to @Dentrax for drawing my attention to this.

### DIFF
--- a/assets/queries/k8s/rbac_roles_with_attach_permission/metadata.json
+++ b/assets/queries/k8s/rbac_roles_with_attach_permission/metadata.json
@@ -1,0 +1,10 @@
+{
+  "id": "df4a3b93-53a4-4e8a-8d3b-3e91d0a664af",
+  "queryName": "RBAC Roles with Attach Permission",
+  "severity": "MEDIUM",
+  "category": "Access Control",
+  "descriptionText": "Roles or ClusterRoles with RBAC permissions to attach to containers via 'kubectl attach' could be abused by attackers to read log output (stdout, stderr) and send input data (stdin) to running processes. To prevent this, the 'pods/attach' verb should not be used in production environments",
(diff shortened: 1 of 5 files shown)


kaplanlior:
The unit test didn't finish, we'll check why to see if it's a service/CI issue or bug that needs to be fixed in the code.

rafaela-soares:
Hello, @Churro! Thank you so much for taking care of this additional suggestion!

Can you push an empty commit only to verify if the GitHub action will be targeted, please?

rafaela-soares (dismissed):
LGTM 🚀 Thank you so much for another contribution!

I will only wait for our security team's approval.

rjegoncalves:
We reviewed the proposed new rule and associated Severity, Category, and Description. The only suggestion we have is related to the Description. @nunolope5 did an analysis on the possible impacts created by 'kubectl attach', and in the case of a privileged container, it would be possible to escalate privileges. The new Description suggestion was added as a comment to the _metadata.json_ of the query.

Churro:
Thanks for the review, @rjegoncalves! I appreciate your cross-check and the extended description text you suggested. The most recent commit here includes it 🚀

rafaela-soares (approved):
LGTM! Thank you so much, @Churro 🚀

rafaela-soares:
@Dentrax, thank you so much for raising this.

rafaela-soares:
@Dentrax, thank you so much for raising this.

Dentrax:
Also thanks to @erkanzileli who noticed and reported the concern to us. ❤️

Issue #4584, opened 2021-12-06
Use .gitignore to automatically exclude paths


jgrumboe:
**Is your feature request related to a problem? Please describe.**
I tested kics for Terraform code and it will also report problems it finds in the external modules I use in my code.
To be more precise it scans also `.terraform` folders and files found there.

**Describe the solution you'd like**
I already ignore `.terraform` folders in my `.gitignore` and would like kics to understand/respect the definitions there - automatically or via cmdline argument.

**Describe alternatives you've considered**
I can define the folders in this case also directly using `kics scan -e '.terraform,*/.terraform' -p .`
But for future improvements, I think it would be good to support .gitignore.

anterosilva1985:
Hi @jgrumboe,

First of all i'm sorry for the delay on the feedback. Also i would like to thanks you on this input. 
Regarding this feature request, i've decided to include it on Kics 1.6 version. 

Regards,
Antero Silva
Product Manager - Kics

rafaela-soares:
Hi, @jgrumboe!

Thank you so much for the suggestion. KICS 1.6 version implements it 😊

PR #5506, opened 2022-06-21
feat(scan): consider .gitignore to automatically exclude paths by default

rafaela-soares:
Issue #4584 will be closed when the branch `release/1.6` is merged into the master.

**Proposed Changes**
- consider `.gitignore` to automatically exclude paths by default


roi-orca:
@rafaela-soares I would suggest having the default behavior as of today, meaning .gitignore content will be taken into consideration as part of the scan, and the newly added flag, which can be called `"--include-gitignore"` will exclude the content from the scan.

If we do that, we do not break the current usage of the CLI, and we are having a "strict by default" approach

docs/commands.md
```diff
@@ -106,6 +107,9 @@ Global Flags:
 
 The other commands have no further options.
 
+## Exclude Paths
+By default, KICS now automatically excludes paths according to the .gitignore file. If you want to disable this behavior, you use the flag `--add-gitignore`.
```
nunoocx, line 111: Change to
By default, KICS excludes paths specified in the .gitignore file in the root of the repository. To disable this behavior, use flag `--add-gitignore`.

docs/dockerhub.md
```diff
@@ -53,6 +53,7 @@ Usage:
   kics scan [flags]
 
 Flags:
+      --add-gitignore                 disables the usage of .gitignore file to automatically exclude paths
```
nunoocx, line 56: disables the exclusion of paths specified within .gitignore file

e2e/fixtures/assets/scan_help
```diff
@@ -2,6 +2,7 @@ Usage:
   kics scan [flags]
 
 Flags:
+      --add-gitignore                 disables the usage of .gitignore file to automatically exclude paths
```
nunoocx, line 5: same as above

internal/console/assets/scan-flags.json
```diff
@@ -185,5 +185,11 @@
     "defaultValue": "",
     "usage": "case insensitive list of platform types to scan\n(${supportedPlatforms})",
     "validation": "validateMultiStrEnum"
+  },
+  "add-gitignore": {
+    "flagType": "bool",
+    "shorthandFlag": "",
+    "defaultValue": "false",
+    "usage": "disables the usage of .gitignore file to automatically exclude paths"
```
nunoocx, line 193: same as above

nunoocx (changes requested):
Please review the suggestions

rafaela-soares:
> @rafaela-soares I would suggest having the default behavior as of today, meaning .gitignore content will be taken into consideration as part of the scan, and the newly added flag, which can be called `"--include-gitignore"` will exclude the content from the scan.
> 
> If we do that, we do not break the current usage of the CLI, and we are having a "strict by default" approach

Hello, @roi-orca!

Thank you so much for being involved in this question and for your suggestion.

It was a product decision to exclude paths specified within .gitignore file by default. A few users suggested that since the .gitignore file presents the files that will be not committed. Since it can be considered a breaking change, this new behavior will be only included from version 1.6. To disable this behavior, you can use flag `--add-gitignore`.

docs/commands.md
```diff
@@ -43,6 +43,7 @@ Usage:
   kics scan [flags]
 
 Flags:
+      --add-gitignore                 disables the exclusion of paths specified within .gitignore file
```
nunoocx, line 46: Hi @rafaela-soares , revisiting this, the flag add-gitignore sounds misleading.
Taking the suggestion from @roi-orca (and considering that the default will be to include gitignore) WDYT if we change to `--exclude-gitignore` ?
If agreed, please make the remainder of the code consistent to that new naming.

nunoocx (changes requested):

nunoocx (approved):
Great work!

PR #5533, opened 2022-06-26
fix(scan behavior): ignore broken synlink

### LINKED ISSUE #5512: Failed to exclude missing files from scan
when you have a symbolic link to a file that is either missing(removed for some reason) or link to a file that is out of the attached docker volume, kics stops the scan and throw some error such as:

> Preparing Scan Assets: Done
Error: failed to read sources: failed to walk directory: failed to open file: open blabla.json: no such file or directory

I would prefer that in that case, we will just show a log that the file is missing and continue with the scan. 
In addition, the bug is when I tried to exclude that file from the scan using '-e' flag I get the same error like it wasn't affecting at all.

for example, let's say we have a file named a.yaml and we create a symlink to that file called b.yaml(in the same directory). if for some reason a.yaml is removed and we run kics like:
`kics scan --disable-secrets  -p $PWD -e $PWD/b.yaml`
then we are getting the error like above.

liorj-orca:
Closes #5512

**Proposed Changes**
-ignores any broken symlink and continues with the scan 

Future changes are better taken into consideration: 
- I would add a flag to let the customer decide which mode he would like to run "fail open"/"fail close" - meaning fail/stop the scan on any broken/damaged file/dir or continue with the scan and add the list of the damaged files to the output.

### DIFF
--- a/pkg/engine/provider/filesystem.go
+++ b/pkg/engine/provider/filesystem.go
@@ -90,6 +90,24 @@ func (s *FileSystemSourceProvider) GetBasePaths() []string {
 	return s.paths
 }
 
+// ignoreDamagedFiles checks whether we should ignore a damaged file from a scan or not.
+func ignoreDamagedFiles(path string) bool {
+	shouldIgnoreFile := false
+	fileInfo, err := os.Lstat(path)
+	if err != nil {
+		log.Warn().Msgf("Failed getting the file info for file '%s'", path)
+		return shouldIgnoreFile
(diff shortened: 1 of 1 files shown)


liorj-orca:
@rafaela-soares please review :)

rafaela-soares (approved):
LGTM 🚀 Thank you so much, @liorj-orca! 🚀

joaoReigota1 (approved):
LGTM

PR #5570, opened 2022-07-08
feat(cli & engine): kics auto remediation support

rafaela-soares:
**Proposed Changes**
- added KICS AUTO REMEDIATION
- added E2E tests:
     **E2E-CLI-057**: tests if fix command fixes all remediation in a given context
     **E2E-CLI-058**: tests if fix command fixes the remediation pointed in the include-ids flags
     **E2E-CLI-059**: tests the fix help command output
     **E2E-CLI-059**: tests the fix command output without results flag
- added unit tests
- added integration tests
- added KICS AR documentation

### DIFF
--- a/assets/queries/terraform/alicloud/cmk_is_unusable/query.rego
+++ b/assets/queries/terraform/alicloud/cmk_is_unusable/query.rego
@@ -17,6 +17,8 @@ CxPolicy[result] {
 		"keyExpectedValue": sprintf("alicloud_kms_key[%s].is_enabled to be set to true", [name]),
 		"keyActualValue": sprintf("alicloud_kms_key[%s].is_enabled is set to false", [name]),
         "searchLine": common_lib.build_search_line(["resource", "alicloud_kms_key", name, "is_enabled"], []),
+		"remediation": "is_enabled = true",
+		"remediation_type": "replacement",
 	}
 }
 
@@ -34,5 +36,7 @@ CxPolicy[result] {
 		"keyExpectedValue": sprintf("alicloud_kms_key[%s].is_enabled to be set to true", [name]),
 		"keyActualValue": sprintf("alicloud_kms_key[%s].is_enabled is not set", [name]),
         "searchLine": common_lib.build_search_line(["resource", "alicloud_kms_key", name], []),
+		"remediation": "is_enabled = true",
+		"remediation_type": "addition",
 	}
 }
--- a/assets/queries/terraform/alicloud/cmk_is_unusable/query.rego
+++ b/assets/queries/terraform/alicloud/cmk_is_unusable/query.rego
@@ -8,6 +8,8 @@ CxPolicy[result] {
 
 	resource.is_enabled == false
 
+	remediation := {"before": "false", "after": "true"}
+
 	result := {
 		"documentId": input.document[i].id,
 		"resourceType": "alicloud_kms_key",
@@ -17,8 +19,8 @@ CxPolicy[result] {
 		"keyExpectedValue": sprintf("alicloud_kms_key[%s].is_enabled to be set to true", [name]),
 		"keyActualValue": sprintf("alicloud_kms_key[%s].is_enabled is set to false", [name]),
         "searchLine": common_lib.build_search_line(["resource", "alicloud_kms_key", name, "is_enabled"], []),
-		"remediation": "is_enabled = true",
-		"remediation_type": "replacement",
+		"remediation": json.marshal(remediation),
+		"remediationType": "replacement",
 	}
 }
 
@@ -37,6 +39,6 @@ CxPolicy[result] {
 		"keyActualValue": sprintf("alicloud_kms_key[%s].is_enabled is not set", [name]),
         "searchLine": common_lib.build_search_line(["resource", "alicloud_kms_key", name], []),
 		"remediation": "is_enabled = true",
-		"remediation_type": "addition",
+		"remediationType": "addition",
 	}
 }
--- a/docs/kics-auto-remediation.md
+++ b/docs/kics-auto-remediation.md
@@ -1,0 +1,30 @@
+# KICS AUTO REMEDIATION
+
+With this new feature, KICS provides auto remediation for simple replacements and simple additions in a single line.
+
+Note that this feature will be only available for Terraform, for now.
+
+<p align="center">
+<img width="950" alt="image" src="https://user-images.githubusercontent.com/74001161/177953750-3d279868-8cdb-44c9-86f2-379b05bb85d4.png">
+</p>
+
+
+
+## HOW KICS AR WORKS?
+
+
+1. As a first step, you will need to scan your project/file and generate a JSON report.
+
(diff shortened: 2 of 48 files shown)


assets/queries/terraform/alicloud/cmk_is_unusable/query.rego
```diff
@@ -8,6 +8,8 @@ CxPolicy[result] {
 
 	resource.is_enabled == false
 
+	remediation := {"before": "false", "after": "true"}
```
nunoocx, line 11: A suggestion (for future versions) would be to refactor this JSON construction into a function, for instance:
`replace(A,B)`, whose output would be exactly `json.marshal({"before":A, "after":B})`
And for other cases
`addProperty(Prop,Val)`, whose output would be exactly `Prop: Val`

Or we can be even further and create a library of constant remediations...
`replaceTrueWithFalse()` that would return exactly `json.marshal({"before":"true", "after":"false"}`

This would reduce the repeated code throughout queries

docs/kics-auto-remediation.md
```diff
@@ -0,0 +1,30 @@
+# KICS AUTO REMEDIATION
+
+With this new feature, KICS provides auto remediation for simple replacements and simple additions in a single line.
```
nunoocx, line 3: auto remediation for simple replacements and simple additions in a single line 
---> 
auto remediation for single-line replacements and additions

docs/kics-auto-remediation.md
```diff
@@ -0,0 +1,30 @@
+# KICS AUTO REMEDIATION
```
nunoocx, line 1: KICS AUTO REMEDIATION
--->
KICS AUTO REMEDIATION (KICS AR)

docs/kics-auto-remediation.md
```diff
@@ -0,0 +1,30 @@
...
+
+Note that this feature will be only available for Terraform, for now.
+
+<p align="center">
+<img width="950" alt="image" src="https://user-images.githubusercontent.com/74001161/177953750-3d279868-8cdb-44c9-86f2-379b05bb85d4.png">
+</p>
+
+
+
+## HOW KICS AR WORKS?
```
nunoocx, line 13: HOW TO USE KICS AR

docs/kics-auto-remediation.md
```diff
@@ -0,0 +1,30 @@
...
+
+
+1. As a first step, you will need to scan your project/file and generate a JSON report.
+
+   Example: ```docker run -v /home/cosmicgirl/:/path/ kics scan -p /path/sample.tf -i "41a38329-d81b-4be4-aef4-55b2615d3282,a9dfec39-a740-4105-bbd6-721ba163c053,2bb13841-7575-439e-8e0a-cccd9ede2fa8" --no-progress -o /path/results --report-formats json```
```
nunoocx, line 18: Do we really need to add the -i parameter?
It may cause confusion, since if the query doesn't have auto-remediation suggestion, the AR will be calculated anyway, right?

docs/kics-auto-remediation.md
```diff
@@ -0,0 +1,30 @@
...
+   
+   If KICS makes available a remediation for a result, the result will have the fields `remediation` and `remediation_type` defined. As an example, please see:
+   <p align="center">
+   <img width="700" alt="image" src="https://user-images.githubusercontent.com/74001161/177957089-7007d5c0-aea5-4f3a-8300-7008ab0e6312.png">
+   </p>
+   
+   
+2. If your JSON report has any result with remediation, you will need to run the new KICS command: fix. 
+
+   If you want KICS to remediate all the remediation, you can run ```docker run -v /home/cosmicgirl/:/path/ kics fix --results /path/results/results.json -v```.
```
nunoocx, line 28: If you want KICS to remediate all the remediation
--->
To fix the code with all remediation suggestions,

docs/kics-auto-remediation.md
```diff
@@ -0,0 +1,30 @@
...
+   
+   If you want to specify which remediation KICS should fix, you can use the flag `--include-ids`. In this flag, you should point the `similarity_id` of the result. For example: ```docker run -v /home/cosmicgirl/:/path/ kics fix --results /path/results/results.json --include-ids "f282fa13cf5e4ffd4bbb0ee2059f8d0240edcd2ca54b3bb71633145d961de5ce" -v```
```
nunoocx, line 30: If you want to specify which remediation KICS should fix,
--->
To fix the code only with some remediation suggestions,

e2e/fixtures/assets/fix_help
```diff
@@ -0,0 +1,18 @@
+Usage:
+  kics fix [flags]
```
nunoocx, line 2: I'd suggest to change this command to **remediate**, WDYT?
By changing this, everywhere "fix" appears should be revised.

nunoocx (changes requested):
Is there any penalty in performance to calculate the remediation by default in every scan?
I assume not, because it is pretty much static "calculation"... but in the future for more complex remediation should we consider calculate it on demand only?

docs/kics-auto-remediation.md
rafaela-soares: No. It is only an example. I will delete it. Thank you

assets/queries/terraform/alicloud/cmk_is_unusable/query.rego
rafaela-soares: > A suggestion (for future versions) would be to refactor this JSON construction into a function, for instance: `replace(A,B)`, whose output would be exactly `json.marshal({"before":A, "after":B})` And for other cases `addProperty(Prop,Val)`, whose output would be exactly `Prop: Val`
> 
> Or we can be even further and create a library of constant remediations... `replaceTrueWithFalse()` that would return exactly `json.marshal({"before":"true", "after":"false"}`
> 
> This would reduce the repeated code throughout queries

Yes, I totally agree. Thank you so much for the suggestions.

e2e/fixtures/assets/fix_help
rafaela-soares: For me, it is a good suggestion. I will change it.

pkg/remediation/remediation.go
```diff
@@ -0,0 +1,162 @@
...
+		})
+
+		for i := range remediationSet.Addition {
+			a := remediationSet.Addition[i]
+			remediatedLines := addition(a, &lines)
+			if len(remediatedLines) > 0 && willRemediate(remediatedLines, filePath, &a) {
+				lines = s.writeRemediation(remediatedLines, lines, filePath, a.SimilarityID)
+			}
+		}
+	}
```
joaoReigota1, line 84: I think there is no need to sort here we can transverse the slice backward allowing for less complexity

pkg/remediation/scan.go
```diff
@@ -0,0 +1,259 @@
...
+	tmpFile   string
+	files     model.FileMetadatas
+}
+
+// scanTmpFile scans a temporary file against a specific query
+func scanTmpFile(tmpFile, queryID string, remediated []byte) ([]model.Vulnerability, error) {
+	// get payload
+	files, err := getPayload(tmpFile, remediated)
+
+	os.Remove(tmpFile)
```
joaoReigota1, line 44: we should check this error and return a warning in case there as an issue, but not break the flow

joaoReigota1 (changes requested):

joaoReigota1 (approved):
LGTM

nunoocx (approved):
LGTM

PR #5574, opened 2022-07-10
fix(keyExpectedValue): convert to a recommendation rather than a current status

liorj-orca:
I propose changing the "keyExpectedValue" field to be more of a recommendation than what it sounds like today as the current status.

I will open some more PRs like this, covering most of the queries.
Closes #

**Proposed Changes**
-
-
-


rafaela-soares (approved):
LGTM 🚀 

Thank you so much for improving it, @liorj-orca!

PR #5591, opened 2022-07-13
fix(keyExpectedValue): ansible-gcp queries convert to a recommendation rather than a current status

liorj-orca:
Closes #

**Proposed Changes**
-
-
-

### DIFF
--- a/assets/queries/ansible/gcp/cluster_labels_disabled/query.rego
+++ b/assets/queries/ansible/gcp/cluster_labels_disabled/query.rego
@@ -18,7 +18,7 @@ CxPolicy[result] {
 		"resourceName": task.name,
 		"searchKey": sprintf("name={{%s}}.{{%s}}", [task.name, modules[m]]),
 		"issueType": "MissingAttribute",
-		"keyExpectedValue": sprintf("%s is defined and not null", [modules[m]]),
+		"keyExpectedValue": sprintf("%s should be defined and not null", [modules[m]]),
 		"keyActualValue": sprintf("%s is undefined and null", [modules[m]]),
 	}
 }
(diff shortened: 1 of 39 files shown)


assets/queries/ansible/gcp/cluster_labels_disabled/query.rego
```diff
@@ -18,7 +18,7 @@ CxPolicy[result] {
 		"resourceName": task.name,
 		"searchKey": sprintf("name={{%s}}.{{%s}}", [task.name, modules[m]]),
 		"issueType": "MissingAttribute",
-		"keyExpectedValue": sprintf("%s is defined and not null", [modules[m]]),
+		"keyExpectedValue": sprintf("%s should be defined and not null", [modules[m]]),
 		"keyActualValue": sprintf("%s is undefined and null", [modules[m]]),
```
kaplanlior, line 22: Should we also change the KeyActualValue from "undefined **and** null" to "undefined **or** null "?
See assets/queries/ansible/gcp/cluster_master_authentication_disabled/query.rego as a reference.

liorj-orca: @kaplanlior you are more than right, but I decided to go over all the keyExpectedValue fields first and I'll go over all the keyActualValue fields in another PR if it's ok bu you.

kaplanlior: Sure, no problem with splitting that to another PR. Just want to make sure this doesn't get forgotten.

kaplanlior:
We have a few instances of "should differ from 'none' ".
Would "should not be 'none'" be better phrasing?

rafaela-soares (dismissed):
LGTM 🚀 

Thank you so much, @liorj-orca!

cxMiguelSilva (approved):
LGTM
TYSM for your contribution @liorj-orca !!

PR #5651, opened 2022-07-29
feat(engine): Kubernetes API support for runtime k8s clusters scan​

rafaela-soares:
**Proposed Changes**
- added Kubernetes API support for runtime k8s clusters scan​


docs/kuberneter.md
```diff
@@ -0,0 +1,302 @@
...
+Examples:
+
+
+#### Using Config File
+```sh
+docker run -v <credentials_path>:/credentials -v ${PWD}:/path/ -e K8S_CONFIG_FILE=/credentials/<config-file-name> checkmarx/kics:latest scan -p "kuberneter::*:*:*" -v --no-progress -o /path/results
+```
+
+#### Using Service Account Token
+```sh
```
nunoocx, line 145: Something looks off and repeated from this point to the end...

pkg/scan/utils.go
```diff
@@ -174,10 +187,12 @@ func logLoadingQueriesType(types []string) {
 	log.Info().Msgf("Loading queries of type: %s", strings.Join(types, ", "))
 }
 
-func extractPathType(paths []string) (regular, terraformer []string) {
+func extractPathType(paths []string) (regular, terraformer, k8s []string) {
```
nunoocx, line 190: use kuberneter instead of k8s

nunoocx (changes requested):

docs/kuberneter.md
rafaela-soares: My bad. It was not supposed

rafaela-soares: My bad. It was not supposed. Sorry.

gabriel-cx (approved):
LGTM

PR #5696, opened 2022-08-12
fix(docker parser): added resolver for args

### LINKED ISSUE #5621: Add support for variables inbase image name
### Context

Because we want to analyse some multistage docker builds or just use variables in the process, we'd like kics to analyse correctly our variables. 

### Expected Behavior

Kics should send an error on the tag latest of the image

### Actual Behavior

With the right variables, kics could miss some warnings on tags of the base image. 

1. Have the following dockerfile : 

```dockerfile
# syntax=docker/dockerfile:1
ARG BASE_IMAGE=alpine
ARG BASE_IMAGE_TAG=latest

FROM ${BASE_IMAGE}:${BASE_IMAGE_TAG} as main
```

2. Run kics analysis
```bash
docker run -t -v "$PWD/kics":/path checkmarx/kics scan -p "/path" -o "/path/"
```
3.  You do not get the following warning `Image Version Using 'latest', Severity: MEDIUM`

While you get the warning with the following docker file : 

```dockerfile
# syntax=docker/dockerfile:1
ARG BASE_IMAGE=alpine

FROM ${BASE_IMAGE}:latest as main
```

rafaela-soares:
Closes #5621

**Proposed Changes**
- added resolver for args

### DIFF
--- a/pkg/parser/docker/parser.go
+++ b/pkg/parser/docker/parser.go
@@ -3,6 +3,7 @@ package docker
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"strings"
 
 	"github.com/Checkmarx/kics/pkg/model"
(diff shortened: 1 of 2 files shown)


cxMiguelSilva (approved):
LGTM

PR #5706, opened 2022-08-17
fix(query): improve RegEx rule in curl_or_wget_instead_of_add

### LINKED ISSUE #5699: Query "Curl or Wget Instead of Add" wrongly detects some filenames as urls
### Expected Behavior

File names like "httpd-foreground" or "./drop-http-proxy-header.conf" are not detected as URLs.

### Actual Behavior

The following ADD statements are wrongly detected as smelly by this query:

```Dockerfile
ADD ./drop-http-proxy-header.conf /etc/apache2/conf-available
```

```Dockerfile
ADD httpd-foreground /usr/local/bin/
```

The query seems only to check if the sequence "http" or "https" can be found anywhere in the `ADD` statement not if it is actually an URL. It might be better to check if it also includes `://` directly afterward. (see https://github.com/Checkmarx/kics/blob/master/assets/queries/dockerfile/curl_or_wget_instead_of_add/query.rego#L18)

### Specifications

-   Version:  snapshot-34973e9d
-   Platform: Docker
-   Subsystem: Dockerfiles
-   Query: 4b410d24-1cbe-4430-a632-62c9a931cf1c

cxMiguelSilva:
Closes #5699 

**Proposed Changes**
- improve RegEx to check `://` in http/https check inside query

### DIFF
--- a/assets/queries/dockerfile/curl_or_wget_instead_of_add/metadata.json
+++ b/assets/queries/dockerfile/curl_or_wget_instead_of_add/metadata.json
@@ -3,7 +3,7 @@
   "queryName": "Curl or Wget Instead of Add",
   "severity": "LOW",
   "category": "Best Practices",
-  "descriptionText": "Use Curl or Wget instead of Add to fetch packages from remote URLs, because using Add is strongly discouraged",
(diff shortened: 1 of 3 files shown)


rafaela-soares (changes requested):
Maybe we should add a negative sample. What do you think?

rafaela-soares (approved):
LGTM 🚀

PR #5707, opened 2022-08-17
fix(query): update_instruction_alone

### LINKED ISSUE #5703: Query "Update Instruction Alone" false detection for multiple update/install commands
### Expected Behavior

Multiple `apt-get update` and `apt-get install` commands can be used in a single `RUN` instruction.

### Actual Behavior

For the following two Dockerfiles the query "Update Instruction Alone" reports a problem. There is no update command that is not followed by an install command in these Dockerfiles.

```Dockerfile
FROM alpine

RUN if ! which gpg ; then \
      (apt-get update ;apt-get install --no-install-recommends gnupg=2.2.19 -y ) ; \
    fi ; \
    if ! gpg --version | grep -q '^gpg (GnuPG) 1\.' ; then \
      (apt-get update ;apt-get install --no-install-recommends dirmngr=2.2.19 -y ) ; \
    fi ; \
    rm -rf /var/lib/apt/lists/*
```

```Dockerfile
FROM node:19.7.0
RUN npm install sqlite3; apt-get update ;apt-get install --no-install-recommends g++;
```

### Specifications

-   Version:  snapshot-34973e9d
-   Platform: Docker
-   Subsystem: Dockerfiles
-   Query: 9bae49be-0aa3-4de5-bab2-4c3a069e40cd

cxMiguelSilva:
Closes #5703 

**Proposed Changes**
- fix update_instruction_alone to check multiple occurrences of update/install in the command

### DIFF
--- a/assets/queries/dockerfile/update_instruction_alone/query.rego
+++ b/assets/queries/dockerfile/update_instruction_alone/query.rego
@@ -13,7 +13,7 @@ CxPolicy[result] {
 		"documentId": input.document[i].id,
 		"searchKey": sprintf("FROM={{%s}}.RUN={{%s}}", [name, resource.Value[0]]),
 		"issueType": "IncorrectValue",
-		"keyExpectedValue": "Instruction 'RUN <package-manager> update' is followed by 'RUN <package-manager> install' ",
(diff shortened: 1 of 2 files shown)


rafaela-soares (changes requested):
Maybe we should add a negative sample. What do you think?

rafaela-soares (approved):
LGTM 🚀

PR #5728, opened 2022-08-23
fix(queries): add check for ec2messages actions for iam policy grants full permissions

### LINKED ISSUE #5714: TF IaC - AWS IAM Policy Check not Compatible with Resources that Don't support ARNs
Hi guys, me again.

I believe I've found an issue with this check that's running on my Terraform infrastructure code:

Check if an IAM policy is granting full permissions to resources from the get-go, instead of granting permissions gradually as necessary.

The check is flagging on resources which do not support identification by ARN. For example [Amazon Message Delivery Service](https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonmessagedeliveryservice.html). Relevant excerpt from AWS documentation:

> Amazon Message Delivery Service does not support specifying a resource ARN in the Resource element of an IAM policy statement. To allow access to Amazon Message Delivery Service, specify “Resource”: “*” in your policy.

### Expected Behavior

KICS should not flag this as a critical SAST finding, because AWS's recommended (and, in fact, the only) way to allow actions on these resources in an IAM policy is by using `"Resource": "*"`

### Actual Behavior

KICS flags these policies as critical vulnerabilities

### Example policy that gets flagged but should not:

```
{
  "Version": "2012-10-17",
  "Statement": [
    {
        "Effect": "Allow",
        "Action": [
            "ec2messages:*"
        ],
        "Resource": "*"
    }
  ]
}
```

cxMiguelSilva:
Closes #5714 

**Proposed Changes**
- add check for ec2messages actions for IAM policy grants full permissions due to only ec2messages only support wildcard in resources
- add negative samples


assets/queries/ansible/aws/iam_policy_grants_full_permissions/query.rego
```diff
@@ -15,6 +15,7 @@ CxPolicy[result] {
 
 	common_lib.is_allow_effect(statement)
 	common_lib.equalsOrInArray(statement.Resource, "*")
+	common_lib.not_only_ec2messages_actions(statement.Action)
```
rafaela-soares, line 18: I think we should look to the CF query approach and implement it in ANS and TF query

rafaela-soares (changes requested):

rafaela-soares (changes requested):

gabriel-cx (approved):
LGTM

PR #5737, opened 2022-08-26
fix(bom): fix bug in cloudformation bill of materials

cxMiguelSilva:
**Proposed Changes**
- fix bug in cloudformation bill of materials
-
-


gabriel-cx (approved):
LGTM

PR #5743, opened 2022-08-29
fix(query): rds db instance iam auth not enabled

cxMiguelSilva:
**Proposed Changes**
- add lower to engine version check 


gabriel-cx (approved):
LGTM

PR #5783, opened 2022-09-12
fix(github_action): add tflint config file

cxMiguelSilva:
**Proposed Changes**
- add tflint config file


gabriel-cx (approved):
LGTM

PR #5786, opened 2022-09-12
K8s files are not being considered in the scan

gabriel-cx:
**Proposed Changes**
- renderHelm output paths update with the root folder name


joaoReigota1 (approved):
LGTM

PR #5358, opened 2022-05-17
feat(query): added "App Service Without Latest PHP Version" query for Terraform Azure 

cxAndreFelicidade:
**Proposed Changes**
- Query: Php Version Not Latest When Running Web App

Notes: 
- Query severity and category need review.
- Commenting after a value in the samples results in the whole line being ignored (when using it in the search key).

`application_stack{

      php_version = "7.4"

    } `
The php_version line gets parsed.

`application_stack{

      php_version = "7.4" # this is a comment

    } `
The php_version line is ignored.

I submit this contribution under the Apache-2.0 license.

PR #5823, opened 2022-09-22
feat(workflows): add workflow to check latest software versions

cxMiguelSilva:
**Proposed Changes**
- add workflow to check latest software versions


.github/scripts/update_versions/update.py
```diff
@@ -0,0 +1,56 @@
...
+# Gets data from apiUrl argument and check if current version is outdated and returns new versions if applicable
+def checkVersion(name, apiUrl, currentVersion):
+    # make Http Get request
+    r = requests.get(apiUrl)
+
+    #checks if request is successful
+    if r.status_code == 200:
+
+        # get data from request
+        request_version = r.json()[0]['latest']
```
rafaela-soares, line 12: As discussed [here](https://github.com/Checkmarx/kics/pull/5358#discussion_r878156912), we should keep the latest stable version. Does the API URL chosen points only stable releases?

.github/scripts/update_versions/update.py
```diff
@@ -0,0 +1,56 @@
+import json
+import requests
+# Gets data from apiUrl argument and check if current version is outdated and returns new versions if applicable
```
rafaela-soares, line 3:
```suggestion
# Gets data from apiUrl argument and checks if current version is outdated and returns new versions if applicable
```

rafaela-soares (changes requested):

.github/scripts/update_versions/update.py
cxMiguelSilva: In the end-of-life API not in pre-release or in-development releases appear, so I think the returned value is what we expect.

rafaela-soares (approved):
LGTM 🚀

gabriel-cx (approved):
LGTM

PR #5829, opened 2022-09-23
update(docs): improve information in  the configuration docs

VladMasarik:
**Proposed Changes**
- Docs suggestions. When I was trying to setup docs, I had to use the online docs, and finding out anything was an awful pain especially because I had to search through the code.

I did not create an issue first because having something concrete to comment about is faster than explaining my suggestions, and then implementing them. The PR is of course not final, I would like to first know you thoughts, improvements, and I have a few questions. When all is done, I would adjust the JSON, TOML etc. parts as well.

Questions:
- Why is the section "Using commands on scanned files as comments" in "Running KICS" rather than configuration section?
- Does `verbse` flag work? When we set it `verbose=false` there was no difference with `verbose=true` that we could see.
- `type: "type of queries to use in the scan"` does that require list of queries?

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -66,6 +66,7 @@ Flags:
       --exclude-severities strings    exclude results by providing the severity of a result
                                       can be provided multiple times or as a comma separated string
                                       example: 'info,low'
+                                      possible values: 'high, medium, low, info, trace'
       --fail-on strings               which kind of results should return an exit code different from 0
                                       accepts: high, medium, low and info
                                       example: "high,low" (default [high,medium,low,info])
--- a/docs/configuration-file.md
+++ b/docs/configuration-file.md
@@ -2,7 +2,9 @@
 
 KICS allow you to provide all configurations either as command line arguments or as code.
 
-Here is a Configuration file example:
+  > 📝 &nbsp; flags that can receive multiple values can be either provided as a comma separated string or an array as in the example above
+
+## Examples
 
 ```JSON
 {
@@ -35,7 +37,6 @@ exclude-paths:
 output-path: "results"
 ```
 
-> 📝 &nbsp; flags that can receive multiple values can be either provided as a comma separated string or an array as in the example above
 
 ---
 
--- a/docs/configuration-file.md
+++ b/docs/configuration-file.md
@@ -1,6 +1,6 @@
 ## Configuration File
 
-KICS allow you to provide all configurations either as command line arguments or as code.
+KICS allow you to provide all configurations either as command line arguments or as code. You can see all possible configurations in the [CLI](commands.md#Scan-Command-Options).
 
 KICS supports JSON, TOML, YAML, and HCL formats for the configuration files, and it is able to infer the format without the need of file extension.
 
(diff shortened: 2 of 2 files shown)


VladMasarik:
Also, I am happy to do more changes in the docs in this PR if consistency with other docs would be an issue, just let me know.

rafaela-soares (commented):
Hello, @VladMasarik!

Thank you so much for using KICS and also contributing 🚀 

You can find a more detailed documentation about the scan flags [here](https://github.com/Checkmarx/kics/blob/master/docs/commands.md#scan-command-options).

VladMasarik:
> You can find a more detailed documentation about the scan flags [here](https://github.com/Checkmarx/kics/blob/master/docs/commands.md#scan-command-options).

Oh so that is where all the explanations are! Completely my opinion, but wouldn't it be better to have this information in the configuration part as well? Or at least I think that a link in the configuration saying ... "Full configuration options can be found in CLI, and if you want to disable kics in some parts of files see Running KICS", would immensely help. I would say duplicating the information would be good as well, but I don't think it is worth the extra maintenance from your side.

Also, another topic would be could KICS have only docs for actually "getting started" in the getting started section? Rather than basically all the docs? Eg. I would expect maybe installation, running KICS examples and references to the actual configuration, and how to read results than all this: 
image

Then again I think this would be too much conversation in a single PR.

rafaela-soares:
> > You can find a more detailed documentation about the scan flags [here](https://github.com/Checkmarx/kics/blob/master/docs/commands.md#scan-command-options).
> 
> Oh so that is where all the explanations are! Completely my opinion, but wouldn't it be better to have this information in the configuration part as well? Or at least I think that a link in the configuration saying ... "Full configuration options can be found in CLI, and if you want to disable kics in some parts of files see Running KICS", would immensely help. I would say duplicating the information would be good as well, but I don't think it is worth the extra maintenance from your side.

We totally agree with you! The configuration documentation should be definitely improved. We can present an example of a configuration and refer the command's doc as you suggested. Do you mind updating?

I would like to thank you for noticing, taking the time to contribute, and give your feedback. 



> Also, another topic would be could KICS have only docs for actually "getting started" in the getting started section? Rather than basically all the docs? Eg. I would expect maybe installation, running KICS examples and references to the actual configuration, and how to read results than all this: image

We will discuss this matter with the Product Manager. As soon as we have some decision, I will let you know. Once again, thank you so much for your feedback 🚀 

> Then again I think this would be too much conversation in a single PR.

You can use the GitHub discussions or contact us via `kics@checkmarx.com`, if you want 😊

VladMasarik:
I have added all the changes that I have "imagined". Let me know what you think, and what would need to be adjusted. :+1:

rafaela-soares (approved):
Thank you so much, @VladMasarik! It seems great!

Regarding the "Getting Started" section on the website, we will try to take care of it as soon as possible. We agree that should be improved. Thank you so much for the suggestion!

VladMasarik:
@rafaela-soares 
> Regarding the "Getting Started" section on the website, we will try to take care of it as soon as possible.

I dont think you need to rush it, but it was somewhat confusing to see all of the documentation in the "Getting Started" section.

> The verbose flag gives information about the logs (WRN, INF, ...). If you need help with that, let me know. For example:

Maybe I was making a mistake somewhere, but I could swear that I still saw these WARN messages. Although, we did enable only WARN and above warning messages.
image

Still thank you for such a quick responses!

PR #5831, opened 2022-09-23
fix(query): remove string check in open api security query

### LINKED ISSUE #5747: Update invalid_format query for OpenAPI
### Platform
OpenAPI-General

### Description
This query: https://github.com/Checkmarx/kics/blob/master/assets/queries/openAPI/general/invalid_format
lists a number of valid formats for a string type field. However, the [OpenAPI documentation says](https://swagger.io/docs/specification/data-models/data-types/#format) string format is an open value, and can contain anything, not just the list in the query.

I believe this check should be removed for strings as it does not violate the OpenAPI specification and causes false positives.

cxMiguelSilva:
Closes #5747 

**Proposed Changes**
- remove string format check for security query "Invalid Format"

### DIFF
--- a/assets/queries/openAPI/general/invalid_format/metadata.json
+++ b/assets/queries/openAPI/general/invalid_format/metadata.json
@@ -3,7 +3,7 @@
   "queryName": "Invalid Format (v3)",
   "severity": "LOW",
   "category": "Insecure Configurations",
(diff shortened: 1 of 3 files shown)


rafaela-soares (approved):
LGTM 🚀

PR #5832, opened 2022-09-23
fix(query): update s3 bucket policy accepts http requests security query

### LINKED ISSUE #5763: S3 Bucket HTTP access false positive

- **Please make sure to:**
 Having an explicit "DENY" policy within a statement without 
``` 
"Condition": {
        "Bool": {
          "aws:SecureTransport": "false"
        }
      }
``` 
will result in a false positive stating that the bucket allows HTTP traffic, even if there is another explicit DENY statement which includes this condition
  N/A
  Example policy:
```
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Sid": "AllowSSLRequestsOnly",
      "Action": "s3:*",
      "Effect": "Deny",
      "Resource": [
        "${bucket_arn}",
        "${bucket_arn}/*"
      ],
      "Condition": {
        "Bool": {
          "aws:SecureTransport": "false"
        }
      },
      "Principal": "*"
    },
    {
      "Effect": "Deny",
      "Sid": "DenyActionsFromOutsideVpc",
      "Principal": "*",
      "Action": [
        "s3:*
      ],
      "Resource": [
        "arn:aws:s3:::${bucket_name}",
        "arn:aws:s3:::${bucket_name}/*"
      ],
      "Condition": {
        "StringNotLike": {
          "aws:SourceVpc": "${vpc_id}"
        }
      }
    }
  ]
}
```

### Expected Behavior

Policy should check if any one of the deny policies within the statement contain the condition to reject non-HTTPS connection

### Actual Behavior
The policy appears to check that each one of the deny statements contains the condition to reject non-HTTPS connection

1.  Create an s3 bucket
2.  Append policy with multiple deny statements - at least one of which denies non-https traffic (includes condition block) and at least one of which does not

### Specifications
N/A

cxMiguelSilva:
Closes #5763

**Proposed Changes**
- update  s3 bucket policy accepts http requests security to unmarshal JSON before the function
- add a negative sample

### DIFF
--- a/assets/queries/terraform/aws/s3_bucket_policy_accepts_http_requests/query.rego
+++ b/assets/queries/terraform/aws/s3_bucket_policy_accepts_http_requests/query.rego
@@ -9,7 +9,8 @@ CxPolicy[result] {
 	resourceType := resources[r]
 	resource := input.document[i].resource[resourceType][name]
 
-	not deny_http_requests(resource.policy)
+	policy_unmarshaled := common_lib.json_unmarshal(resource.policy)
+	not deny_http_requests(policy_unmarshaled)
 
 	result := {
 		"documentId": input.document[i].id,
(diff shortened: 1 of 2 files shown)


rafaela-soares (approved):
LGTM 🚀

PR #5833, opened 2022-09-23
fix(query): updated lambda_function_with_privileged_role

rafaela-soares:
**Proposed Changes**
- lambda_function_with_privileged_role was failing with 'multiple outputs for same input' when the policy had multiple statements


cxMiguelSilva (approved):
LGTM

PR #5834, opened 2022-09-23
fix(query): fix responses with wrong http status code security query

### LINKED ISSUE #5805: Open API queries - 'default' response type is not handled
Query "Responses With Wrong HTTP Status Code (v3)" checks for int response (in the range of [200-599]) **while 'default' is a valid option**

See https://swagger.io/docs/specification/describing-responses/
image

cxMiguelSilva:
Closes #5805 

**Proposed Changes**
- responses with wrong http status code


rafaela-soares (approved):
LGTM 🚀

PR #5863, opened 2022-09-30
fix(query): correct GCP KMS crypto key rotation period queries + descriptions

Churro:
**Problem**

- The same query differs for Ansible and TF queries regarding the checked criteria: > 7776000, < 7776000
- The Ansible testcase is wrong as it checks the opposite

**Proposed Changes**

- Streamline both queries
- Extend description to include reasoning


cxMiguelSilva (approved):
LGTM

rafaela-soares (approved):
LGTM 🚀 Thank you so much, @Churro!

PR #5864, opened 2022-09-30
fix(query): remove redundant and flawed GCP KMS key rotation query

Churro:
**Problem**

- Currently, the rule to check the key rotation period GCP KMS keys exists twice:
  - [high_google_kms_crypto_key_rotation_period](https://github.com/Checkmarx/kics/tree/master/assets/queries/terraform/gcp/high_google_kms_crypto_key_rotation_period) checks if the period is > 90d
  - [high_kms_rotation_period](https://github.com/Checkmarx/kics/tree/master/assets/queries/terraform/gcp/high_kms_rotation_period) checks if the period is > 365d
    - The included description "_KMS Rotation Period should be greater than 365 days_." is **wrong**. It advises the opposite of what is implemented and also contradicts general recommendations on key lifetime.
- For SOC2 compliance, rotation of encryption keys at least every 90 days is mandatory. Frequent rotation is also advisable to reduce the potential blast radius in case keys are leaked, as attackers could only use it on data that was encrypted while the key has been active and not yet  rotated.

**Proposed Changes**

- Remove the rule entirely, as recommendations conflict with [high_google_kms_crypto_key_rotation_period](https://github.com/Checkmarx/kics/tree/master/assets/queries/terraform/gcp/high_google_kms_crypto_key_rotation_period) and the 90d period meets stricter compliance requirements.


rafaela-soares:
Hello, @Churro!

Thank you so much for noticing! I will only wait for the AppSec approval.

rjegoncalves:
Hi Johannes (@Churro),

Thank you very much for your contribution! 
I agree with your points and suggestions for these queries. 
@rafaela-soares, I believe we can move forward with it.

rafaela-soares (approved):

PR #5867, opened 2022-10-01
fix(query): cover additional deprecated API versions in k8s rule

Churro:
**Proposed Changes**

- Add more CRD deprecations to the rule (previously mentioned here: https://github.com/Checkmarx/kics/pull/4830#pullrequestreview-881390156)
- Extend description
- Change category to "Best Practices": Using a Kubernetes resource with a deprecated API version is not a security risk, comparable to others in that category (e.g. seccomp profile missing). Newer versions are not "more secure", only less likely to cause problems during K8s upgrades.


Churro:
Done, thanks for the hint, @rafaela-soares 👍

Churro:
Done, thanks for cross-checking 😊

rafaela-soares (approved):
LGTM 🚀

PR #5907, opened 2022-10-04
feat(automatic_version_update): Add Version Check From Common Data

cxMiguelSilva:
**Proposed Changes**
- Add Version Check From Common Data

### DIFF
--- a/assets/queries/terraform/azure/app_service_without_latest_php_version/query.rego
+++ b/assets/queries/terraform/azure/app_service_without_latest_php_version/query.rego
@@ -7,7 +7,7 @@ import data.generic.terraform as tf_lib
 CxPolicy[result] {
 	resource := input.document[i].resource.azurerm_app_service[name]
 	php_version := resource.site_config.php_version
-    to_number(php_version) != 8.1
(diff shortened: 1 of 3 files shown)


gabriel-cx (dismissed):
LGTM

assets/queries/terraform/azure/app_service_without_latest_php_version/query.rego
```diff
@@ -7,7 +7,7 @@ import data.generic.terraform as tf_lib
 CxPolicy[result] {
 	resource := input.document[i].resource.azurerm_app_service[name]
 	php_version := resource.site_config.php_version
-    to_number(php_version) != 8.1
+    to_number(php_version) != to_number(common_lib.get_version("php"))
```
rafaela-soares, line 10: Not sure if we can adopt this approach because Terraform allows specific values for `php_version`. Check the doc: https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/app_service

rafaela-soares (changes requested):

PR #5943, opened 2022-10-17
fix(query): update CloudWatch Log Group Without KMS Security Query MetaData

### LINKED ISSUE #5862: cloudwatch_log_group encryption rule is misleading or incorrect
As [documented by AWS](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/encrypt-log-data-kms.html),

> Log group data is always encrypted in CloudWatch Logs. By default, CloudWatch Logs uses server-side encryption for the log data at rest. As an alternative, you can use AWS Key Management Service for this encryption. If you do, the encryption is done using an AWS KMS customer managed key. Encryption using AWS KMS is enabled at the log group level, by associating a key with a log group, either when you create the log group or after it exists.

However, **kics** has a rule that necessitates the usage of KMS for encryption.

https://github.com/Checkmarx/kics/blob/cd24fef396d5c9b954cfc7995496f151e86b0683/assets/queries/terraform/aws/cloudwatch_log_group_not_encrypted/metadata.json#L3-L4

I would argue that this may constitute a severity of `"INFO"` if it is necessary at all.

At the very least it is disingenuous to suggest that the cloudwatch log groups are not encrypted if an AWS KMS key is not applied.

cxMiguelSilva:
Closes #5862

**Proposed Changes**
- update query name
- update query severity

### DIFF
--- a/assets/queries/terraform/aws/cloudwatch_log_group_not_encrypted/metadata.json
+++ b/assets/queries/terraform/aws/cloudwatch_log_group_not_encrypted/metadata.json
@@ -1,7 +1,7 @@
 {
   "id": "0afbcfe9-d341-4b92-a64c-7e6de0543879",
-  "queryName": "CloudWatch Log Group Not Encrypted",
-  "severity": "HIGH",
+  "queryName": "CloudWatch Log Group Without KMS",
+  "severity": "MEDIUM",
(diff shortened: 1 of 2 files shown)


rafaela-soares (approved):
LGTM 🚀

PR #5984, opened 2022-11-04
fix(query): improved regex of "Asymmetric private key"

rafaela-soares:
**Proposed Changes**
- improved regex of "Asymmetric private key"


gabriel-cx (approved):
LGTM

PR #5988, opened 2022-11-08
feat(bom): added Cassandra for CloudFormation

rafaela-soares:
**Proposed Changes**
- added Cassandra support for CloudFormation (there is no resource for AWS Terraform)


cxMiguelSilva (changes requested):
Add checks for accessibility like DynamoDB CF BoM

cxMiguelSilva (approved):
LGTM

PR #6006, opened 2022-11-18
fix(query): update with version regex for unpinned package version in apk add security query

### LINKED ISSUE #6000: False positive on Unpinned Package Version in Apk Add result when libstdc++ is pinned

- **Please make sure to:**
  - When attaching files to the issue make sure they are properly formatted

### Expected Behavior
KICS should report no finding

### Actual Behavior
KICS reports Unpinned Package Version in Apk Add

Removing the `++` from the `libstdc++` package makes the results come out correctly, but then the package cannot be installed.

I tried to fix it locally in the `libraries/dockerfile.rego` in the `withVersion` rules but failed to find a working regex even after validating that the string in step 2 below validates properly at regexpal.com with the following regex: `[A-Za-z0-9_+-]+=(.+)`

You can see it doesn't validate in the below screenshot with the rule in the master branch, as the text is not highlighted by the linter:
image

You can see that it validates properly when removing the `++` part from `libstdc++` with the same rule, as the text is now highlighted by the linter:
image

Finally, you can see it validates with the `++` part in `libstdc++` once I add the `+` character in between `_-` as in the regex above, as the text is highlighted by the linter:
image

Please help me to find the right fix.

1.  Create a Dockerfile based on Alpine
2.  Add `RUN apk --quiet --update --no-cache add libstdc++==11.2.1_git20220219-r2`
3.  Run `kics scan --config configfile.yaml`

### Specifications
(N/A if not applicable)

-   Version: 1.2.6
-   Platform: MacOS
-   Subsystem: dockerfile?

cxMiguelSilva:
Closes #6000 

**Proposed Changes**
- update with version regex in dockerfile.rego to support packages with '+' in name

### DIFF
--- a/assets/libraries/dockerfile.rego
+++ b/assets/libraries/dockerfile.rego
@@ -46,15 +46,15 @@ getCommands(commands) = output{
 }
 
 withVersion(pack) {
-	regex.match("[A-Za-z0-9_-]+[-:][$](.+)", pack)
+	regex.match("[A-Za-z0-9_\\+-]+[-:][$](.+)", pack)
 }
 
 withVersion(pack) {
-	regex.match("[A-Za-z0-9_-]+[:-]([0-9]+.)+[0-9]+", pack)
+	regex.match("[A-Za-z0-9_\\+-]+[:-]([0-9]+.)+[0-9]+", pack)
 }
 
 withVersion(pack) {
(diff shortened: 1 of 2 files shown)


tspearconquest:
LGTM

gabriel-cx (approved):
LGTM

PR #6052, opened 2022-12-14
Refactor: analyzer.go enhancements

ohaval:
Hey kics team,

When working a bit with the `analyzer.go` file, I found some logical flows and tiny code improvements.
I wasn't sure if to open an issue for that, so I am just opening this small PR. Please let me know if there is something you need me to do.

In short:

- Fixed a weird check for an ansible vault using regex, and returning nothing if found (which is the opposite of what we would have been thinking should happen).
- A missing double quotes in a regex pattern.
- Changed regex expressions to use a backtick instead of `"` to avoid the double escape of the `\` character.
- Removed a check at the start of each regex expression of `\\s*` as when using `regex.Match`, as it doesn't search only at the start of the searched string. [Reference](https://pkg.go.dev/regexp#Match)
- Small typo


cxMiguelSilva (approved):
Hi @ohaval,
Thank you so much for your amazing contribution!! 🚀
The proposed changes LGTM

joaoReigota1 (approved):

PR #6057, opened 2022-12-16
added check for RDS backup_retention_period default

joaoReigota1:


assets/libraries/common.rego
```diff
@@ -710,3 +710,4 @@ user_unrecommended_permission_policy_scenarios(targetUser, permission) {
 	# verify that the policy is unrecommended
 	unrecommended_permission_policy(resourcePolicy, permission)
 }
+is
```
cxMiguelSilva, line 713: Maybe we should remove the line 713
```suggestion
```

cxMiguelSilva (changes requested):
The changes in the query.rego file LGTM.
We should remove the `is` added to `common.rego` file
And delete the `negative3.tf` and `negative4.tf` since they are positive samples now.

cxMiguelSilva (approved):
LGTM

PR #6187, opened 2023-02-22
azure-public-storage-account

Tohar-orca:
**Proposed Changes**
- Add AzureResourceManager, Ansible, and Terraform detections for Azure Storage Accounts which allow public access


cxMiguelSilva (changes requested):
Hi @Tohar-orca,
Thank you for another amazing contribution!
After talking with our application security team, we come to the agreement that queries should be aligned with the metadata of the existing one for ARM [Default Azure Storage Account Network Access Is Too Permissive](https://github.com/Checkmarx/kics/tree/master/assets/queries/terraform/azure/default_azure_storage_account_network_access_is_too_permissive). In the case of the ARM query, we believe that an update of the existing one is more appropriate.
The security queries name should be set to "Default Azure Storage Account Network Access Is Too Permissive" with the Severity set to "MEDIUM" and category to "Access Control"

Tohar-orca:
Thank you for your feedback @cxMiguelSilva , I've made the required changes and the PR is ready for re-review

cxMiguelSilva (approved):
Hi @Tohar-orca,
Thank you for the update on your amazing contribution!
The proposed changes LGTM

PR #6190, opened 2023-02-23
fix(query): fix missing detection of used security_group on eks_cluster

Meroje:
**Proposed Changes**
- Adds detection of security_group under an eks_cluster's vpc_config


cxMiguelSilva (approved):
Hi @Meroje, 
Thank you so much for this contribution!
The proposed changes LGTM.

Issue #140, opened 2020-11-16
Add Stackdriver Logging is Disabled query for Terraform


cx-ruben-silva:
### Platform
Terraform

### Provider
GCP

### Description
Kubernetes Engine Clusters must have Stackdriver Logging enabled, which means the attribute 'logging_service' must be defined and different from 'none'

Issue #137, opened 2020-11-16
Add Stackdriver Monitoring is Disabled query for Terraform


cx-ruben-silva:
### Platform
Terraform

### Provider
GCP

### Description
Kubernetes Engine Clusters must have Stackdriver Monitoring enabled, which means the attribute 'monitoring_service' must be defined and different than 'none'

PR #6204, opened 2023-03-01
refactor(tf-gcp): Legacy Stackdriver was Decomissioned 21/03/31

meldaravaniel:
Related to #140 and #137

Legacy Stackdriver was deprecated, then decomissioned March 31, 2021.  If you leave the fields undefined, GKE will default to using Stackdriver for both logging and monitoring.  This allows the fields to be undefined OR set, but must be set to the current stackdriver values.

**Proposed Changes**
- Adds `'monitoring.googleapis.com'` and `'logging.googleapis.com'` as disallowed values for `monitoring_service` and `logging_service`, respectively
- allows `monitoring_service` and `logging_service` to be undefined (default is Stackdriver)
- updates tests, docs, and url text for both queries


meldaravaniel:
Setting to draft because I thought it would be a good idea to branch of my master...so this includes the changes from my first PR...

If/when that one gets accepted, will update this accordingly.

assets/queries/terraform/gcp/stackdriver_logging_disabled/query.rego
```diff
@@ -4,30 +4,31 @@ import data.generic.terraform as tf_lib
...
 CxPolicy[result] {
 	resource := input.document[i].resource.google_container_cluster[primary]
-	not resource.logging_service
+	resource.logging_service == "none"
 
 	result := {
 		"documentId": input.document[i].id,
 		"resourceType": "google_container_cluster",
 		"resourceName": tf_lib.get_resource_name(resource, primary),
 		"searchKey": sprintf("google_container_cluster[%s]", [primary]),
```
cxMiguelSilva, line 13:
```suggestion
		"searchKey": sprintf("google_container_cluster[%s].logging_service", [primary]),
```

assets/queries/terraform/gcp/stackdriver_logging_disabled/query.rego
```diff
@@ -4,30 +4,31 @@ import data.generic.terraform as tf_lib
...
 CxPolicy[result] {
 	resource := input.document[i].resource.google_container_cluster[primary]
-	resource.logging_service == "none"
+	resource.logging_service == "logging.googleapis.com"
 
 	result := {
 		"documentId": input.document[i].id,
 		"resourceType": "google_container_cluster",
 		"resourceName": tf_lib.get_resource_name(resource, primary),
 		"searchKey": sprintf("google_container_cluster[%s]", [primary]),
```
cxMiguelSilva, line 29:
```suggestion
		"searchKey": sprintf("google_container_cluster[%s].logging_service", [primary]),
```

assets/queries/terraform/gcp/stackdriver_logging_disabled/test/positive.tf
```diff
@@ -16,6 +16,18 @@ resource "google_container_cluster" "positive2" {
   initial_node_count = 3
```
cxMiguelSilva, line 16: Since the default behavior is no longer a vulnerability please remove the `positive1` resource

assets/queries/terraform/gcp/stackdriver_logging_disabled/test/positive_expected_result.json
```diff
@@ -8,5 +8,10 @@
     "queryName": "Stackdriver Logging Disabled",
     "severity": "HIGH",
     "line": 13
```
cxMiguelSilva, line 10: Also, update the vulnerability lines accordingly to the findings in `positive.tf` file
Please also add the  `"fileName": "positive.tf"` to the `positive_expected_result.json` items

assets/queries/terraform/gcp/stackdriver_monitoring_disabled/query.rego
```diff
@@ -4,30 +4,31 @@ import data.generic.terraform as tf_lib
...
 CxPolicy[result] {
 	resource := input.document[i].resource.google_container_cluster[primary]
-	not resource.monitoring_service
+	resource.monitoring_service == "none"
 
 	result := {
 		"documentId": input.document[i].id,
 		"resourceType": "google_container_cluster",
 		"resourceName": tf_lib.get_resource_name(resource, primary),
 		"searchKey": sprintf("google_container_cluster[%s]", [primary]),
```
cxMiguelSilva, line 13:
```suggestion
		"searchKey": sprintf("google_container_cluster[%s].monitoring_service", [primary]),
```

assets/queries/terraform/gcp/stackdriver_monitoring_disabled/query.rego
```diff
@@ -4,30 +4,31 @@ import data.generic.terraform as tf_lib
...
 CxPolicy[result] {
 	resource := input.document[i].resource.google_container_cluster[primary]
-	resource.monitoring_service == "none"
+	resource.monitoring_service == "monitoring.googleapis.com"
 
 	result := {
 		"documentId": input.document[i].id,
 		"resourceType": "google_container_cluster",
 		"resourceName": tf_lib.get_resource_name(resource, primary),
 		"searchKey": sprintf("google_container_cluster[%s]", [primary]),
```
cxMiguelSilva, line 29:
```suggestion
		"searchKey": sprintf("google_container_cluster[%s].monitoring_service", [primary]),
```

cxMiguelSilva (changes requested):
Hi @meldaravaniel, 
Thank you so much for another great contribution!
Regarding the proposed changes LGTM, I just have some requests to improve the line detection of the query and fix the unit tests.

assets/queries/terraform/gcp/stackdriver_logging_disabled/query.rego
meldaravaniel: ooh missed that.  thanks.

assets/queries/terraform/gcp/stackdriver_logging_disabled/test/positive.tf
meldaravaniel: ah! also good catch.  ta!

meldaravaniel:
Looks like the failing test is from a merge to the main kics branch, not my changes.  I'll update my branch when that's fixed.

assets/queries/terraform/gcp/stackdriver_logging_disabled/query.rego
meldaravaniel: Rejected, this causes tests to fail, I think in the test case where `logging_service` is undefined?  When I reverted the change, anyways, the tests passed...(also the == "none" test was originally using the searchKey without the .logging_service portion)

assets/queries/terraform/gcp/stackdriver_logging_disabled/query.rego
meldaravaniel: On second thought, rejected, this causes tests to fail, I think in the test case where `logging_service` is undefined?  When I reverted the change, anyways, the tests passed...(also the == "none" test was originally using the searchKey without the .logging_service portion)

meldaravaniel:
> Not sure where that would be failing...I'll poke around.

It seems to have been the addition of `.logging_service`/`.monitoring_service` in the `searchKey` in the tests.  I removed that again.

cxMiguelSilva:
Hi @meldaravaniel, 
The tests were probably failing due to the vulnerability detection line not being the same as in the `positive_expected_result.json` file. You can update the query to use still  `.logging_service/.monitoring_service` in the `searchKey` but you also need to update the expected line in the `positive_expected_result.json`  file.
Let me know what you think.

assets/queries/terraform/gcp/stackdriver_logging_disabled/query.rego
cxMiguelSilva: But to return the result in this rule `.logging_service` needs to be defined. If not defined there is no result for this rule.

assets/queries/terraform/gcp/stackdriver_logging_disabled/query.rego
cxMiguelSilva: Again to return the result in this rule `.logging_service` needs to be defined. If not defined there is no result for this rule. This would only be a problem with the use of the `not` operator.

meldaravaniel:
> You can update the query to use still `.logging_service/.monitoring_service` in the `searchKey` but you also need to update the expected line in the `positive_expected_result.json` file.

Mmm...no, I don't think it's that.  The expected lines in `positive_expected_result.json` refer to the lines in the `positive.tf` file, not the query.  Changing the query shouldn't affect that.  The expected lines match, otherwise the tests would be failing right now (I haven't changed the expected line values other than to match them to the new positive tests

I'm not running the tests locally, so I might try that, but is there a way to run them such that the error is better than just the exit code?  Or maybe I can use a debugger...

cxMiguelSilva:
> > You can update the query to use still `.logging_service/.monitoring_service` in the `searchKey` but you also need to update the expected line in the `positive_expected_result.json` file.
> 
> Mmm...no, I don't think it's that. The expected lines in `positive_expected_result.json` refer to the lines in the `positive.tf` file, not the query. Changing the query shouldn't affect that. The expected lines match, otherwise the tests would be failing right now (I haven't changed the expected line values other than to match them to the new positive tests
> 
> I'm not running the tests locally, so I might try that, but is there a way to run them such that the error is better than just the exit code? Or maybe I can use a debugger...

Yes you can run the unit tests locally using the following command on the repo root
`go test ./test`

cxMiguelSilva:
Hi @meldaravaniel hope you are doing great!
Regarding the changes, I suggested triggering problems on the tests
I have updated the values on my side and ran the Unit tests and E2E tests and seem to be ok, with no error whatsoever.

The changes I made:
assets/queries/terraform/gcp/stackdriver_logging_disabled/query.rego
- added `.logging_service` to the search key in both policies
 `"searchKey": sprintf("google_container_cluster[%s].logging_service", [primary]),`

assets/queries/terraform/gcp/stackdriver_logging_disabled/test/positive_expected_result.json
- updated the vuln lines to `6` and `18`
- added `"fileName": "positive.tf"` in both results

assets/queries/terraform/gcp/stackdriver_monitoring_disabled/query.rego
- added `.monitoring_service` to the search key in both policies
 `"searchKey": sprintf("google_container_cluster[%s].monitoring_service", [primary]),`

assets/queries/terraform/gcp/stackdriver_monitoring_disabled/test/positive_expected_result.json
- updated the vuln lines to `6` and `18`
- added `"fileName": "positive.tf"` in both results

Let me know if this helps!

meldaravaniel:
@cxMiguelSilva omg my brain just turned back on and I understand the expected/actual line numbers now.  idk why that was so hard for me to get.  we'll pretend it was sleep deprivation. :P  I was thinking they just meant "the line at which each resource starts".  But it's linked to the search key, so in my other commits, that *happened* to be the same as "line where resource starts"...because the search key was just the parent level resource.  Had to read the test code a bit before that hit me. *sigh

Got Go up and running on my computer, much faster to test locally (obviously heh) and yes, that all works and makes sense now.

test/fixtures/test_terraform_variables/test.tf
```diff
@@ -15,7 +15,7 @@ resource "test" "test1" {
   test_map        = var.map2
   test_bool       = var.test1
   test_list       = var.test2
-  test_neted_map  = var.map2[var.map1["map1key1"]]]
+  test_neted_map  = var.map2[var.map1["map1key1"]]
```
meldaravaniel, line 18: this and t`test_without_variables_block.tf` were making vscode show an error...so I went ahead and fixed them to not have to see red. :)

cxMiguelSilva:
Hi @meldaravaniel,
Don't worry everyone has those moments 😅. I'm glad I was able to help.

test/fixtures/test_terraform_variables/test.tf
cxMiguelSilva: I kindly ask you to revert the fix in that `test.tf` file. It is used for testing purposes 😅
The fix on that file is causing this [test](https://github.com/Checkmarx/kics/blob/master/pkg/parser/terraform/terraform_test.go#L238) to fail.

meldaravaniel: eheh this is what I get for not running the *full test suite before committing.

cxMiguelSilva (approved):

Issue #91, opened 2020-11-16
Add Client Certificate is Disabled query for Terraform


cx-ruben-silva:
### Platform
Terraform

### Provider
GCP

### Description
Kubernetes Clusters must be created with Client Certificate enabled, which means 'master_auth' must have 'client_certificate_config' with the attribute 'issue_client_certificate' equal to true

PR #6205, opened 2023-03-01
feat(tf-gcp-auth): check for legacy authentication enabled

meldaravaniel:
related to #91

Google now recommends against using Client Certificates as authentication and they automatically configure OAuth on all clusters.  For new clusters, master_auth can be left undefined.  For existing clusters, it can be undefined or set the issue_cert to false (probably just prefer `master_auth` undefined, but that might be too opinionated?).  Unless I'm mistaken, all clusters have OAuth enabled, even those that had client certs issued.  

You can't remove a client cert if it's already there, so they recommend making sure nothing in the cluster is allowed to use it anymore and the cluster has RBAC enabled.  More detail [here](https://cloud.google.com/kubernetes-engine/docs/how-to/api-server-authentication#disabling_authentication_with_a_client_certificate).  

More info [in the GCloud docs](https://cloud.google.com/kubernetes-engine/docs/how-to/api-server-authentication). 

**Proposed Changes**
- repurpose the "client certificate disabled" query into "legacy auth enabled"
- clusters should either not define the `master_auth` block, or they should set `master_auth.client_certificate_config.issue_client_certificate` to false
- updates tests, docs, and text in the metadata file


assets/queries/terraform/gcp/legacy_auth_enabled/test/negative.tf
```diff
@@ -5,9 +5,6 @@ resource "google_container_cluster" "negative1" {
   initial_node_count = 3
 
   master_auth {
-    username = ""
```
meldaravaniel, line 8: also removed these because they're not allowed in current versions of GKE

meldaravaniel:
Setting to draft because I thought it would be a good idea to branch of my master...so this includes the changes from my first PR...

If/when that one gets accepted, will update this accordingly.

assets/queries/terraform/gcp/legacy_auth_enabled/metadata.json
```diff
@@ -1,9 +1,9 @@
 {
   "id": "73fb21a1-b19a-45b1-b648-b47b1678681e",
-  "queryName": "Client Certificate Disabled",
+  "queryName": "Legacy Auth Enabled",
```
cxMiguelSilva, line 3:
```suggestion
  "queryName": "Legacy Client Certificate Auth Enabled",
```

cxMiguelSilva (changes requested):
Hi @meldaravaniel, 
Thank you for another outstanding contribution!
Regarding the proposed changes, I would like to ask you to update the Security Query name to go accordingly to our Application Security team review of this PR.

cxMiguelSilva (approved):

PR #6240, opened 2023-03-21
feat(engine): ignore terraform cache folders

### LINKED ISSUE #6217: Ignore .terra* cache directories by default globally
When running KICS the Terraform module cache directories are scanned by default. Most consumers of modules have little to no ability to change a 3rd part module. 

By default KICS should ignore `**/.terra*` when executing a `scan`. This would ignore both .terraform and .terragrunt cache directories that are present in module Terraform deployment modules.

Currently using `--exclude-paths ".terra*, ./.terra*, ./**/.terra*, .terra*/*, ./.terra*/*, ./**/.terra*/*", ...". This is functional but contrived.

**Additional context**

The full command:

```
kics scan \
        --cloud-provider "aws" \
        --exclude-paths ".terra*, **/.terra*" \
        --no-color \
        --path "." \
        --queries-path "$PATH_TO/assets/queries" \
        --type "Terraform" \
        --verbose
```

dim-ops:
Closes #6217
Proposed Changes

- To improve scan Kics can ignore terraform cache folders;


pkg/scan/scan.go
kaplanlior, line 1: @dim-ops Can you add the missing space to satisfy the linter ?

dim-ops: I can but the PR is not over because actually this code works well when the path is ".". I have to spend time to manage all paths

dim-ops:
It works for me, it ignores all .terra folders in current path or in subdirectories

cxMiguelSilva:
Hi @dim-ops, 
Thank you for another great contribution! 🚀
From an initial overview, the changes make sense to me. I would only suggest adding some tests to the [`filesystem_test.go`](https://github.com/Checkmarx/kics/blob/fba7a8a46332902c6123f26a689f04c6781da533/pkg/engine/provider/filesystem_test.go#L269) just to make sure the behavior from now onwards works as respected. 
Let us know in case you need any help!

dim-ops:
I begin to look, I have 2 questions @cxMiguelSilva :
- Should I create new folders tests in test/fixture with terraform cache folder `.terraform` ?
- Should I create new test function or I should add a test in `TestFileSystemSourceProvider_checkConditions` ?

pkg/engine/provider/filesystem_test.go
```diff
@@ -310,7 +315,7 @@ func TestFileSystemSourceProvider_checkConditions(t *testing.T) {
 				path: filepath.FromSlash("assets/queries"),
 			},
 			want: want{
-				got: true,
```
dim-ops, line 313: I changed this test because I didn't understand why "true" was expected. During the test I passed in this line:
`_, err := os.Stat(filepath.Join(path, "Chart.yaml"))`

Which returned `true, nil `because I had no `Chart.yaml` file in my path, so I understood if there is no `Chart.yaml` file it should be ignored and I don't understand why.

Can you explain me? @cxMiguelSilva

cxMiguelSilva: Hi @dim-ops, 
That check is in place to ignore all folders of Helm Charts to later make a scan to the Helm templates

dim-ops: Yes I understand the interest of `_, err := os.Stat(filepath.Join(path, "Chart.yaml"))`but I don't understand why `true` is returned when Chart.yaml doesn't exist. If I understand well true is when folder must be skipped, if Chart.yaml is not present folder should be parsed.

It's why I modify this check 
```
		_, err := os.Stat(filepath.Join(path, "Chart.yaml"))
		if errors.Is(err, os.ErrNotExist) {
			return false, nil
		} else if err != nil || resolved {
			log.Error().Msgf("failed to check helm: %s", err)
			return true, nil
		}
```

dim-ops:
Can I have more information @gabriel-cx ?

dim-ops:
Up @gabriel-cx please 🙏

JoaoCxMartins (commented):
Hi @dim-ops , how are you? 
So I was looking into your code and made some changes,
first I changed the detection of .terra* to use regex, to cover all the cases present on the issue;
second I removed the logic that you put for the chart.yaml, because I ran regressions, and the files that should scanned with the helm resolver were being treated by the normal yaml resolver and were giving exceptions with this we were losing results.
Let me know what you think.
Br,
Joao Martins

dim-ops:
Thanks @JoaoCxMartins it seems better 👍

JoaoCxMartins:
> Thanks @JoaoCxMartins it seems better 👍

TY for your help improving Kics :)

JulioSCX (dismissed):

cxMiguelSilva (approved):

PR #6241, opened 2023-03-21
Policy effect should be case insensitive

Tohar-orca:
**Proposed Changes**
- Per [Terraform's documentation](https://developer.hashicorp.com/terraform/tutorials/aws/aws-iam-policy#:~:text=Copy%20the-,aws_iam_policy_document,-configuration%20below%20into), iam policy statements' `Effect` can be declared as either `Effect` or `effect`
- The proposed change ensures both cases are validated when checking whether the statements is an `Allow` statement
- This solves false-positive cases where statements with `effect: Deny` are evaluated as being `Allow`

### DIFF
--- a/assets/libraries/common.rego
+++ b/assets/libraries/common.rego
@@ -412,7 +412,7 @@ is_allow_effect(statement) {
 } else {
 	statement.Effect == "Allow"
 } else {
-    satement.effect == "Allow"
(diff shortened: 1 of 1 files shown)


cxMiguelSilva (approved):
Hi @Tohar-orca,
Thank you for another amazing contribution!

Issue #6243, opened 2023-03-22
Add GKE Cluster Shielded Nodes is Disabled query for Terraform


bbergstrom:
### Platform
Terraform

### Provider
GCP

### Description
GKE Cluster node instances must be launched with Shielded VM enabled, which means the attribute 'enable_shielded_nodes' must be defined and be set `true`.  While it does default to `true`, setting it explicitly mitigates potential drift.  Similar example can be found in https://github.com/Checkmarx/kics/issues/197 for non-GKE VM instances.

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#enable_shielded_nodes

PR #6248, opened 2023-03-23
feat(terraform/gcp): Add GKE Shielded Nodes is Disabled query for Terraform.

bbergstrom:
Add Terraform query for "GKE Shielded Nodes is Disabled".  

Closes https://github.com/Checkmarx/kics/issues/6243


cxMiguelSilva:
Hi @bbergstrom, 
Thank you for this amazing contribution!
Regarding the proposed changes. The query logic LGTM. 
I will just ask our Application Security to validate the metadata and should be good to go!

cxMiguelSilva (changes requested):
Upon further inspection, I noticed that the terraform docs suggest that by default the value is set to `true`.  I believe the first policy that detects the default value should be removed.

Let me know what you think.

bbergstrom:
@cxMiguelSilva The first query detects for the implicit default, which should be a positive because the value could drift if it is changed outside of Terraform.  The best practice for these is to explicitly set the default value in Terraform to ensure it catches and corrects drift to the secure value.

cxMiguelSilva (changes requested):
Hi @bbergstrom,
In KICS we do not create policies to cover default values that are not deemed as vulnerable or misconfiguration. We have come across bug reports and issues handling from the community to improve our Security Queries in order to remove those False Positive results from incorrect checks on default values. 
I kindly ask you to remove the policy that checks the default value since the Terraform documentation is indicative that it is set to the correct value. If at any time the guidelines change feel free to update the Security Query and add that policy.

kaplanlior:
Hi @bbergstrom ,

Would you have time in the near future to do the changes Miguel asked for?

bbergstrom:
@kaplanlior Pushed suggested changes.  Awaiting approval for test suite.

bbergstrom:
@gabriel-cx Tests all passed. 🎉   Would you finish the review or should I wait for @cxMiguelSilva ?

PR #6252, opened 2023-03-24
fix(query): add exception for Github id-token Default Values rule

henrique-alvelos:
**Proposed Changes**
- Added Avoiding Github id-token Default Values rule
- Added negative test


cxMiguelSilva (approved):
LGTM

PR #6255, opened 2023-03-27
fix(query): Effect 'Allow' conditions added

gabriel-cx:
**Proposed Changes**
- Effect 'Allow' conditions added
- Tests added and updated


cxMiguelSilva (approved):
LGTM

PR #6261, opened 2023-03-29
fix(query): added issue.solution var to differentiate issue types

cx-henriqueAlvelos:
**Proposed Changes**
- Add var issue.solution to show different outputs (Missing Output or Incorrect Value - True)


cxMiguelSilva (changes requested):
The Proposed changes LGTM
I will just ask you to change the keyExpetedValue to align with the other queries

cxMiguelSilva (approved):

PR #6296, opened 2023-04-13
fix(query): add rule for generic access_key 

cx-henriqueAlvelos:
**Proposed Changes**
- Added regex to catch access_key values
- Added positive (with expected result)
- Added negative test


gabriel-cx (approved):

cxMiguelSilva:
changes moved to #6360

PR #6297, opened 2023-04-13
fix(analyser) Fix to ignore files (pnpm-lock.yaml)

cx-henriqueAlvelos:
**Proposed Changes**
- Global var ignoreFiles
- Path comparison with Exception Paths


cx-henriqueAlvelos:
Proposed Changes

- Global var defaultConfigFiles
- Path comparison with defaultConfigFiles


gabriel-cx (approved):

PR #6357, opened 2023-05-05
feat(query): EFS Volume With Disabled Transit Encryption

cx-henriqueAlvelos:
**Proposed Changes**
- Added query for Terraform and Cloudplatform each
- Added 2 positive tests and 1 negative for  Terraform and Cloudplatform each
- Added positive_expected_result for Terraform and Cloudplatform each


pereiramarco011 (approved):
LGTM! 😄

sherwind:
Folks, it seems that`transit_encryption` is a not valid `aws_efs_file_system` argument as of AWS Terraform provider v5.5.0

See:
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/efs_file_system
https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/efs/file_system.go

PR #6372, opened 2023-05-11
Bug(query) - Fix not correct line for query RDS Storage Encryption Disabled

pereiramarco011:
Closes #

**Proposed Changes**
- Change line in which query detects in case key is undefined


assets/queries/cloudFormation/aws/rds_storage_encryption_disabled/query.rego
```diff
@@ -32,7 +32,7 @@ CxPolicy[result] {
 		"documentId": input.document[i].id,
 		"resourceType": resource.Type,
 		"resourceName": cf_lib.get_resource_name(resource, name),
-		"searchKey": sprintf("%s%s.Properties.StorageEncrypted", [cf_lib.getPath(path), name]),
+		"searchKey": sprintf("%s%s", [cf_lib.getPath(path), name]),
```
cxMiguelSilva, line 35:
```suggestion
		"searchKey": sprintf("%s%s.Properties", [cf_lib.getPath(path), name]),
```

cxMiguelSilva (changes requested):
The approach taken to solve this problem seems good to me.
I would remove the `StorageEncrypted` and keep the Properties since it is where the checked attribute is missing.
Update the `positive_expected_result.json` accordingly.

assets/queries/cloudFormation/aws/rds_storage_encryption_disabled/query.rego
pereiramarco011: Done [here](https://github.com/Checkmarx/kics/pull/6372/commits/92992b0c3c7a1eaf629569bd677188085b31d212)

PR #6393, opened 2023-05-18
feat(query): Elasticsearch with HTTPS disabled for Ansible

cx-henriqueAlvelos:
**Proposed Changes**
- Added query and metadata for Ansible
- Added tests
- Added positive_expected_result


gabriel-cx (changes requested):
We need to create a CxPolicy in the rego file to trigger a vulnerability result when the **EnforceHTTPS** attribute is not set in the configuration.

Adding a positive test to cover the previous situation is also needed.

cx-henriqueAlvelos:
> We need to create a CxPolicy in the rego file to trigger a vulnerability result when the **EnforceHTTPS** attribute is not set in the configuration.
> 
> Adding a positive test to cover the previous situation is also needed.

Hi. Done.

gabriel-cx (approved):

PR #6394, opened 2023-05-18
feat(query): Elasticsearch with HTTPS disabled for Terraform

cx-henriqueAlvelos:
**Proposed Changes**
- Added query and metadata for Terraform
- Added tests
- Added positive_expected_result


gabriel-cx (changes requested):
We need to create a CxPolicy in the rego file to trigger a vulnerability result when the **EnforceHTTPS** attribute is not set in the configuration.

Adding a positive test to cover the previous situation is also needed.

cx-henriqueAlvelos:
> We need to create a CxPolicy in the rego file to trigger a vulnerability result when the **EnforceHTTPS** attribute is not set in the configuration.
> 
> Adding a positive test to cover the previous situation is also needed.

Hi. In [Terraform description about EnforceHTTPS attribute](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/elasticsearch_domain#enforce_https) it says that "Defaults to true", so there is no need to create a CxPolicy

gabriel-cx (approved):

PR #6428, opened 2023-06-05
Dockerfile: upgrade alphine to 3.18.0

kaplanlior:
Remove hard coded version for git and gcompat


Dockerfile
```diff
@@ -48,16 +48,16 @@ RUN wget https://releases.hashicorp.com/terraform/1.3.9/terraform_1.3.9_linux_am
...
+    git
 
 
 # Install Terraformer
 RUN wget https://github.com/GoogleCloudPlatform/terraformer/releases/download/0.8.22/terraformer-all-linux-amd64 \
     && chmod +x terraformer-all-linux-amd64 \
     && mv terraformer-all-linux-amd64 /usr/bin/terraformer \
-    && apk add gcompat=1.1.0-r0 --no-cache
+    && apk add --no-cache \
+    gcompat
```
gabriel-cx, line 60: is recommended to specify the package version.
kindly check the KICS scan warning about this situation and update it accordingly.

gabriel-cx (changes requested):

Dockerfile
kaplanlior: I think in this case it's counter productive as the version is dictated by Alpine repo and there's only one copy/version of each package. I would prefer to control the docker version than the apk package version as well. Also, this allows us to merge dependabot suggestions without requiring a manual PR to update the packages version.

The query is correct in cases when you might have multiple options (in the same apk repo or in multiple repos).

gabriel-cx: We need to consult the AppSec team regarding this situation.
And then we can decide the best approach, having in mind their suggestion.

sanmai-NL: @kaplanlior Can you elaborate about Dependabot? Couldn't Dependabot update the specific APK package versions as well? Indeed, how to make this scannable etc. is a challenge (generally, I use Renovate, which is very smart about this).

Version pinning is a method to solve the problem of non-reproducible builds, however, the challenge of maintainability is also a point. Reproducible builds are useful if they can indeed be reproduced. As @kaplanlior, when the upstream OS package provider decides to change the versions offered for download, you may not be able to reproduce the image even with will version specifications. For that, you'll need to clone or mirror the OS packages repository using, e.g. Artifactory of copies of the apk files in the repo. Either way, full self-contained reproducibility may not be practical to attain ...

kaplanlior: OK.

labzz2000 (approved):

gabriel-cx (approved):

PR #6477, opened 2023-06-27
fix(query): transit_encryption attribute changed

### LINKED ISSUE #6524: Kics complains about a configuration that is not valid.
https://docs.kics.io/latest/queries/terraform-queries/aws/4c3267c9-b2ac-40bf-93f6-b610fb8c7b9f/

refers to a configuration that is not valid for aws_efs_file_system (https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/efs_file_system) terraform resource. There's no such attribute called transit_encryption on that resource. It is however an attribute on aws_ecs_task_definition (https://registry.terraform.io/providers/hashicorp/aws/3.9.0/docs/resources/ecs_task_definition) which is a completely different service.

The check is incorrectly flagging as a vulnerability. The documentation is incorrect as well.

cx-henriqueAlvelos:
Closes #6524 
**Proposed Changes**
- transit_encryption attribute changed


cxMiguelSilva (dismissed):
Hi @cx-henriqueAlvelos, regarding the query in overall LGTM, I just request some changes to ensure that the query is aligned with all the terraform queries.
Please ensure that all the recently introduced queries are using the platform documentation.
Also when trying to detect if a attribute/obj make use of the `valid_key()`function from the `common_lib`

cxMiguelSilva (approved):

PR #6497, opened 2023-07-05
Fix FN in string due to bad allowRule matches

pereiramarco011:
**Proposed Changes**
- Modify the way allowRules are matched with regex results


cx-henriqueAlvelos (approved):
LGTM 🚀

PR #6506, opened 2023-07-07
fix(regex): added regex into allow rule list

### LINKED ISSUE #6504: Passwords And Secrets - Generic Password fale positive on null values
### Expected Behavior

`Passwords And Secrets - Generic Password` (`487f4be7-3fd9-4506-a07a-eae252180c08`) finds valid plaintext passwords in infrastructure code

### Actual Behavior

`Passwords And Secrets - Generic Password` is being triggered incorrectly by a `null` value, which can be valid in certain instances such as creating an RDS Replica (which inherits the password from it's parent). Specifically this happens when the null value is in a tenary:

```
Passwords And Secrets - Generic Password, Severity: HIGH, Results: 1
Description: Query to find passwords and secrets in infrastructure code.
Platform: Common
Learn more about this vulnerability: https://docs.kics.io/latest/queries/common-queries/common/487f4be7-3fd9-4506-a07a-eae252180c08

	[1]: ../../path/terraform/modules/gitlab_ref_arch_aws/rds.tf:67

		010:   username = local.rds_postgres_is_primary ? var.rds_postgres_username : null
		011:   password = local.rds_postgres_is_primary ? var.rds_postgres_password : <SECRET-MASKED-ON-PURPOSE>
		012:
```

Where `<SECRET-MASKED-ON-PURPOSE>` is `null`, which is equivalent to "don't set".

This worked fine recently but started to fail with `v.1.7.2`.

Attempt to configure an RDS instance with password set based on condition with `null` being the second option.

freitasmillena:
Closes #6504 

**Proposed Changes**
- Added regex into allow rule list

### DIFF
--- a/assets/queries/common/passwords_and_secrets/regex_rules.json
+++ b/assets/queries/common/passwords_and_secrets/regex_rules.json
@@ -6,6 +6,10 @@
       "regex": "(?i)['\"]?password['\"]?\\s*[:=]\\s*['\"]?([A-Za-z0-9/~^_!@&%()=?*+-. ]{4,})['\"]?",
       "allowRules": [
         {
+          "description": "Avoiding TF resource access",
+          "regex": "(?i)['\"]?password['\"]?\\s*=\\s*([a-zA-z_]+(.))?[a-zA-z_]+(.)[a-zA-z_]+(.)[a-zA-z_]+\\s*:\\s*null"
+        },
+        {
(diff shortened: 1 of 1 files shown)


pereiramarco011 (changes requested):
rename negativo55.tf to negative55.tf

other than that LGTM 😄

pereiramarco011 (approved):
LGTM!! 😄

PR #6522, opened 2023-07-18
fix(secrets): improve oAuthSecret secret to detect more valid characters

cxMiguelSilva:
**Proposed Changes**
- improve secret to detect more valid chars
- add test case to validate


pereiramarco011 (dismissed):
LGTM 😄

pereiramarco011 (approved):

PR #6527, opened 2023-07-19
feat(panic): add panic handler to possible panic places

pereiramarco011:
**Proposed Changes**
- Increase KICS scan resiliency
- https://github.com/Checkmarx/kics/discussions/6521


kaplanlior:
@liorj-orca, I would appreciate your feedback on this PR.
CC @lior-orca

liorj-orca:
@pereiramarco011 overall looks good. As @lior-orca stated, I think we can push as is and add more cases later on along the way.
One more case that you can think wether or not we want to avoid the error:
- Getsources function (https://github.com/Checkmarx/kics/blob/fdea929f6895fa3c6e6fb2df5b197591308af9e7/pkg/engine/provider/extract.go#L132) could fail for several reasons, among other thing a bug of go-getter.  what about skipping any path that we fail to resolve?

lior-orca:
@pereiramarco011 any updates? what is preventing us from pushing the change?

gabriel-cx (approved):

gabriel-cx (dismissed):

gabriel-cx (approved):

PR #6534, opened 2023-07-21
fix(query):  Non detection of . in vars for Terraform 

pereiramarco011:
**Proposed Changes**
- Add correct variable detection in allowRules
- Add unit test


cx-henriqueAlvelos (approved):
LGTM 🚀

PR #6539, opened 2023-07-24
fix(analyzer): Remove Optional DockerCompose Regex

### LINKED ISSUE #6509: Compose files without a `version` property do not trigger loading of `DockerCompose` queries
KICS does not load [`DockerCompose` queries](https://docs.kics.io/latest/queries/dockercompose-queries/) when there is not at least one Compose file anywhere in the scanned project that includes a `version` property. This property is [optional per the Compose specification](https://github.com/compose-spec/compose-spec/blob/66f4f2214fbe9038c8251e8322bc3270f8659df1/04-version-and-name.md), and even those files missing it should trigger loading of said queries.

When specifically only scanning a single Compose file missing the property, the only loaded query type is `ansible`. Manually specifying `--type DockerCompose` on the CLI does not appear to resolve this; it merely leads to no queries being loaded at all.

Adding at least one Compose file with the `version` property somewhere in the project results in the queries being loaded and **all** Compose files being scanned properly; even those "missing" a `version`.

### Expected Behavior

`DockerCompose` queries are loaded when any Compose file is encountered, even if it's missing the optional `version` property.

### Actual Behavior

Compose files without a `version` property do not trigger loading of `DockerCompose` queries.

1.  Create a project directory only containing the following minimal `docker-compose.yml`:

    ```yaml
    services:
      postgres:
        image: 'postgres:15.3-alpine'
        environment:
          POSTGRES_DB: 'test'
          POSTGRES_USER: 'test'
          POSTGRES_PASSWORD: 'hardcoded'
    ```

2.  Run a scan using the latest KICS version:

    ```shell
    docker run -v ./:/project -t checkmarx/kics:latest scan -v -p "/project"
    ```

    Only the hardcoded credential is detected, as that is covered by the generic _Password and Secrets_ query. However, no Docker Compose–specific issues ("Healthcheck Not Set", "Security Opt Not Set", etc.) are highlighted.

3.  Next, either add a `version` property to the file or – alternatively – create a second Compose file containing one. The file(s) should now be analyzed as expected.

### Specifications

-   Version: v1.7.3
-   Platform: Linux (Docker)
-   Subsystem: DockerCompose

cx-henriqueAlvelos:
Closes #6509

**Proposed Changes**
- Removed Version Regex


pereiramarco011 (approved):
LGTM 😄

pereiramarco011 (approved):

PR #6628, opened 2023-08-09
feat(query): ansible config privilege_escalation_using_become_plugin

cx-henriqueAlvelos:
**Proposed Changes**
- Query
- Tests
- Positive Expected Results


gabriel-cx (changes requested):
We need to add a `negative` test where we set the `become_user=root` and `become=True`

gabriel-cx (approved):

PR #6695, opened 2023-09-05
feat(query): ansible playbooks privilege escalation using become plugin

cx-henriqueAlvelos:
**Proposed Changes**
- Query
- Tests
- Positive Expected Result


gabriel-cx (changes requested):
we need a `positive `test where we set the `become_user `and `become `is set as `false`

and

we need a `negative `test where we set `become_user `and `become `is set as `true`

gabriel-cx (dismissed):

gabriel-cx (approved):

PR #6717, opened 2023-09-09
fix(query): openapi Maximum Length Undefined

cx-henriqueAlvelos:
**Proposed Changes**
- Added pattern condition


ArturRibeiro-CX (dismissed):
LGTM. Now this query doesn't flag for situations where a schema for type string contains a pattern that restricts the size of that string, as it is the intended behaviour for this query.

PR #6769, opened 2023-10-13
feat(engine): experimental queries as feature flag

JoaoCxMartins:
with this commit, we are changing our system of experimental queries from a file that we should read for a feature flag

Closes #

**Proposed Changes**
- change the feature of the experimental queries to work with a feature flag


pkg/engine/source/filesystem.go
```diff
@@ -449,13 +395,17 @@ func ReadQuery(queryDir string) (model.QueryMetadata, error) {
 		aggregation = int(agg.(float64))
 	}
 
+	readExperimental, _ := metadata["experimental"].(string)
```
cxMiguelSilva, line 398: Since you are using an auxiliary function to get the experimental value, maybe you could add  the type casting there also.

cxMiguelSilva (changes requested):
Hi @JoaoCxMartins, great work on the engine changes. I just have a code suggestion placed in the comment. Also please make sure that the dockerfiles are properly synched with the need for the file `assets/utils/experimental-queries.json`.

baptfh (approved):

gabriel-cx:
@JoaoCxMartins kindly remove the `COPY --from=build_env /app/assets/utils/* /app/bin/assets/utils/` from the following files:

- https://github.com/Checkmarx/kics/blob/master/docker/Dockerfile.apispec.debian#L73C1-L73C65
- https://github.com/Checkmarx/kics/blob/master/docker/Dockerfile.debian#L72
- https://github.com/Checkmarx/kics/blob/master/docker/Dockerfile.apispec#L59

JoaoCxMartins:
Thanks, @gabriel-cx and @cxMiguelSilva for your review.
Can you review it again? :)

pkg/engine/source/filesystem.go
```diff
@@ -365,6 +306,10 @@ func (s *FilesystemSource) iterateQueryDirs(queryDirs []string, queryParameters
 			continue
 		}
 
+		if !queryParameters.ExperimentalQueries && query.Experimental {
```
gabriel-cx, line 309:
```suggestion
		if query.Experimental && !queryParameters.ExperimentalQueries {
```

Since the majority of the queries do not have the `experimental` field, means the `query.Experimental` value is `false`, with the code suggestion, **we just check one of the if conditions** (because the first condition is `false `so the program do not need to check the second condition since it is a `&&` operator) for the majority of cases.

gabriel-cx (changes requested):

pkg/engine/source/filesystem.go
JoaoCxMartins: Thanks @gabriel-cx

gabriel-cx (approved):

PR #6798, opened 2023-11-13
feat(engine): improve experimental signal on the results and cli

JoaoCxMartins:
Closes #

**Proposed Changes**
- improve the signal of experimental on cli interface
- improve the signal of experimental on file exported
- tests


pkg/engine/vulnerability_utils.go
```diff
@@ -129,6 +129,30 @@ func getStringFromMap(vulnParam, defaultParam, overrideKey string, vObj map[stri
 	}
 	return *ts
 }
+func getBoolFromMap(
```
gabriel-cx, line 132: this method is triggering unnecessary errors like: `Saving result. failed to detect experimental` when trying to scan a project using **"no experimental" queries**.

gabriel-cx (changes requested):

pkg/engine/vulnerability_utils.go
JoaoCxMartins: done

gabriel-cx (approved):

PR #6815, opened 2023-11-23
fix(converter): improve check dynamic known types

liorj-orca:
```
This PR improves the case where the resource/block needs some recursive handling.
without this case(an example was added to the UT), we will get an error such as:
"Error: failed to read sources: failed to save file content: json: error calling MarshalJSON for type model.Document: value is not known
"
```
**Proposed Changes**
-
-
-


gabriel-cx:
Hi @liorj-orca ,

Thank you for this PR!
It's possible for you to provide us the example file? So we can check from our side. Thanks in advance.

JoaoCxMartins (approved):

PR #6877, opened 2024-02-09
fix(flag): validating if output path is valid

### LINKED ISSUE #5513: Kics does not validate if output folder is valid
### Expected Behavior

When the user sets an output folder (for results and/or logs), kics may validate if the return of file creation in this path was successful.

### Actual Behavior

Even if the user defines an output folder that is not valid (with invalid characters), at the end of the scan, kics will display that files were saved in the output folder.

#### Example of kics scan using an invalid output folder and all report formats:
image

Open a terminal and run: `docker run -v "c/:/path" checkmarx/kics:latest scan --no-progress -t dockerfile --path "git::https://github.com/cxlucas/kics-scan-builder" --output-path "/path/*<?>*/:??/folder"`

Kics will display that the results file was saved to the specified path, but this folder contains invalid characters, so it's not possible to create a file in this specific path.
*Note:*: Kics is converting the invalid characters to some valid characters, by this way, kics is creating the result file(s) but not exactly in the same path that kics provides on the message "Results saved to file...".

image
### Specifications

-   Version: N/A
-   Platform: N/A
-   Subsystem: N/A

JulioSCX:
Closes #5513

**Proposed Changes**
- validating if output path is valid


ArturRibeiro-CX (commented):
Seems clean to me. Great job @JulioSCX! 🎉

ArturRibeiro-CX (approved):

PR #6896, opened 2024-02-21
fix(analyzer): gitignore only being used to exclude files from the project itself

JulioSCX:
**Proposed Changes**
- gitignore file is now only used to exclude files from the project itself, not its full path


pkg/analyzer/analyzer.go
```diff
@@ -694,18 +694,19 @@ func isConfigFile(path string, exc []string) bool {
...
+	gitIgnoreRes *ignore.GitIgnore, pathRes string) {
 	gitIgnorePath := filepath.ToSlash(filepath.Join(path, gitIgnore))
 	_, err := os.Stat(gitIgnorePath)
 
 	if !excludeGitIgnoreFile && err == nil && gitIgnore != "" {
 		gitIgnore, _ := ignore.CompileIgnoreFile(gitIgnorePath)
 		if gitIgnore != nil {
 			log.Info().Msgf(".gitignore file was found in '%s' and it will be used to automatically exclude paths", path)
-			return true, gitIgnore
+			return true, gitIgnore, path
```
pereiramarco011, line 706: why is the variable returned the same as the first parameter? is this the correct variable to return?

pereiramarco011 (changes requested):
is it possible to create an e2e test where the user scenario is replicated to be sure it keeps working?

small comment other than that LGTM 😄

pkg/analyzer/analyzer.go
JulioSCX: I added an E2E test and fixed the variable, thanks for the suggestions!

pereiramarco011: https://github.com/Checkmarx/kics/pull/6896/commits/8ac4222fe8a2595377523d2a363894549dccce6a

pereiramarco011 (approved):

PR #6913, opened 2024-02-28
fix(utils): exclude empty files

pereiramarco011:
**Proposed Changes**
- exclude empty files from being possibleDocker
- add unit test


JoaoCxMartins (dismissed):

cx-andrep:
The master version already has this fixed, with the same results as this branch.

PR #6916, opened 2024-02-29
fix(query): vulnerability_alert_score_epss_features

Tohar-orca:
**Proposed Changes**

- Fix a false negative in `terraform/aws/vulnerability_alert_score_epss_features` in cases where a terraform plan json is scanned instead of a .tf file

### DIFF
--- a/assets/queries/terraform/aws/sensitive_port_is_exposed_to_entire_network/query.rego
+++ b/assets/queries/terraform/aws/sensitive_port_is_exposed_to_entire_network/query.rego
@@ -27,6 +27,31 @@ CxPolicy[result] {
 	}
 }
 
+CxPolicy[result] {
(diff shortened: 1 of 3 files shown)


gabriel-cx:
hi @Tohar-orca ,

thank you for your contribution!
kindly be aware of this failing unit test:
image

Tohar-orca:
Thanks @gabriel-cx I've fixed the unit tests and added a `negative.json` file
It seems the tests are failing on something unrelated though, could you please re-run them?

gabriel-cx:
No problem @Tohar-orca !
I will re-run the tests, the e2e tests sometimes fail but the reason is not related to the PR's changes (re-run make them pass), and is the case here! We have a fix for it on the way!

gabriel-cx (approved):

PR #7051, opened 2024-05-13
fix(query): tf mfa delete doing checks out of its scope

JulioSCX:
**Proposed Changes**
- S3 Bucket Without Enabled MFA Delete query unnecessarily checking for versioning


cx-ruiaraujo (approved):
LGTM!

PR #7057, opened 2024-05-16
fix(query): implicit flow in oauth2 queries duplicated

ArturRibeiro-CX:
**Reason**
- Duplicated queries with the same logic, one needs to be deleted;

**Proposed Changes**
- delete newest query Implicit Flow query in OAuth2 (v3);
- add new unit tests to OAuth2 With Implicit Flow query;
- improve phrasing of actual and expected results;


cx-ruiaraujo (approved):
Good work 🙌

PR #7060, opened 2024-05-17
fix(query): apt-get Missing '-y' To Avoid Manual Input

### LINKED ISSUE #7053: bug(dockerfile): apt-get Missing '-y' To Avoid Manual Input`
When linting a Dockerfile that has an `apt-get install` command I get the error `77783205-c4ca-4f80-bb80-c777f267c547 # APT-GET Missing '-y' To Avoid Manual Input`.  This is a false positive when using `apt-get --quiet --quiet install sl` or  `apt-get --yes install sl`

### Expected Behavior

- No error is shown when using `apt get --quiet --quiet` as it implies `-y`/`--yes`
- No error is shown when using `apt get --yes` which is the long form of `-y`

### Actual Behavior

```
[Severity: LOW, Query description: Check if apt-get calls use the flag -y to avoid user manual input.] Problem found on 'docker/local-toolbox/Dockerfile_test' file in line 2. Expected value: {{RUN apt-get --yes install sl}} sould avoid manual input. Actual value: {{RUN apt-get --yes install sl}} doesn't avoid manual input.
```

ArturRibeiro-CX:
Closes #7053

**Reason**
- Bug found where some flags that avoid manual input weren't being taken into consideration by the query;
- This threw false positive results;
- New flags to add: --yes, --assume-yes, -qq, --quiet --quiet, -q=2, -q -q; 

**Proposed Changes**
- add more flags that avoid manual input;
- add unit tests for the new flags;
- fix typos on queries expected and actual results message;
- change query's folder name, name and description to better reflect it's behaviour.  

*APPSEC-2556*


cx-ruiaraujo (approved):
Good work 🙌

PR #7085, opened 2024-05-28
update(queries): prefix "(beta)" added to queries that are still under review

gabriel-cx:
**Reason**
For queries that are still under review (`severity` and `description`), we added the prefix "_(beta)_", so the users have a clear perspective if the triggered queries are still under review or not.

**Proposed Changes**
- prefix "(beta)" added to queries that are still under review


EduardoSemanas (approved):
LGTM

PR #7136, opened 2024-06-14
feat(query): add new query for tencentcloud CVM resource

SevenEarth:
Hello, There is something new query for TencentCloud TKE resource:

cvm_instance_using_user_data

Please help review the PR
Thanks


SevenEarth:
This is my local test
image

SevenEarth:
Hello
Please help review the PR
Thanks

ArturRibeiro-CX (changes requested):
Looks good to me but could we also add positive expected results for "secretId" and "secretKey" since the regex is also looking for those values? 
Thanks for your contribution!

ArturRibeiro-CX (approved):

SevenEarth:
> Looks good to me but could we also add positive expected results for "secretId" and "secretKey" since the regex is also looking for those values? Thanks for your contribution!

Yes, you are right. I will improve it

PR #7160, opened 2024-06-26
docs(installation): add homebrew core instructions

gforien:
- kics is now available on the homebrew core tap

**Proposed Changes**
- I suggest to add this to the documentation, along with migration instructions from (checkmarks/tap to core tap).


gforien:
closed for inactivity

PR #7205, opened 2024-07-18
fix(query): fix unexpected behaviour in parameter-checking function for ARM queries

JulioSCX:
- ARM queries could give incorrect results in some cases because of a bug in an ARM-specific parameter-checking function.

**Proposed Changes**
- Fix parameter-checking function for ARM queries


ArturRibeiro-CX (approved):
LGTM! Simple change to fix an unexpected problem. 
Well done in finding that out!

PR #7207, opened 2024-07-22
fix(query): fix CWE field not appearing in KICS CLI and sarif reports

ArturRibeiro-CX:
- CWE field was not appearing on KICS CLI and sarif reports;

**Proposed Changes**
- Add CWE field to secrets inspector in order to appear on KICS CLI and Sarif reports;


cx-ruiaraujo (approved):
LGTM

PR #7260, opened 2024-09-20
fix(gitGuardian): add gitGuardian ignore file to fix FP on secret detection

ArturRibeiro-CX:
- GitGuardian is creating noise due to passwords and secrets test files;

**Proposed Changes**
- Add gitguardian ignore file to reduce development noise and fix FP results on secret detection;


cx-miguel-silva (dismissed):
LGTM

PR #6897, opened 2024-02-22
feat(nifcloud): add terraform nifcloud queries

tunakyonn:
**Proposed Changes**
- Related https://github.com/Checkmarx/kics/pull/6265 .
- Added these Terraform NIFCLOUD queries.
- And added Terraform NIFCLOUD queries docs. (Execute `make generate-queries-docs`)
  - https://github.com/Checkmarx/kics/pull/6265#issuecomment-1519512423

I submit this contribution under the Apache-2.0 license.

PR #7312, opened 2025-01-15
feat(nifcloud): Fix terraform nifcloud bug queries

tunakyonn:
**Proposed Changes**
- Related https://github.com/Checkmarx/kics/pull/6897
  - Related load_balancer_use_insecure_tls_policy_***
  - In the load_balancer terraform, we almost always enter only one of the ssl_policy_id or ssl_policy_name.
  - Therefore, in the current configuration, an error occurs if both are not entered, so I would like to make it so that even blank entries are considered correct.


ArturRibeiro-CX:
Hey @tunakyonn, 
Thanks for contributing again to KICS, it's always great to see the community come back 😃.

After an initial analysis, while `ssl_policy_id` and `ssl_policy_name` are optional in Terraform, missing values could indicate a lack of security best practices. To enhance security control, our KICS queries will be refined to better distinguish missing values from insecure policy configurations.

After our AppSec team approves, we will make the following next steps: 
- Update the queries to flag cases where both ssl_policy_id and ssl_policy_name are missing, ensuring security risks are properly identified;
- Consider implementing a new check that raises a warning when neither field is provided, rather than treating it as a full misconfiguration.

We'll keep you updated as we move forward with these changes.
Thanks again for your contribution! 🚀

PR #7313, opened 2025-01-16
feat(engine): add new QueryID pattern

cx-ruiaraujo:
- CxOne Platform generates custom KICS queries with the following ID pattern `<prefix>:<uint64>`
- This is not supported by KICS when using the flag `-i, --include-queries`

**Proposed Changes**
- Add support for new QueryID pattern


test/fixtures/new_queryid_validation/metadata.json
```diff
@@ -0,0 +1,12 @@
+{
+  "id": "t:8820143918834007824",
+  "queryName": "Last User Is 'root'",
+  "severity": "HIGH",
+  "category": "Best Practices",
+  "descriptionText": "Leaving the last user as root can cause security risks. Change to another user after running the commands the need privileges",
```
ArturRibeiro-CX, line 6: Hi Rui, great changes overall and a great approach to adding granularity to the query validation!
I just had a small note about this description. While I understand that the original query has the same phrasing, we could improve its clarity and grammar. Instead of 'the need privileges,' we could change it to 'that need privileges.'
I would also suggest updating the original query, this test case, and any others that use this query, if you find it suitable.
Thanks!

ArturRibeiro-CX (changes requested):
Read my comment please 💪. 
Thanks!

test/fixtures/new_queryid_validation/metadata.json
cx-ruiaraujo: Hi @ArturRibeiro-CX ,
Resolved in commit f44aaa7a4cbb47bfa5f3c7bae8fc1e49bcc49d19
I also did a quick search for 'the need' keywords but did not find other similar cases.
Thanks 🚀

ArturRibeiro-CX (approved):
Great changes 😃

cxMiguelSilva (approved):
LGTM

PR #7322, opened 2025-01-31
update(dockerfile): revert KICS user change from 65532 back to root

cx-ruiaraujo:
- KICS always used root, but in the latest release, we changed it to 65532 due to FedRAMP policies. This change is causing issues in AST-CLI. Since we only use the binaries inside the image, there is no relevant reason to use 65532 instead of root. A possible fix is to run the KICS image with the same user as the container running it. However, since this would require users to change their setup, I am reverting this change to avoid that and to prevent other potential issues.

**Proposed Changes**
- revert KICS user change from 65532 back to root
- update KICS gh action to v2.1.4
- fix vulnerabilities

### DIFF
--- a/.github/scripts/report/go.mod
+++ b/.github/scripts/report/go.mod
@@ -1,6 +1,6 @@
 module github.com/Checkmarx/e2e-report
 
-go 1.23.1
+go 1.23.5
 
 require (
 	github.com/rs/zerolog v1.31.0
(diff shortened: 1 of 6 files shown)


EduardoSemanas (approved):
LGTM

ArturRibeiro-CX (approved):
I would much rather prefer the solution you provided "A possible fix is to run the KICS image with the same user as the container running it" but I understand the problem with that.
LGTM eitherway.

PR #7328, opened 2025-02-03
fix(query): correct keyActualValue and keyExpectedValue for maxItems validation

ArturRibeiro-CX:
- Actual and Expected values are swapped.

**Proposed Changes**
- Swap Actual and Expected values on query results.


EduardoSemanas (approved):
LGTM

cx-ruiaraujo (approved):
LGTM

PR #7350, opened 2025-02-26
fix(query): fix FP in openAPI Invalid Media Type Value query

ArturRibeiro-CX:
- Invalid Media Type Value openAPI query was incorrectly flagging when `content` was not an object;

**Proposed Changes**
- fix FP on Invalid Media Type Value openAPI query which now checks if the content is an object or not;
- fix documentation to properly explain how a media type should be constructed;
- add negative unit tests to ensure no regression issues;
- fix typo on expectedValue (from `a` to `an`);


EduardoSemanas (approved):
LGTM

PR #7361, opened 2025-03-01
fix(documentation): update descriptionText metadata field for invalid media type openAPI query 

ArturRibeiro-CX:
- The correct media type format should be \<type>/\<subtype>[+suffix][;parameters], but the single quotes (') were missing.
- This caused the displayed information to be inaccurate.
- See the issue in the attached screenshot.
image

**Proposed Changes**
- Update the query documentation to correctly display the media type value pattern, ensuring it reflects the intended use case (v2 and v3).
image


cx-ruiaraujo (approved):
LGTM

PR #7423, opened 2025-04-02
fix(engine): direct circular dependency is causing resource exhaustion

cx-miguel-silva:
Reason for Proposed Changes
- undetected direct circular references

Proposed Changes:
- fix detection for direct circular references with relative path


cx-bruno-silva (approved):
LGTM

cx-rui-araujo (approved):

PR #7450, opened 2025-04-25
fix(queries): support all valid CloudWatch Logs retention periods

jamesbascle:
Pulled from latest documentation @ https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudwatch_log_group

Closes #

- Some valid CloudWatch Logs retention periods are not supported and trigger false positives

**Proposed Changes**
- Adds the full list of valid CloudWatch Logs retention periods

### DIFF
--- a/assets/queries/ansible/aws/cloudwatch_without_retention_period_specified/query.rego
+++ b/assets/queries/ansible/aws/cloudwatch_without_retention_period_specified/query.rego
@@ -29,7 +29,7 @@ CxPolicy[result] {
 	ansLib.checkState(cloudwatchlogs_log_group)
 	value := cloudwatchlogs_log_group.retention
 
-	validValues = [1, 3, 5, 7, 14, 30, 60, 90, 120, 150, 180, 365, 400, 545, 731, 1827, 3653]
+	validValues = [1, 3, 5, 7, 14, 30, 60, 90, 120, 150, 180, 365, 400, 545, 731, 1096, 1827, 2192, 2557, 2922, 3288, 3653]
 
 	not commonLib.inArray(validValues, value)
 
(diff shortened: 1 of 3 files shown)


jamesbascle:
Sorry, I do not understand what the 
Capture

check is asking me to do.  Happy to fix with guidance.

cx-rui-araujo:
Hi @jamesbascle, how are you?

Thanks for your contribution 🙌
We also have this query for two other platforms besides Crossplane:

- Ansible - [/assets/queries/ansible/aws/cloudwatch_without_retention_period_specified/query.rego](https://github.com/Checkmarx/kics/blob/v2.1.7/assets/queries/ansible/aws/cloudwatch_without_retention_period_specified/query.rego#L32)
- Terraform - [/assets/queries/terraform/aws/cloudwatch_without_retention_period_specified/query.rego](https://github.com/Checkmarx/kics/blob/v2.1.7/assets/queries/terraform/aws/cloudwatch_without_retention_period_specified/query.rego#L27)

Would you like to update them too, or would you prefer that I handle it?

jamesbascle:
Hey! I'll try to update those tomorrow when I'm back at my computer.  Thanks

jamesbascle:
@cx-rui-araujo The unit test runner failed on this...I don't really know what i'm looking at here, but it doesn't seem related to the change I made.  Any advice?

cx-artur-ribeiro (approved):
LGTM! 
As @cx-rui-araujo explained, I don't see a reason why we should be testing the full list of values, but we could improve the unit tests with better use cases... Either way, we can do that in the future; doesn't need to be added in this PR.  

Thanks for your contribution!

PR #7455, opened 2025-05-02
fix(query): fn for s3_bucket_allows_delete_action_from_all_principals query

cx-artur-ribeiro:
- The existing query for detecting S3 buckets that allows delete actions from all principals did not account for plural forms of keys such as "principals" and "actions";
- This caused the query to miss policy violations when those plural forms were used.

**Proposed Changes**
- Updated the internal Terraform library to handle both singular and plural forms: "principal" / "principals" and "action" / "actions";
- Modified the existing query (S3 Bucket Allows Delete Action From All Principals) to detect these variants correctly;
- Added additional policy examples to the query's test files (3 positive, one negative) to ensure plural form support is covered;
- Included unit tests to validate the new parsing logic for both positive and negative scenarios;
- Added coverage for actions "\*" and "s3:*";


assets/libraries/terraform.rego
```diff
@@ -538,7 +561,41 @@ allows_action_from_all_principals(json_policy, action) {
 	statement := st[_]
 	statement.Effect == "Allow"
     anyPrincipal(statement)
-    common_lib.containsOrInArrayContains(statement.Action, action)
+    common_lib.containsOrInArrayContains(get_action(statement), action)
+}
+
+allows_action_from_all_principals_match(json_policy) {
```
cx-miguel-silva, line 567:
```suggestion
allows_all_s3_actions_from_all_principals_match(json_policy) {
```

cx-miguel-silva (changes requested):

cx-miguel-silva (approved):
LGTM 🚀

cx-eduardo-semanas (approved):
LGTM

PR #7456, opened 2025-05-05
fix(query): fix fn for S3_Bucket_Allows_Public_Policy query

cx-artur-ribeiro:
- Enhanced query logic to meet updated security requirements for `block_public_policy` handling;
- Fixed gaps where `aws_s3_account_public_access_block` was defined but did not explicitly set block_public_policy to true;
- Ensured that vulnerable configurations are flagged only when the account-level block is present and improperly configured, in combination with an explicitly insecure bucket-level setting.

**Proposed Changes**
- Added logic to validate that `aws_s3_account_public_access_block` is defined and does not explicitly set block_public_policy to true;
- Added logic to flag a result only when `aws_s3_bucket_public_access_block` explicitly sets block_public_policy to false;
- Improved handling for scenarios where the bucket-level resource is present but the account-level policy is misconfigured;
- Added comprehensive positive and negative test cases to ensure coverage of both presence and configuration correctness across resource types.

**Updated Logic** 
The current rule (S3 Bucket Allows Public Policy) flags a result if and only if:
- `aws_s3_bucket_public_access_block` explicitly sets block_public_policy = false;
- `aws_s3_account_public_access_block` is defined, but does not explicitly set block_public_policy = true.


cx-eduardo-semanas (dismissed):
LGTM, nice approach!

assets/queries/terraform/aws/s3_bucket_with_public_policy/query.rego
```diff
@@ -3,69 +3,31 @@ package Cx
...
 import data.generic.terraform as tf_lib
 
-#default of block_public_policy is false
+# checks if aws_s3_account_public_access_block and aws_s3_bucket_public_access_block resources exist
+# checks if block_public_policy is set to false on bucket-level and account-level is set to false or missing
 CxPolicy[result] {
-	pubACL := input.document[i].resource.aws_s3_bucket_public_access_block[name]
-	not common_lib.valid_key(pubACL, "block_public_policy")
+	pubACL := input.document[i].resource.aws_s3_account_public_access_block[account_name]
+    pubBucket := input.document[i].resource.aws_s3_bucket_public_access_block[bucket_name]
```
cx-miguel-silva, line 10:
```suggestion
    pubBucket := input.document[_].resource.aws_s3_bucket_public_access_block[bucket_name]
```

cx-miguel-silva (changes requested):

assets/queries/terraform/aws/s3_bucket_with_public_policy/query.rego
cx-artur-ribeiro: changed in the [last commit](https://github.com/Checkmarx/kics/pull/7456/commits/947f742e074846e7fff2a25c7238d64b8308bbb6)

cx-miguel-silva (approved):
Great Job 🚀

PR #7492, opened 2025-06-12
fix(query): fix fp for mssql_server_auditing_disabled

cx-romeu-silva:
- The actual connection between the auditing policy and the SQL Server is made through the `server_id` field. This field must reference the correct SQL Server instance's ID, not the resource name.

**Proposed Changes**
- In the case an `azurerm_mssql_server_extended_auditing_policy` resource exists, verify its `server_id` points to a valid `azurerm_mssql_server` id.


cx-rui-araujo (approved):
LGTM 💪

PR #7497, opened 2025-06-16
fix(query): fix fp for missing_flag_from_dnf_install

cx-romeu-silva:
- Add support for `tdnf`.

**Proposed Changes**
- Include `tdnf` in the `hasYesFlag` regex.


cx-artur-ribeiro (approved):
Great work Romeu, nice improve to the regex!

cx-rui-araujo (approved):
LGTM 🚀

PR #7502, opened 2025-06-18
fix(query): fix fp for iam_group_without_users

cx-romeu-silva:
- The query currently checks only for the `aws_iam_group_membership` resource to determine if an IAM group has users.
- However, users can also be associated with groups via the `aws_iam_user_group_membership` resource.

**Proposed Changes**
- Update the query to also check for `aws_iam_user_group_membership` associated with the same `aws_iam_group`.
- Report the group only if neither `aws_iam_group_membership` (with users) nor `aws_iam_user_group_membership` exists.
- No need to validate users in `aws_iam_user_group_membership` since it has a required `user` field.


cx-artur-ribeiro (approved):
LGTM

cx-rui-araujo (approved):
LGTM 🚀

PR #7503, opened 2025-06-23
fix(query): fix fn in password and secrets Dockerfile ENV variable cases

cx-eduardo-semanas:
- The query currently only catches hardcoded passwords with "=" or ":" between key and value.
- However, in Dockerfiles is possible to define environment variables with the "=" omitted (ex., ENV PASSWORD testPassword).

**Proposed Changes**
- Added new Passwords and Secrets query for Dockerfile ENV hardcoded password with omitted equals cases


cx-artur-ribeiro (approved):
LGTM

cx-miguel-silva (approved):

PR #7506, opened 2025-06-25
fix(query): fix fn for s3_bucket_without_restriction_of_public_bucket

cx-romeu-silva:
- The query currently only checks the [s3_bucket_public_access_block](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_public_access_block) resource and does not account for the [s3_account_public_access_block](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_account_public_access_block), which also has the `restrict_public_buckets` attribute and can affect the effective configuration.
- This is similar to the issue fixed in #7456, but applied to the `restrict_public_buckets` attribute instead of `block_public_policy`.

**Proposed Changes**
- Taking a similar approach to #7456, the rule (S3 Bucket Without Restriction of Public Bucket) flags a result if and only if:
  - `aws_s3_account_public_access_block` is defined, but does not explicitly set `restrict_public_buckets = true`, and
  - `aws_s3_bucket_public_access_block` explicitly sets `restrict_public_buckets = false`.

### DIFF
--- a/docker/Dockerfile.debian
+++ b/docker/Dockerfile.debian
@@ -41,7 +41,7 @@ HEALTHCHECK CMD wget -q --method=HEAD localhost/system-status.txt
 # be sure to use /app/bin/kics as a path to the binary
 #
 # runtime image
-FROM debian:buster-slim
+FROM debian:stable-slim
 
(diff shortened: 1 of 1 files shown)


cx-artur-ribeiro (approved):
LGTM! Nice work Romeu

PR #7508, opened 2025-06-25
fix(query): improving Volume Mount With OS Directory Write Permissions k8s query

cx-artur-ribeiro:
- Improve the query to not only search for the `readOnly` property, but also for `recursiveReadOnly`;
- `configMaps` and `secrets` should not be detected since they are `readOnly` by default;

**Proposed Changes**
- Add a check for `configMaps` and `secrets` volumes;
- Add a check for the `recursiveReadOnly` property ;
- Update query fields in order to correctly reflect the actual and expected values, as well as the searchKey;
- Add more positive and negative cases to reflect all the possibilities for the query update (missing, vulnerable and not vulnerable cases);

### DIFF
--- a/docs/platforms.md
+++ b/docs/platforms.md
@@ -238,3 +238,35 @@ resource "aws_launch_configuration" "example" {
   user_data_base64 = "a123B=="
 }
 ```
+#### Analyzer Blacklist for Unsupported File Types
+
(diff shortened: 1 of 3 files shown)


cx-rui-araujo (approved):
LGTM! Great work 🐈‍⬛

PR #7510, opened 2025-06-26
fix(query): fix fp for ecs_cluster_not_encrypted_at_rest

cx-romeu-silva:
- Currently, in YAML, the query only supports `TaskDefinition` references written as a single line:
```yaml
TaskDefinition: !Ref TaskDef54694570
```
- However, the reference can also be specified using a nested mapping:
```yaml
TaskDefinition:
  Ref: TaskDef54694570
```
- This change is needed to support both styles, ensuring templates using either format are handled correctly.

**Proposed Changes**
- Update the query to support both formats for specifying `TaskDefinition` references.


cx-miguel-silva (approved):
LGTM

PR #7518, opened 2025-07-04
fix(query): fix fp for s3_bucket_access_to_any_principal

cx-andre-pereira:
- Currently the query ```s3_bucket_access_to_any_principal``` can only determine if a "Principal" has permissions "\*" or "s3:\*" from YAML files if it has no "!If" statements associated.
- With "!If" statements the flag will trigger every time due to lack of specific case handling .  The query does not take into consideration the existence of If´s. 
-  The If structure is as follows : (YAML)
```!If:
    - <CONDITION>
    - <THEN>
    - <ELSE>
```
- The values for ```<THEN>``` and ```<ELSE>``` require scanning that is currently inexistent. Either of these values can have meaningful policy statements for this query.

**Proposed Changes**
- Introduced two helper functions (```handle_if_statements``` and ```handle_if_statement```) to correctly traverse and extract valid policy statements from structures resulting from !If conditions.

- This improves the query’s accuracy by avoiding false positives when Principal is not actually "*" in resolved policy logic, and also guarantees the verifications of each and every meaningful object in the if statement be it a ```<THEN>``` or ```<ELSE>``` associated object.

### DIFF
--- a/pkg/parser/yaml/parser.go
+++ b/pkg/parser/yaml/parser.go
@@ -137,7 +137,12 @@ func processCertContent(elements map[string]interface{}, content, filePath strin
 
 func processElements(elements map[string]interface{}, filePath string) {
 	if elements["certificate"] != nil {
-		processCertContent(elements, utils.CheckCertificate(elements["certificate"].(string)), filePath)
+		certificate, ok := elements["certificate"].(string)
+		if !ok {
+			log.Warn().Msgf("Failed to parse certificate: %s", filePath)
+			return
+		}
+		processCertContent(elements, utils.CheckCertificate(certificate), filePath)
 	}
 }
 
--- a/pkg/parser/yaml/parser_test.go
+++ b/pkg/parser/yaml/parser_test.go
@@ -8,6 +8,7 @@ import (
 	"testing"
 
 	"github.com/Checkmarx/kics/v2/pkg/model"
+	"github.com/pkg/errors"
 	"github.com/stretchr/testify/require"
 )
 
@@ -404,6 +405,7 @@ func TestYaml_processElements(t *testing.T) {
 		args     args
 		wantCert map[string]interface{}
 		wantSwag string
+		wantErr  bool
 	}{
 		{
 			name: "test_process_elements",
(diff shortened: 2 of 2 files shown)


assets/queries/cloudFormation/aws/s3_bucket_access_to_any_principal/query.rego
```diff
@@ -1,5 +1,7 @@
+
```
cx-artur-ribeiro: Can you remove this line please?

assets/queries/cloudFormation/aws/s3_bucket_access_to_any_principal/test/negative3.yaml
```diff
@@ -0,0 +1,75 @@
+
```
cx-artur-ribeiro: I understand this is a test file, but can we also remove the unnecessary trailing newline, please?

assets/queries/cloudFormation/aws/s3_bucket_access_to_any_principal/test/positive3.yaml
```diff
@@ -0,0 +1,74 @@
+
```
cx-artur-ribeiro: Same as above please :)

assets/queries/cloudFormation/aws/s3_bucket_access_to_any_principal/query.rego
```diff
@@ -11,8 +13,9 @@ CxPolicy[result] {
 		resourcePolicy := input.document[indexBucket].Resources[_]
 		resourcePolicy.Type == "AWS::S3::BucketPolicy"
 		check_ref(resourcePolicy.Properties.Bucket, resourceBucket, nameBucket)
-		policy := resourcePolicy.Properties.PolicyDocument
-		st := common_lib.get_statement(common_lib.get_policy(policy))
+		raw_policy := resourcePolicy.Properties.PolicyDocument
+		policy := common_lib.get_statement(common_lib.get_policy(raw_policy))
+        st := handle_if_statements(policy)
```
cx-artur-ribeiro: Can we have a better variable naming for both of these cases please? 
For example (you may change for a better name, these are just suggestions): 

```suggestion
		statements := common_lib.get_statement(common_lib.get_policy(raw_policy))
        clean_statements := handle_if_statements(statements)
```

cx-artur-ribeiro (changes requested):
Hi André! 
First of all, nice work!
Can you take a look at my comments and see if you agree? The suggestions are just quality of life changes, the query seems to work as expected, great job! 

Thanks for the work done!

cx-artur-ribeiro (approved):

PR #7521, opened 2025-07-07
fix(query): fix fp in password and secrets Generic Token

cx-andre-pereira:
- Currently the implementation to prevent flagging of github id-token(s) such as : ```'id-token: write'``` is not detecting relevant instances reliably. 
- The regex used at the moment (```(?i)['\"]?id-token\\s*[:=]\\s*(write|read|none)\\s*$```) is faulty and inconsistent relative to the rest of the regex rules for [passwords_and_secrets's regex_rules](https://github.com/Checkmarx/kics/blob/master/assets/queries/common/passwords_and_secrets/regex_rules.json).
- This regex stands out for its requirement to scan a full line to detect the id-token ($ => \"end of string\"), this way something like a trailing comment will have to be explicitly captured by the regex or a text like : ```permissions: {id-token: write, contents: ...} ``` will also require the capture of unnecessary text. 

**Proposed Changes**
- To better standardize regex logic i propose the implementation of an updated regex: 
    - ```(?i)['\"]?id-token['\"]?\\s*[:=]\\s*(write|read|none)```
- Aside from the obvious removal of the undesirable  ```\\s*$```, i added an extra check ```['\"]?``` after the \"id-token\" key-word to better align the regex composition logic with most other regex rules.
- With this implementation detection of the \"id-token\" is much more reliable and independent of trailing comments or any data structure it might find itself within.
- Finally the 2 new tests positive53 and negative53 verify, respectively, that the id-token is detected when there is a trailing comment and that it is flagged with an invalid command. 

NOTE: the regex are written with double backslashes to represent the exact characters used in the [passwords_and_secrets's regex_rules](https://github.com/Checkmarx/kics/blob/master/assets/queries/common/passwords_and_secrets/regex_rules.json) file strings for they are literal strings, not raw.

### DIFF
--- a/pkg/parser/yaml/parser.go
+++ b/pkg/parser/yaml/parser.go
@@ -137,7 +137,12 @@ func processCertContent(elements map[string]interface{}, content, filePath strin
 
 func processElements(elements map[string]interface{}, filePath string) {
 	if elements["certificate"] != nil {
-		processCertContent(elements, utils.CheckCertificate(elements["certificate"].(string)), filePath)
+		certificate, ok := elements["certificate"].(string)
+		if !ok {
+			log.Warn().Msgf("Failed to parse certificate: %s", filePath)
+			return
+		}
+		processCertContent(elements, utils.CheckCertificate(certificate), filePath)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


cx-artur-ribeiro (approved):
All good to me, we could improve the actual positive results with valid cases, without using comments at the start at the line since the problem is actually with having comments after `id-token: write` (for example).
I'll leave to another reviewer the need to update this test cases or not.
Nice work eitherway!

PR #7522, opened 2025-07-07
fix(query): fix fp for dynamic network_rules for Trusted Microsoft Services Not Enabled

cx-ricardo-jesus:
Closes #

- The current query incorrectly flags a false positive when network_rules are defined using a dynamic block in Terraform.
- This happens because the query only checks for the presence of the network_rules key at the static level, and does not account for cases where the dynamic keywork is used.

**Proposed Changes**
- Updated the query to include checks for the presence of the network_rules key inside the dynamic block, using: `not common_lib.valid_key(resource.dynamic, "network_rules")`.
- Added a validation path or the `network_rules.bypass` inside `dynamic.network_rules.content`, mirroring the logic already applied to the static structure.
- Ensure that the verification that the bypass field should contain 'AzureServices' is performed even when this attribute is part of a dynamic block.
- Preserved existing logic for the resource `azurerm_storage_account_network_rules`, which is a separate resource.


assets/queries/terraform/azure/trusted_microsoft_services_not_enabled/query.rego
```diff
@@ -5,10 +5,12 @@ import data.generic.terraform as tf_lib
 
 CxPolicy[result] {
 	resource := input.document[i].resource.azurerm_storage_account[name]
-	not common_lib.valid_key(resource, "network_rules")
+    not common_lib.valid_key(resource.dynamic, "network_rules") 
+    not common_lib.valid_key(resource, "network_rules") 
 
 	result := {
-		"documentId": input.document[i].id,
+
```
cx-artur-ribeiro, line 10: Can you remove the unnecessary trailing new line here please?

assets/queries/terraform/azure/trusted_microsoft_services_not_enabled/query.rego
```diff
@@ -49,6 +68,24 @@ CxPolicy[result] {
 	}
 }
 
+CxPolicy[result] {
+	resource := input.document[i].resource.azurerm_storage_account[name]
+	bypass := input.document[i].variable.network_rules_list["default"][j].bypass
```
cx-artur-ribeiro: Although the current code works, it incorrectly assumes that both the resource and the variable are defined within the same document (`input.document[i]`). In practice, Terraform configurations are often split across multiple files, meaning the variable and resource definitions might reside in different entries of input.document.

Furthermore, the index `j` is not used for any specific purpose, so it would be clearer and more idiomatic to iterate using `[_]` instead.

assets/queries/terraform/azure/trusted_microsoft_services_not_enabled/query.rego
```diff
@@ -49,6 +68,24 @@ CxPolicy[result] {
...
+
+    not common_lib.inArray(bypass, "AzureServices")
+
+	result := {
+		"documentId": input.document[i].id,
+		"resourceType": "azurerm_storage_account",
+		"resourceName": tf_lib.get_resource_name(resource, name),
+		"searchKey": sprintf("azurerm_storage_account[%s].dynamic.network_rules.content.bypass", [name]),
```
cx-artur-ribeiro: Is the `searchKey` here correct? Based on the context, it seems it should be targeting the default block rather than dynamic.
Could this be a copy-paste oversight from a different rule that handles dynamic instead of default?

cx-artur-ribeiro (changes requested):
Hey Ricardo, 
First of all, nice job tackling different available options for terraform azure instances!

Can you check my comments and see if you agree, please?
Thanks!

assets/queries/terraform/azure/trusted_microsoft_services_not_enabled/test/positive_expected_result.json
```diff
@@ -1,12 +1,38 @@
...
+  {
+    "queryName": "Trusted Microsoft Services Not Enabled",
+    "severity": "MEDIUM",
+    "line": 24,
+    "fileName": "positive2.tf"
+  },
+  {
+    "queryName": "Trusted Microsoft Services Not Enabled",
+    "severity": "MEDIUM",
+    "line": 1,
```
cx-artur-ribeiro: I believe we should be pointing to line 11 here, since the bypass field exists but has a vulnerable value.
When the bypass field is defined, we should point to that field. If it isn't, we should point to the resource or variable that uses the field, to maintain the same behavior the query already had.
Let me know if you disagree.

assets/queries/terraform/azure/trusted_microsoft_services_not_enabled/test/positive_expected_result.json
```diff
@@ -1,12 +1,38 @@
...
+    "fileName": "positive3.tf"
+  },
+  {
+    "queryName": "Trusted Microsoft Services Not Enabled",
+    "severity": "MEDIUM",
+    "line": 1,
```
cx-artur-ribeiro: Same as above.

cx-artur-ribeiro (changes requested):
Hey Ricardo, take a look at my comments please.

assets/queries/terraform/azure/trusted_microsoft_services_not_enabled/query.rego
```diff
@@ -19,14 +20,20 @@ CxPolicy[result] {
...
 
 CxPolicy[result] {
-	network_rules := input.document[i].resource.azurerm_storage_account[name].network_rules
-	not common_lib.valid_key(network_rules, "bypass")
-
+	resource := input.document[i].resource.azurerm_storage_account[name]
+    
+    res := has_bypass_field(resource, name)
+    
+    searchKey_val := res["search_key_value"]
```
cx-eduardo-semanas: It seems unnecessary to be creating a new variable inside the policy logic to store the searchKey (which is metadata for the response). 

I think it would make more sense to insert the res["search_key_value"] directly in the result.searchKey.

cx-eduardo-semanas (approved):
LGTM

PR #7525, opened 2025-07-08
fix(query): fix fp for web app not using tls last version

cx-andre-pereira:
- Simply put TLS has updated to 1.3 and the query has not yet been adjusted to account for that fact. (https://www.rfc-editor.org/rfc/rfc8446)
- The metadata, query and tests are now out of date.

**Proposed Changes**
- I have replaced all references to version 1.3 and it is now the version that is verified to determine whether the flag must be raised or not.


cx-artur-ribeiro (approved):
LGTM!

PR #7526, opened 2025-07-08
fix(parser): add type assertion verification to certificate elements process

cx-artur-ribeiro:
- The `processElements` function was causing a panic when the certificate field in YAML files was structured as a map/object instead of a string, due to unsafe type assertion (`elements["certificate"].(string)`) which failed with "interface conversion: interface {} is map[string]interface {}, not string";

**Proposed Changes**
- Replace unsafe type assertion with safe type assertion;
- Add warning log message when certificate field is not a string type to provide better debugging information;
- Add early return when type assertion fails to prevent further processing and potential errors;
- Add comprehensive test cases to verify both successful processing (certificate as string) and graceful handling of error cases (certificate as map);

### DIFF
--- a/pkg/parser/yaml/parser.go
+++ b/pkg/parser/yaml/parser.go
@@ -137,7 +137,12 @@ func processCertContent(elements map[string]interface{}, content, filePath strin
 
 func processElements(elements map[string]interface{}, filePath string) {
 	if elements["certificate"] != nil {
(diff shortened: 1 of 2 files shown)


cx-bruno-silva (approved):
LGTM

cx-rui-araujo (approved):

PR #7530, opened 2025-07-09
fix(query): fix fp for api_gateway_method_does_not_contains_an_api_key

cx-andre-pereira:
- The current implementation of the query flags every declaration of an **AWS API Gateway method** (in Terraform) if it either omits the ```api_key_required``` field or sets it to a value other than ```true```.
- However, the HTTP method \"OPTIONS\",commonly used for Cross-Origin Resource Sharing (CORS) preflight requests, cannot , let alone require having this field set to true.
- As a result, the current logic incorrectly flags valid \"OPTIONS\" methods as **F**alse **P**ositives.

**Proposed Changes**
- To prevent these **FP**s, the query has been updated to include a conditional check: if the http_method is \"OPTIONS\", the rule will no longer raise a flag, regardless of the ```api_key_required``` value.
- This change ensures that CORS preflight configurations are not mistakenly flagged.

**Note**: it was considered an extra check to ensure the field is actually set to false if the HTTP method is \"OPTIONS\", but due to not representing a security flaw / representing an idea that directly opposes the objective of this query (**i.e.** guaranteeing a field is set to ```false``` when the whole purpose of the query is assuring it is set to ```true```).


cx-artur-ribeiro (approved):
LGTM, nice clean changes!

cx-eduardo-semanas (approved):
LGTM

PR #7531, opened 2025-07-09
fix(query): fix fp for image_version_not_explicit

cx-andre-pereira:
- This query exists to guarantee that any image referenced in a \"dockerfile\" is either an explicit image (```ubuntu:22.04```) or a non explicit image that has been declared in a valid manner.
- Given this, the type of expected statements to look out for are as follows:   
```
FROM ubuntu:22.04 
FROM ubuntu:22.04 AS new_image
FROM declared_image
FROM declared_image AS new_image
```
- The query must enforce 2 similar rules to ensure valid image referencing and declaration: 
   - No ```FROM x AS y``` statement has an invalid x image. 
   - No ```FROM x``` statement has an invalid x image. 
- Any \"x\" image can be either explicit image declared previously through a chain that ends in an explicit image, that is to say all images must be based, originally, on explicit images. 
- Currently the query fails when a \"FROM x\" statement is correctly used with an x representing a valid declared image, it is not verifying previous declarations to compare validity or lack there of.
- Additionally the current implementation does not take into account sequential declaration; this is to say something like : 
```
FROM Image1
FROM ubuntu:22.04 AS Image1
```
will not raise the flag even though it should; a \"dockerfile\" does not support forward reference and so the query should definitely consider the image version not explicit.

**Proposed Changes**
- To solve the issue in \"FROM x\" statements being incorrectly flagged, i simply updated the current logic based on the identical use case already implemented for \"FROM x AS y\" statements.
- For the issue related to invalid assumption of forward referencing i made use of the \"EndLine\" field located in each \"command\" generated for the payload, which allows for easy distinction between what should be accounted (**i.e.** declarations written before the one being considered) and the ones written later in the \"dockerfile\" which should be ignored. 
- With this implementation any complete hierarchical definition of images in the \"dockerfile\" can be correctly assessed and new tests have been added to corroborate this behavior: 
   - **positive3** : Flags \"FROM x\" commands dependent on undeclared images;
   - **positive4**: Flags invalid attempt at forward referencing for both relevant commands;
   - **negative4**: Demonstrates support for referencing an image generated through hierarchical declarations;  


assets/queries/dockerfile/image_version_not_explicit/query.rego
```diff
@@ -16,13 +16,15 @@ CxPolicy[result] {
 	}
 }
 
-versionNotExplicit(cmd) {
+versionNotExplicit(cmd,line) {
 	count(cmd) == 1
 	regex.match("^\\$[{}A-z0-9-_+].*", cmd[0]) == false
 	not contains(cmd[0], ":")
+    count([x | x := input.document[i].command[name][_]; input.document[i].command[name][_].EndLine < line; build_name_exists(x, cmd[0])]) == 0
+    #
```
cx-artur-ribeiro: I believe this was a typo, can you remove the `#` please?

assets/queries/dockerfile/image_version_not_explicit/query.rego
```diff
@@ -5,7 +5,7 @@ CxPolicy[result] {
 	resource.Cmd == "from"
 	not resource.Value[0] == "scratch"
 
-	versionNotExplicit(resource.Value)
+	versionNotExplicit(resource.Value,resource.EndLine)
```
cx-artur-ribeiro, line 8: Shouldn't this line be `resource._kics_line` since we want to compare if the value at the end of the line is lower than the actual line?
We are now comparing if the EndLine is lower than EndLine.
Let me know what you think.

cx-artur-ribeiro (changes requested):
Hey André, take a look at my comments please and let me know what you think.

assets/queries/dockerfile/image_version_not_explicit/query.rego
cx-andre-pereira: Hi, in that policy ```versionNotExplicit```, the \"EndLine\" of the resource value is being compared to other dockerfile command´s \"EndLine\" value to ensure that, in case a the image reference is not explicit, it has been declared in a line previous to the current line.

PR #7536, opened 2025-07-10
fix(query): fn for ssh_is_exposed_to_the_internet and rdp_is_exposed_to_the_internet

cx-andre-pereira:
- These queries are meant to detect azurerm network rules that expose sensitive ports: ```port 22(ssh)``` and ```port 3389(rdp)```, however they are not taking into account ```azurerm_network_security_group´s```. 
- This has bestowed upon the queries a massive blind spot for this dangerous vulnerabilities.

**Proposed Changes**
- To allow for complete thorough search for network rules i have implemented an extra valid pattern for detecting these statements when inside a ```azurerm_network_security_group``` block. 
- All the tests have been updated to ensure logic is consistently whether rules are defined inside or outside of a group.

**Note:** The searchKey uses rule names for traceability. This approach was chosen over index-based referencing due to consistent misalignment between ```idx``` values and the actual rule objects, which led to duplicate or misplaced warnings. 

### DIFF
--- a/assets/queries/terraform/azure/rdp_is_exposed_to_the_internet/query.rego
+++ b/assets/queries/terraform/azure/rdp_is_exposed_to_the_internet/query.rego
@@ -22,6 +22,30 @@ CxPolicy[result] {
 	}
 }
 
+CxPolicy[result] {
+  group := input.document[i].resource.azurerm_network_security_group[groupName]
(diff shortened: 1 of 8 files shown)


cx-eduardo-semanas (approved):
LGTM

PR #7538, opened 2025-07-11
fix(query): fix fn for cloudTrail_multi_region_disabled

cx-andre-pereira:
- This query is meant to identify ansible tasks that configure an AWS CloudTrail trail with the ```is_multi_region_trail``` flag set to false.
- In its current implementation it does not treat the lack of the relevant field (```is_multi_region_trail```) as false, it should do so because that is its default value.
- This behavior leads to false negatives.

**Proposed Changes**
- Refinement of the query`s ```CxPolicy``` logic to flag whether the field is set explicitly to false or missing altogether from the ansible task.

### DIFF
--- a/assets/queries/ansible/aws/cloudtrail_multi_region_disabled/query.rego
+++ b/assets/queries/ansible/aws/cloudtrail_multi_region_disabled/query.rego
@@ -20,3 +20,22 @@ CxPolicy[result] {
 		"keyActualValue": "cloudtrail.is_multi_region_trail is false",
 	}
 }
+
+CxPolicy[result] {
+	task := ansLib.tasks[id][t]
+	modules := {"community.aws.cloudtrail", "cloudtrail"}
+	cloudtrail := task[modules[m]]
+	ansLib.checkState(cloudtrail)
+
+	not cloudtrail.is_multi_region_trail
+
+	result := {
+		"documentId": id,
+		"resourceType": modules[m],
+		"resourceName": task.name,
+		"searchKey": sprintf("name={{%s}}.{{%s}}.is_multi_region_trail", [task.name, modules[m]]),
+		"issueType": "IncorrectValue",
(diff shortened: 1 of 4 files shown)


assets/queries/ansible/aws/cloudtrail_multi_region_disabled/query.rego
```diff
@@ -20,3 +20,22 @@ CxPolicy[result] {
...
+	cloudtrail := task[modules[m]]
+	ansLib.checkState(cloudtrail)
+
+	not cloudtrail.is_multi_region_trail
+
+	result := {
+		"documentId": id,
+		"resourceType": modules[m],
+		"resourceName": task.name,
+		"searchKey": sprintf("name={{%s}}.{{%s}}.is_multi_region_trail", [task.name, modules[m]]),
```
cx-artur-ribeiro: I believe we should update these three fields, as the reason for the vulnerability is that the field is not defined (even though the default value is false when it’s undefined).
The motivation for this change is to align with other query implementations and ensure consistency in vulnerability text.

In my opinion, to maintain the same level of user remediation and information, we should point to name={{%s}}.{{%s}} instead of name={{%s}}.{{%s}}.is_multi_region_trail. We should also update the query fields to:
```suggestion
		"keyExpectedValue": "cloudtrail.is_multi_region_trail should be defined and set to true",
		"keyActualValue": "cloudtrail.is_multi_region_trail is undefined",
```

cx-artur-ribeiro (changes requested):
Hey André, can you take a look at my comment and see what you think please?

assets/queries/ansible/aws/cloudtrail_multi_region_disabled/query.rego
cx-eduardo-semanas: I agree, otherwise LGTM

cx-eduardo-semanas (approved):
LGTM

PR #7558, opened 2025-07-16
fix(query): fix fn for cloudTrail_multi_region_disabled

cx-andre-pereira:
- This query is meant to identify ansible tasks that configure an AWS CloudTrail trail with the ```is_multi_region_trail``` flag set to false.
- In its current implementation it does not treat the lack of the relevant field (```is_multi_region_trail```) as false, it should do so because that is its default value.
- This behavior leads to false negatives.

**Proposed Changes**
- Refinement of the query`s ```CxPolicy``` logic to flag whether the field is set explicitly to false or missing altogether from the ansible task.


cx-artur-ribeiro (approved):
LGTM

PR #7563, opened 2025-07-16
fix(query): added one extra verification on the ECS Cluster Not Encrypted At Rest query

cx-ricardo-jesus:
- The query correctly detected when the `TransitEncryption` field in the `EFSVolumeConfiguration` block was explicitly set to `"DISABLED"`.
- However, it did not handle the case where the `TransitEncryption` field is missing.
- According to the AWS documentation, when `TransitEncryption` is not defined, the default behavior is to treat it as `"DISABLED"`, which poses a security risk.
- Therefore, the query need to be extended in order to also report a vulnerability when `TransitEncryption` is not defined.

**Proposed Changes**
- Created the `is_transit_encryption_disabled` helper function to check if the `TransitEncryption` field is set to `"DISABLED"` (explicitly written or by default).
- Updated the first policy, adding a verification that uses the helper function created with the purpose of handling the case described above. 
- I also ensured that the expected issue messages clearly refer to the default behavior when the field is not set.
- Added one extra positive test in json format, with the field `TransitEncryption` not set, which should return a positive result and warning that when this field is not set, the default value is set to `DISABLED`.


cx-eduardo-semanas (approved):
LGTM

PR #7565, opened 2025-07-16
feat(query): added new query: ElasticSearch Without Audit Logs - cloudFormation/aws

cx-andre-pereira:
- Currently there is no check for the Audit Logs of Elasticsearch CloudFormation resources to be set to true.
- There is a very similar query for "[slow logs](https://github.com/Checkmarx/kics/tree/master/assets/queries/cloudFormation/aws/elasticsearch_without_slow_logs)" already implemented.
- It is to note this query does check for more than a since resource type, those being: "[AWS::OpenSearchService::Domain](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-resource-opensearchservice-domain.html)" and "[AWS::Elasticsearch::Domain](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-resource-elasticsearch-domain.html)"; both have equivalent [```LogPublishingOption```](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-resource-opensearchservice-domain.html#cfn-opensearchservice-domain-logpublishingoptions) fields which define whether the audit logs are enabled or not through the "AUDIT_LOGS" key.

**Proposed Changes**
- I simply altered the slow logs query to account for audit logs instead.
- The tests are also based on the slow logs query with simple alteration of some fields to fit the context of the new query, plus some tests for the "AWS::OpenSearchService::Domain" resource type.


cx-artur-ribeiro (approved):
LGTM

PR #7569, opened 2025-07-17
fix(query): added one more allow rule on Generic Password query to allow passwords retrieved from ARM parameters

cx-ricardo-jesus:
- The query `Passwords and Secrets - Generic Password`, incorrectly flags a string that collects a password from parameters which is not correct.

**Proposed Changes**
- Added a new `allowRule` with a regex to detect passwords which are passed through ARM parameters via `parameters('parameterName')` syntax in order to prevent false positives in this cases.
- Also, added a new negative test called `"negative53.json"`, which has a line that creates a connection string and collects the password externally.
- Added one extra positive test called `positive53.json`, that instead of having parameter reference, has a hardcoded password which should return a positive result on line 54.

### DIFF
--- a/.github/workflows/go-e2e.yaml
+++ b/.github/workflows/go-e2e.yaml
@@ -12,7 +12,7 @@ jobs:
       matrix:
         go-version: [1.24.x]
         os: [ubuntu-latest]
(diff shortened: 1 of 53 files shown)


cx-artur-ribeiro (approved):
LGTM

PR #7587, opened 2025-07-24
fix(query): fn for Trusted Microsoft Services Not Enabled  - ARM

cx-andre-pereira:
- Currently this query checks if ```Microsoft.Storage/storageAccounts``` resources by assuring the field ```properties.networkAcls.defaultAction``` is not set as "Allow"  , and that ```properties.networkAcls.bypass``` does not contain "AzureServices" as either of those scenarios would enable "Trusted Microsoft Services". 
- The query , however, does not account for the possibility of the resource missing the ```properties```,```properties.networkAcls``` or ```properties.networkAcls.defaultAction/bypass``` fields entirely. 
 

**Proposed Changes**
- I have implemented a "path_check" function that confirms the path to the ```defaultAction``` and/or ```bypass``` nested fields exists. Moreover the lack of a complete path will properly flag the **Microsoft Services as Not Enabled** and point to the "deepest" field along the path route(either```""```,```"properties"``` or ```"properties.networkAcls"```) on the "searchKey".

Note: I could not find explicit documentation stating the effects of missing "networkAcls" or other fields, but it seems to result in unexpected behavior depending on file/system context which, i feel, still makes this flag pertinent.

### DIFF
--- a/assets/queries/azureResourceManager/trusted_microsoft_services_not_enabled/query.rego
+++ b/assets/queries/azureResourceManager/trusted_microsoft_services_not_enabled/query.rego
@@ -60,6 +60,8 @@ path_check(resource) = "" {
   	resource.properties == {} 
   } else = ".properties" {
   	not common_lib.valid_key(resource.properties,"networkAcls")
-  } else = ".properties.networkAcls" {
-  	resource.properties.networkAcls == {} 
-  } else = "there_is_complete_path"
+  } else = "there_is_complete_path" {
+	common_lib.valid_key(resource.properties.networkAcls,"defaultAction")
+  } else = "there_is_complete_path" {
+	common_lib.valid_key(resource.properties.networkAcls,"bypass")
+  } else = ".properties.networkAcls"
--- a/assets/queries/azureResourceManager/trusted_microsoft_services_not_enabled/query.rego
+++ b/assets/queries/azureResourceManager/trusted_microsoft_services_not_enabled/query.rego
@@ -55,11 +55,11 @@ contains_azure_service(bypass) {
 }
 
 path_check(resource) = "" {
-  not common_lib.valid_key(resource,"properties")
- } else = ".properties" {
-  resource.properties == {} 
-} else = ".properties" {
-  not common_lib.valid_key(resource.properties,"networkAcls")
-} else = ".properties.networkAcls" {
-  resource.properties.networkAcls == {} 
-} else = "there_is_complete_path"
+  	not common_lib.valid_key(resource,"properties")
+  } else = ".properties" {
+  	resource.properties == {} 
+  } else = ".properties" {
+  	not common_lib.valid_key(resource.properties,"networkAcls")
+  } else = ".properties.networkAcls" {
+  	resource.properties.networkAcls == {} 
+  } else = "there_is_complete_path"
--- a/assets/queries/azureResourceManager/trusted_microsoft_services_not_enabled/query.rego
+++ b/assets/queries/azureResourceManager/trusted_microsoft_services_not_enabled/query.rego
@@ -4,6 +4,28 @@ import data.generic.common as common_lib
 import data.generic.azureresourcemanager as arm_lib
 
 CxPolicy[result] {
+	#case of the path /properties/networkAcls/defaultAction_or_bypass not existing
+	doc := input.document[i]
+	[path, value] = walk(doc)
+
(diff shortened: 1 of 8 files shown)


assets/queries/azureResourceManager/trusted_microsoft_services_not_enabled/query.rego
```diff
@@ -3,6 +3,28 @@ package Cx
...
+
+    variable_path := path_check(value)
+	variable_path != "there_is_complete_path"
+
+	result := {
+		"documentId": input.document[i].id,
+		"resourceType": value.type,
+		"resourceName": "value.name",
+		"searchKey": sprintf("%s.name=%s%s", [common_lib.concat_path(path), value.name, variable_path]),
+		"issueType": "IncorrectValue",
```
cx-artur-ribeiro: I believe we should be using "MissingAtribute" instead of "IncorrectValue" as the vulnerability arives from not having the fields `properties.networkAcls` defined.

assets/queries/azureResourceManager/trusted_microsoft_services_not_enabled/query.rego
```diff
@@ -3,6 +3,28 @@ package Cx
...
+		"keyExpectedValue": sprintf("resource with type 'Microsoft.Storage/storageAccounts' should have 'Trusted Microsoft Services' enabled",[]),
+		"keyActualValue": "resource with type 'Microsoft.Storage/storageAccounts' doesn't have 'Trusted Microsoft Services' enabled",
+		"searchLine": common_lib.build_search_line(path, ["properties", "networkAcls"]),
```
cx-artur-ribeiro: If you run a scan and check the searchLine (for example on the file positive5.bicep), you can check that the `sarch_line` field comes with -1 value. 

image

This is not the expected result. In this case, the problem is that you are trying to point to a field that does not exist.
That's the reason that multiple queries have more than one policy so we can signal the correct vulnerable line.

You either build multiple policies to tackle this or use an auxiliar function that builds the fields. For example, if the file only has properties and not networkAcls, you would place only properties on the `build_search_line` function.
For these cases, you should also add the field `searchValue` to the output result with the field you are looking for. For the example case above, I would put as shown below: 

Case where properties is defined but networkAcls is not.
```suggestion
		"searchValue": "networkAcls", 
		"searchLine": common_lib.build_search_line(path, ["properties"]),
```

assets/queries/azureResourceManager/trusted_microsoft_services_not_enabled/query.rego
cx-andre-pereira: I have replaced the issueType - https://github.com/Checkmarx/kics/pull/7587/commits/a7cf0f3a854474ede659064509b26ba3de405659 |

assets/queries/azureResourceManager/trusted_microsoft_services_not_enabled/query.rego
cx-andre-pereira: This behavior should be avoided now with the new "searchLine_vars" dynamic array.

assets/queries/azureResourceManager/trusted_microsoft_services_not_enabled/query.rego
```diff
@@ -4,11 +4,36 @@ import data.generic.common as common_lib
...
+	searchLine_vars := [split | split := raw_split[_]; split != ""]
+
+
+	result := {
+		"documentId": input.document[i].id,
+		"resourceType": value.type,
+		"resourceName": "value.name",
+		"searchKey": sprintf("%s.name=%s%s", [common_lib.concat_path(path), value.name, variable_path]),
+		"issueType": "MissingAttribute",
+		"keyExpectedValue": sprintf("resource with type 'Microsoft.Storage/storageAccounts' should have 'Trusted Microsoft Services' enabled",[]),
```
cx-artur-ribeiro: We should make this value a static string instead of using the sprintf with no arguments.
```suggestion
		"keyExpectedValue": "resource with type 'Microsoft.Storage/storageAccounts' should have 'Trusted Microsoft Services' enabled",
```

cx-andre-pereira: Removed in https://github.com/Checkmarx/kics/pull/7587/commits/47cfb278449a3aaffed46f62e556c30817a93875

PR #7590, opened 2025-07-25
fix(query): fn for SQL Server Database Without Auditing - ARM

cx-andre-pereira:
- This query is meant to flag all SQL server databases with auditing disabled; to do this the query "walks" through all resources and, in case they are of a type included in the ```dbTypes``` array, a search of the resource´s ```child``` resources proceeds aiming to find one with a valid type such as "auditingSettings" , those being registered in the "types" array, and , in case they have a ```state``` set to ```enabled```, we can be sure auditing is not missing and the flag is avoided.
- The problem with this query is simple enough, there are some types not account for. In the ```dbTypes``` array there should be the ```Microsoft.Sql/servers``` type and in the ```types``` array there should be the [```Microsoft.Sql/servers/auditingSettings```](https://learn.microsoft.com/en-us/azure/templates/microsoft.sql/servers/auditingsettings?pivots=deployment-language-bicep) type.

**Proposed Changes**
- I started by simple including the relevant missing types for the ```dbTypes``` and ```types``` arrays. With this the negative tests started to flag due to being dependent on a ```Microsoft.Sql/servers``` that itself triggered the query flag.
- To prevent the negative as well as the positive tests from flagging this way i simply added the auditing resources missing from the servers; additionally a new positive5 test to test this behavior in itself was added.


cx-artur-ribeiro (approved):
LGTM

PR #7591, opened 2025-07-25
fix(query): fixed false negative for "App Service Authentication Disabled" query missing resources

cx-ricardo-jesus:
Closes #

- The query `App Service Authentication Disabled` was not taking into consideration the resources `azurerm_linux_web_app` and `azurerm_windows_web_app` not verifying if the authentication settings for these two resources are enabled.

**Proposed Changes**
- After analyzing the documentation regarding the "azurerm_app_service", "azurerm_linux_web_app" and "azurerm_windows_web_app" resources, I maintained the same verifications for the resource that was already being analyzed and extended the query in order to verify the authentication field on the two other resources.
- Comparing the resources `azurerm_linux_web_app` and `azurerm_windows_web_app` with the resource `azurerm_app_service` I realized that there is another authentication field called `auth_settings_v2` that was not present on the last one. 
- So, taking what I said in the last point into account, I kept the policies that tackles the vulnerabilities on the resource `azurerm_app_service`, and added an extra one that uses a helper function called `prepare_issues`, which verifies not only the `auth_settings` field but also the `auth_settings_v2` field and handles all the cases regarding the new resources that are going to be analyzed in this query for these two fields.
- For the new resources, I kept the same reasoning used on the resource that was already on the query for the `auth_settings` field because, according to the information present on the documentation, this field has equal settings and fields within for all the resources.
- For the field `auth_settings_v2`, there is small things that changes in comparison to the field `auth_settings`, more precisely, the field `auth_settings_v2` not being required and set to `false` by default, which should return a positive result, made me deal with it in a slightly different way, returning a positive result on the query when this field is not defined.
- On top of that, I added eight new samples that should return a positive result and four more Terraform samples that should return a negative result. All of these new test sample are provided in order to cover all the possibilities regarding positive and negative results.
- **NOTE**: When both fields `auth_settings` and `auth_settings_v2` are not defined, I only inserted the `auth_settings` block on the remediation field and not the `auth_settings_v2` but this positive case can be solved defining the field `auth_settings_v2` instead of the `auth_settings` field.


assets/queries/terraform/azure/app_service_authentication_disabled/query.rego
```diff
@@ -45,3 +47,81 @@ CxPolicy[result] {
...
+		"it": "IncorrectValue",
+		"kev": sprintf("'%s[%s].auth_settings_v2.auth_enabled' should be defined to 'true'", [type, name]),
+		"kav": sprintf("'%s[%s].auth_settings_v2.auth_enabled' is defined to 'false'", [type, name]),
+		"rem": json.marshal({
+			"before": "false",
+			"after": "true"
+		}),
+		"rt": "replacement",
+	}
+} else = res { # auth_settings_v2 field defined but without the field auth_field defined
```
cx-artur-ribeiro: I think you wanted to put "auth_enabled" and not "auth_field".
```suggestion
} else = res { # auth_settings_v2 field defined but without the field auth_enabled defined
```

cx-artur-ribeiro (commented):
LGTM, nice work on all the test cases included.
Just left a change on your comment but nothing too important.

PR #7592, opened 2025-07-25
fix(query): fixed FN for the missing resources on "App Service HTTP2 Disabled" query

cx-ricardo-jesus:
- The query `App Service HTTP2 Disabled` was not taking into consideration the resources `azure_rm_linux_web_app` and `azurerm_windows_web_app`, which should be given the necessary support besides the `azurerm_app_service`.

**Proposed Changes**
- After reading carefully the documentation of the resources `azurerm_linux_web_app`, `azure_windows_web_app` and `azurerm_app_service`, I came to a conclusion that the field's which are necessary to analyse regarding the scope of this query have, all of them, the same names and are defined inside the same fields (`http2_enabled` is aways inside a `site_config` block for all of them).
- For all policies, I only made small changes, in order to make the same verifications for the three types of resources through the `resources` variable present on top of this query, changing the first line of each policy to `app := input.document[i].resource[resources[m]][name]`. 
- On top of that I also, provided more four positive samples, two for each of the new resource's that are being analyzed on the query, one of them not having the `http2_enabled` field defined and the other one having the field defined to `false`.
- Also, added two more negative test each one of them having the `site_config.http2_enabled` field defined to `true` for the resources `azurerm_windows_web_app and azurerm_linux_web_app`.

### DIFF
--- a/assets/queries/terraform/azure/app_service_http2_disabled/query.rego
+++ b/assets/queries/terraform/azure/app_service_http2_disabled/query.rego
@@ -3,58 +3,60 @@ package Cx
 import data.generic.common as common_lib
 import data.generic.terraform as tf_lib
 
+resources := {"azurerm_app_service", "azurerm_linux_web_app", "azurerm_windows_web_app"}
+
 CxPolicy[result] {
-	app := input.document[i].resource.azurerm_app_service[name]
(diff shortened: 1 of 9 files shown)


cx-artur-ribeiro (approved):
LGTM, nice job adding the new test cases!

PR #7595, opened 2025-07-28
fix(query): fix cloud formation queries producing duplicate similarityIDs

cx-romeu-silva:
- Some Cloud Formation queries were producing the same `similarityID` despite representing different issues.
- Several queries were missing `searchLine`, making it harder to precisely locate vulnerabilities.

**Proposed Changes**
- Updated multiple Cloud Formation queries by adding or modifying `searchValue`, and/or `searchLine` to ensure each detected vulnerability generates a unique `similarityID`.
  - AWS
    - API Gateway Deployment Without API Gateway UsagePlan Associated: Added Search Line. The second policy was always true whenever the first one was.
    - API Gateway Stage Without API Gateway UsagePlan Associated: Added Search Line. The second policy was always true whenever the first one was.
    - API Gateway V2 Stage Access Logging Settings Not Defined: Added Search Line & Added Search Value.
    - CloudTrail Not Integrated With CloudWatch: Added Search Line & Added Search Value.
    - EC2 Sensitive Port Is Publicly Exposed: Added Search Line & Modified Search Value. The query did not function correctly when IpProtocol was set to "-1", and it also attempted to handle an invalid case where the value was "all".
    - ECS Task Definition Invalid CPU or Memory: Added Search Line. The query was not distinguishing between `ContainerDefinitions`.
    - ELB Without Secure Protocol: Added Search Value. The existing query would work without changes, but I added another policy for when a protocol doesn't exist, so the `searchLine` works.
    - GameLift Fleet EC2 InboundPermissions With Port Range: Added Search Line.
    - HTTP Port Open To Internet: Modified Search Line. Excluding the incorrect logic, the `similarityID` aspect appears to be correct. I've fixed the faulty logic - previously, it would almost always flag as long as the `IpProtocol` was "tcp" and open to the entire network.
    - Kinesis SSE Not Configured: Added Search Value.
    - Lambda Functions With Full Privileges: Modified Search Line.
    - S3 Bucket With Unsecured CORS Rule: Added Search Line. `searchLine` is pointing to the wrong location in the json example - it is meant to point to the start of the array.
    - Security Group Egress With All Protocols: Added Search Line.
    - Security Group Ingress Has CIDR Not Recommended: Added Search Line.
    - Security Group Ingress With All Protocols: Added Search Line.
    - Security Group Ingress With Port Range: Added Search Line.
    - Security Group Rule Without Description: Added Search Line.
    - Unrestricted Security Group Ingress: Added Search Line. It should have worked even before the missing `searchLines` were added.
    - Vulnerable Default SSL Certificate: Added Search Line & Added Search Value.


cx-miguel-silva (commented):
Queries changes LGTM
Nice Work on the Resource checking mechanism using walks.

PR #7597, opened 2025-07-28
fix(query): fix azure resource manager queries producing duplicate similarityIDs

cx-romeu-silva:
- Some Azure Resource Manager queries were producing the same `similarityID` despite representing different issues.
- Several queries were missing `searchLine`, making it harder to precisely locate vulnerabilities.

**Proposed Changes**
- Updated multiple Azure Resource Manager queries by adding or modifying `searchValue`, and/or `searchLine` to ensure each detected vulnerability generates a unique `similarityID`.
  - Storage Logging For Read Write And Delete Requests Disabled: Added Search Value.
  - Website with 'Http20Enabled' Disabled: Modified Search Line


cx-artur-ribeiro (approved):
LGTM. Nice removal of those test cases, we should have done the addition of searchKeys and searchLines earlier, glad you are doing this job now.

PR #7602, opened 2025-07-29
feat(query): new query - S3_Bucket_Notifications_Disabled for terraform/aws

cx-andre-pereira:
- Currently there is no query to ensure S3 bucket notifications are not disabled in terraform/aws. 
- For any terraform sample with a ```aws_sns_topic``` ,```aws_sqs_queue```  or ```aws_lambda_function```  this query should flag those resources in case they are not associated with a [```aws_s3_bucket_notification```](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_notification).

**Proposed Changes**
- I have implemented the query for the case of not having a ```aws_s3_bucket_notification```  resource defined, in which case any resource that is one of the 3 the relevant ```types``` , as defined in the query logic and previously mentioned, will be flagged.
- Moreover the resources might be flagged in case the  ```aws_s3_bucket_notification```  resource exists but does not reference the resource(s) being scanned. Each one of the ```types```  must be included in specific fields and so the ```is_valid_reference```  function is used to guarantee that ,be it an array or a single object, the resource being scanned is properly referenced on the ```aws_s3_bucket_notification``` resource.

- Example "**aws_s3_bucket_notification**" resource with each of the 3 possible referencing methods: 
```   
resource "aws_s3_bucket_notification" "bucket_notification" {
   ...
   topic { topic_arn = aws_sns_topic.topic_name.arn ...} 
   queue { queue_arn = aws_sqs_queue.queue_name.arn ...}
   lambda_function { lambda_function_arn = aws_lambda_function.func_name.arn ...}
}
``` 


cx-eduardo-semanas (approved):
LGTM

PR #7603, opened 2025-07-29
fix(query): FN for S3 Bucket Allows Public Policy

cx-ricardo-jesus:
Closes #

- A different approach to this query was requested, in order to flag any insecure value at bucket and account levels.
- It was also asked to change the query severity to Medium.

**Proposed Changes**
- In this bug , I changed the query deeply in order to fulfill all the query logic required. 
- In this query, I made two policies, the first one checks the account-level configurations and the second one checks the bucket level configurations.
- For each one of the policies I used different helper functions, in order to make it easier to understand the code and to have a clear separation between the bucket and the account level.
- For each one of this helper functions, I handled the same cases for the two levels. In each one, first of all, I checked the cases when the field `block_public_policy` is defined but not defined to `true` and also to the cases when the field is not defined at all (is a vulnerability because the default value is `false`).
- On the two policies I used the value returned on the `res` variable to deliver the response.
- The only diference is on the second policy, that as a extra verification (`not is_block_public_policy_defined_to_true_at_account_level(resources)`) that only checks if there is a resource `aws_s3_account_public_access_block` defined explicitly to `true` in which case, according to the information provided on the research linked to the bug's card on Jira, should not return a positive result even when, at the bucket level, the field `aws_s3_bucket_public_access_block.block_public_policy` is not defined to `true` explicitly.
- One more difference it that now, if the sample has insecure configurations at the account-level and at the bucket-level it return's a vulnerability for both.
- I added all the positive samples provided on the research page and the negative sample that is also provided.
- Changed the file `negative2.tf` file in order to have a correct negative result as it should, because with the new implementation it would detect a vulnerability.
- Reduced the severity from `High` to `Medium` as it was asked to do.
- Changed the query description with the text provided on the research.

### DIFF
--- a/assets/queries/terraform/aws/s3_bucket_with_public_policy/metadata.json
+++ b/assets/queries/terraform/aws/s3_bucket_with_public_policy/metadata.json
@@ -1,9 +1,9 @@
 {
   "id": "1a4bc881-9f69-4d44-8c9a-d37d08f54c50",
   "queryName": "S3 Bucket Allows Public Policy",
-  "severity": "HIGH",
+  "severity": "MEDIUM",
   "category": "Access Control",
-  "descriptionText": "S3 bucket allows public policy",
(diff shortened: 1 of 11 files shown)


cx-eduardo-semanas (dismissed):
LGTM

cx-eduardo-semanas (approved):
LGTM

PR #7606, opened 2025-07-30
feat(log): add results and queries summaries to log

cx-laura-rodrigues:
- Adding key details about queries results to the log

**Proposed Changes**
- Added Queries Summary to log printer
- Added the existing Results Summary to the log

### DIFF
--- a/pkg/printer/printer.go
+++ b/pkg/printer/printer.go
@@ -1,11 +1,13 @@
 package printer
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"sort"
 	"strconv"
 	"strings"
+	"text/tabwriter"
 
 	"github.com/Checkmarx/kics/v2/pkg/utils"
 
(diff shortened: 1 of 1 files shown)


pkg/printer/printer.go
```diff
@@ -156,30 +158,82 @@ func PrintResult(summary *model.Summary, printer *Printer, usingCustomQueries bo
...
-	fmt.Printf("\nResults Summary:\n")
-	printSeverityCounter(model.SeverityCritical, summary.SeverityCounters[model.SeverityCritical], printer.Critical)
-	printSeverityCounter(model.SeverityHigh, summary.SeverityCounters[model.SeverityHigh], printer.High)
-	printSeverityCounter(model.SeverityMedium, summary.SeverityCounters[model.SeverityMedium], printer.Medium)
-	printSeverityCounter(model.SeverityLow, summary.SeverityCounters[model.SeverityLow], printer.Low)
-	printSeverityCounter(model.SeverityInfo, summary.SeverityCounters[model.SeverityInfo], printer.Info)
-	fmt.Printf("TOTAL: %d\n\n", summary.TotalCounter)
-
+	if !consoleFlags.GetBoolFlag(consoleFlags.VerboseFlag) {
+		fmt.Printf("\nResults Summary:\n")
```
cx-artur-ribeiro: Since we don't pass any variable to the fmt.Printf function, we could use Print instead. 

I know we had this code as is, but if you could change the function I would be grateful!

```suggestion
		fmt.Print("\nResults Summary:\n")
```

internal/console/flags/flags.go
```diff
@@ -184,6 +184,15 @@ func SetStrFlag(flagName, value string) {
 	}
 }
 
+// SetBoolFlag set a string flag using its name
```
cx-artur-ribeiro: Just a quick fix.

```suggestion
// SetBoolFlag set a boolean flag using its name
```

We could use the function `SetStrFlag` and pass the boolean value as a string (p.e "false" or "true"). In my opinion, I prefer the addition of the function since we have a getter for each type (integer, string and boolean), we could also have a setter for each type if needed. In this case, since we need to set a boolean flag, I'm good with this addition.
I'll wait for others to give their opinion too.

pkg/printer/printer.go
```diff
@@ -156,30 +158,82 @@ func PrintResult(summary *model.Summary, printer *Printer, usingCustomQueries bo
...
+	for idx := range severities {
+		log.Info().Msgf("%s: %d", severities[idx], summary.SeverityCounters[severities[idx]])
+	}
+	log.Info().Msgf("TOTAL: %d\n", summary.TotalCounter)
+}
+
+// Prints the formated Queries summary table
+func prettyPrintResultsTable(summary *model.Summary) error {
+	var buf bytes.Buffer
+	const SpacesInTab = 4
```
cx-artur-ribeiro: Could you move the constant defined on line 214 to the line 25 please? 
That way we can have all constants defined on the same place. Just a quality of life change.

cx-artur-ribeiro (changes requested):
Hey @cx-laura-rodrigues, 
First of all, thanks for the (well-needed) improvements to KICS logs!

I left three quality-of-life suggestions, but from my side, that’s it!
I’ll contact the other maintainers to get their opinions on these additions, especially the `SetBoolFlag` function.

Well done adding test coverage to all the code 😃

pkg/printer/printer.go
```diff
@@ -20,6 +22,7 @@ import (
 const (
 	charsLimitPerLine = 255
 	common            = "common"
+	SpacesInTab       = 4
```
cx-rui-araujo: Please use lowercase for the first letter, like `spacesInTab`. Uppercase is used for exported variables that are accessed from other packages.

pkg/printer/printer_test.go
```diff
@@ -173,6 +185,42 @@ func TestPrintResult(t *testing.T) {
...
+		Short: "Mock cmd",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return nil
+		},
+	}
+
+	data, err := os.ReadFile(filepath.FromSlash("../../internal/console/assets/kics-flags.json"))
+	require.NoError(t, err)
+	flags.InitJSONFlags(mockCmd, string(data), false, []string{"terraform"}, []string{"aws"})
+	flags.SetBoolFlag(flags.VerboseFlag, true)
```
cx-rui-araujo, line 200: Can you use SetStrFlag(flags.VerboseFlag, "true")? Or it does not work?

cx-rui-araujo (changes requested):
@cx-laura-rodrigues please check my comments

pkg/printer/printer_test.go
cx-laura-rodrigues: The verbose flag is defined as a boolean

cx-rui-araujo: Okkapa 🫡

cx-artur-ribeiro (approved):
Thanks for the work done Laura, LGTM!

PR #7607, opened 2025-07-30
feat(query): new query - Secretmanager Secret Without KMS for CloudFormation/aws

cx-ricardo-jesus:
- Currently we have no support in CloudFormation when there's a "Secretmanager Secret Without KMS", which was already implemented for the Terraform platform.

**Proposed Changes**
- For this new query I used a similar approach as used on the query for the Terraform platform.
- Created just one policy that uses a simple helper function called `is_kms_key_id_not_defined` which just checks if the `KmsKeyId` is not defined or is defined with an empty string.
- Added four positive samples, with cover's the cases when the `KmsKeyId` field is not defined or when it's defined with an empty string. 
- Added two negative samples that have both the `KmsKeyId` field defined.

### DIFF
--- a/assets/queries/terraform/aws/s3_bucket_with_public_policy/test/positive10.tf
+++ b/assets/queries/terraform/aws/s3_bucket_with_public_policy/test/positive10.tf
@@ -1,0 +1,2 @@
+resource "aws_s3_account_public_access_block" "allow_public_acc" {
+}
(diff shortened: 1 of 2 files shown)


cx-eduardo-semanas (approved):
LGTM

cx-artur-ribeiro (approved):
LGTM, nice searchLine addition to the query.

PR #7611, opened 2025-07-30
fix(query): fix google deployment manager queries producing duplicate similarityIDs

cx-romeu-silva:
- Some Google Deployment Manager queries were producing the same `similarityID` despite representing different issues.

**Proposed Changes**
- Updated multiple Google Deployment Manager queries by adding or modifying `searchValue`, and/or `searchLine` to ensure each detected vulnerability generates a unique `similarityID`.
  - GCP
    - Cloud Storage Anonymous or Publicly Accessible: Added Search Value.
    - Network Policy Disabled: Added Search Value.
    - Shielded VM Disabled: Added Search Value.


cx-miguel-silva (approved):
LGTM

PR #7612, opened 2025-07-30
fix(query): fp for Storage Share File Allows All ACL Permissions

cx-andre-pereira:
- Currently the query references the [```azurerm_storage_share_file```](https://registry.terraform.io/providers/hashicorp/azurerm/4.37.0/docs/resources/storage_share_file) resource which does not, itself, declare the ACL permissions associated with the query´s naming/purpose. 
- The query, as is currently, finds the ```storage_share_id``` within an ```azurerm_storage_share_file``` , and validates wether the [```azurerm_storage_share```](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_share) resource associated with that id is declaring ```permissions = "rwdl"```. 
- The query naming is deceiving because it seems as though the ```azurerm_storage_share_file``` resource is the one with dangerously permissive permissions field but it is actually the ```azurerm_storage_share``` that defines permissions for ```azurerm_storage_share_file``` resources, and the query should reference the resource that directly "Allows All ACL Permissions" as stated in the query´s name.

**Proposed Changes**
- I have adjusted the query name and the related files (metadata/expected_result/folder) to correctly point out the resource at fault , the ```azurerm_storage_share```.
- The query logic was already correctly flagging this field(```searchKey```,```keyExpectedValue```,...)making this change even more warranted .


assets/queries/terraform/azure/storage_share_allows_all_acl_permissions/metadata.json
```diff
@@ -0,0 +1,12 @@
+{
+  "id": "48bbe0fd-57e4-4678-a4a1-119e79c90fc3",
```
cx-artur-ribeiro: We should change the id as this is a new query in practice. Please follow the documentation, generating a new id and updating the descriptionID field accordingly (https://docs.kics.io/latest/creating-queries/#guidelines).

cx-artur-ribeiro (changes requested):
After reviewing the changes, it seems like we were already looking at the desired `azurerm_storage_share` resource and checking the `permissions` parameter.
Is it obligated to use the name of the resource `azurerm_storage_share_file` in the resource `azurerm_storage_share` in order for it to be used? I didn't find the documentation stating that but I understand that the query was already done this way so you didn't introduce this change.
If you could please make the changes requested I would approve the changes.

assets/queries/terraform/azure/storage_share_allows_all_acl_permissions/metadata.json
cx-andre-pereira: Uid updated in https://github.com/Checkmarx/kics/pull/7612/commits/001bde0a77ba1bd4222159a970a1a2b327fd36e5

cx-artur-ribeiro (approved):

PR #7614, opened 2025-07-30
feat(query): implements Neptune_Logging_Is_Disabled--cloudformation/aws

cx-andre-pereira:
- Currently there is no query equivalent of "[Neptune Logging Is Disabled](https://docs.kics.io/latest/queries/terraform-queries/aws/45cff7b6-3b80-40c1-ba7b-2cf480678bb8/)" for CloudFormation. 
- This query will implement that functionality by ensuring that resources of type  [```AWS::Neptune::DBCluster```](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-resource-neptune-dbcluster.html) contain an [```EnableCloudwatchLogsExports```](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-resource-neptune-dbcluster.html#cfn-neptune-dbcluster-enablecloudwatchlogsexports) field that include either one, or both, of the [valid keywords](slowquery) that ensure logging is enabled, those being : ```"audit"``` and ```"slowquery"``` .

**Proposed Changes**
- Implemented query logic with 2 distinct CxPolicy cases: 
     - First to account for lack of or null value of the ```EnableCloudwatchLogsExports``` field.
     - Second to flag in case the ```EnableCloudwatchLogsExports``` field is defined without any of the 2 valid keywords included within it. 

Note: After an [initial review](https://github.com/Checkmarx/kics/pull/7614#pullrequestreview-3100264534), the possibility of improving precision on [positive5 test](https://github.com/Checkmarx/kics/pull/7614/files#diff-76a54a905f8ded38e10296656b02b5ce6a2493c7a475ebf0fb4a5f561fa82ab7)'s result was raised. A new [commit](https://github.com/Checkmarx/kics/pull/7614/commits/76d803b7fd14013f060b4e1da523b380af7f515a) was pushed that adds a ```is_null_or_missing_LogsExports``` function to allow for precise results by pointing to the ```EnableCloudwatchLogsExports``` field even when the field is null; previously this would cause the results to point to the "Properties" parent field instead making it harder to pinpoint the cause for the result.


cx-eduardo-semanas (approved):
LGTM

cx-eduardo-semanas (approved):
LGTM

PR #7616, opened 2025-07-31
feat(query): new query - "EKS Cluster Encryption Disabled" query implemented for CloudFormation platform

cx-ricardo-jesus:
- Currently we only have the query `EKS Cluster Encryption Disabled` query implemented for Terraform platform and it should have support for CloudFormation.

**Proposed Changes**
- In this query I used a similar approach as used on the query for the Terraform platform.
- Basically I only created two policies.
- The first policy checks if the `EncryptionConfig` field is defined, if not, it returns a positive result.
- The second policy is used on the cases when the `EncryptionConfig` field is defined and the `Resources` array does not have the value "secrets" within. In that case it returns a positive result saying that it should have 'secrets' defined on `Resources`.
- Also added a negative test, that has with the `EncryptionConfig` field defined and, on the `Resources` array, it has the secrets value inside. This test results in two samples, one in json format and the other on yaml format.
- Added two positive tests, with four samples, two of them in json format and the other two in yaml format. The first one, does not have the field `EncryptionConfig` defined and the other one does not have "secrets" defined on `Resources` inside the `EncryptionConfig` field.


cx-artur-ribeiro (dismissed):
LGTM!

PR #7620, opened 2025-07-31
feat(query): lambda function without dead letter queue query implemented for Terraform/aws

cx-ricardo-jesus:
- Currently we have no support in Terraform when there's a "Lambda Function Without Dead Letter Queue", which we currently have for the CloudFormation platform. 

**Proposed Changes**
- To develop this query for Terraform platform, I used a similar approach as the one used for the CloudFormation query. 
- This query only has one policy, that uses a helper function called `is_dead_letter_queue_undefined_or_empty` used to check if the resource `aws_lambda_function` does not have the field `dead_letter_config` field defined or when it's defined but the field `target_arn` inside it, has a empty value(empty string).
- Comparing with the CloudFormation query implementation, this one does not return a positive result is the `target_arn` field is defined but with nothing within, not even a empty string, because having this field defined with nothing is not valid for this resource on Terraform since it is a field that is required when the field `dead_letter_config` is defined. 
- Instead of that, I made a verification that check's the cases when the field `target_arn` is defined with a empty string.
- Added one negative sample that has the field `dead_letter_config.target_arn` correctly defined.
- Added two positive samples, one of them returns a positive result because the field `dead_letter_config` field is not defined and the other one returns a positive result because the field `dead_letter_config.target_arn` is defined with a empty string.


cx-eduardo-semanas (approved):
LGTM

PR #7624, opened 2025-08-04
feat(queries): new queries "Instance Uses Metadata Service IMDSv1" for Terraform/aws, Ansible/aws and CloudFormation/AWS

cx-ricardo-jesus:
Closes #

- Currently we have no support for the verification of the IMDS version in place.

**Proposed Changes**
- I followed a similar approach regarding the implementation of the query for the platforms Ansible and Terraform. On both of them, I only used a single policy.
- In this policy is used two helper functions, the first one called `is_metadata_service_enabled` check's if the metadata service is enabled, i.e., when the field `http_endpoint` is set to `enabled` or it's not defined at all, which is the same as having the value `enabled` set (default's to `enabled`).
- The verification described above is necessary because if the metadata service is not enabled there is no point in checking the `http_tokens` field.
- After using the first helper function, it's used the other one, that is called `http_tokens_undefined_or_not_required` that as the name says, checks if the field `http_tokens` is not defined or is defined as `required`. The field `http_tokens` defines if whether or not the metadata service requires IMDSv2.
- On the implementation for the CloudFormation platform, I made the verifications on the field's for two types of resources `AWS::EC2::Instance` and `AWS:EC2::LaunchTemplate`.
- I made the same verifications as the queries implemented for the Ansible and Terraform platforms, but for this case the `MetadataOptions` block is inside the `LaunchTemplateData` for the resource of type `AWS::EC2::LaunchTemplate`.
- For all the queries I provided seven different samples, two negative samples and five positive samples.
- The negative samples, the first one hast the `metadata_options` block defined with both `http_endpoint` and `http_tokens` field's defined with `enabled` and `required` repectively.
- The positive samples, have all the variations regarding the values that should return a vulnerable result on `http_tokens` and `http_endpoint` fields.


assets/queries/ansible/aws/instance_uses_metadata_service_IMDSv1/query.rego
```diff
@@ -0,0 +1,84 @@
...
+    doc := input.document[i]
+	resource := doc.playbooks[j][modules[type]]
+	name := doc.playbooks[j].name
+
+    is_metadata_service_enabled(resource)
+
+    res := http_tokens_undefined(resource, type, j, name)
+	
+    result := {
+		"[DEBUG] resource value": resource,
```
cx-artur-ribeiro: I believe you wanted to remove the "DEBUG" statements but thats a good initiative to test the code 😄.
Please remove both statements.

assets/queries/ansible/aws/instance_uses_metadata_service_IMDSv1/query.rego
```diff
@@ -0,0 +1,84 @@
...
+    common_lib.valid_key(resource, "metadata_options")
+	common_lib.valid_key(resource.metadata_options, "http_tokens")
+	not resource.metadata_options.http_tokens == "required"
+
+    result := {
+        "documentId": input.document[i].id,
+        "resourceType": modules[type],
+        "resourceName": name,
+        "searchKey": sprintf("name={{%s}}.{{%s}}.metadata_options.http_tokens", [name, modules[type]]),
+        "searchLine": common_lib.build_search_line(["playbooks", j, modules[type], "matadata_options", "http_tokens"], []),
```
cx-artur-ribeiro: Although I find it very funny to have a "matadata", you should replace with 😄 
```suggestion
        "searchLine": common_lib.build_search_line(["playbooks", j, modules[type], "metadata_options", "http_tokens"], []),
```

cx-artur-ribeiro (approved):

PR #7626, opened 2025-08-04
fix(query): fp for SQL db instance publicly accessible -- terraform/gcp

cx-andre-pereira:
- This [query](https://docs.kics.io/latest/queries/terraform-queries/gcp/b187edca-b81e-4fdc-aff4-aab57db45edb/) checks that a [```google_sql_database_instance```](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance) terraform resource is not publicly accessible. A resource is considered publicly accessible, and therefore flagged, in any of the following scenarios:
  - The [```authorized_networks```](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance?=settings.ip_configuration#expiration_time-1) field contains the ```0.0.0.0``` address.
  - The ```authorized_networks``` field is undefined and [```ipv4_enabled```](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance?=settings.ip_configuration#ipv4_enabled-1) is set to true
  - The ```authorized_networks``` field is undefined, ```ipv4_enabled``` is set to false and the [```private_network```](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance?=settings.ip_configuration#private_network-1) field is undefined. 
  - The resource does not contain the [```ip_configuration```](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance?=settings.ip_configuration#ipv4_enabled-1) field (where all relevant fields are defined)

 - To ensure this behavior the query must be able to correctly iterate the payload and find the ```ipv4_enabled```,```private_network``` and ```authorized_networks``` values. Currently if this values are defined inside a "[dynamic](https://developer.hashicorp.com/terraform/language/expressions/dynamic-blocks)" ```ip_configuration``` block of the database, they do not get fetched properly and will cause false positives immediately by the forth scenario of the query since the ```ip_configuration``` field is now declared as:
```
dynamic "ip_configuration" {
   for_each = ...
    content {
      ipv4_enabled = bool
      private_network = "value"

    dynamic "authorized_networks" {
          for_each = ...
           content {
              name  = "some_trusted_network"
              value = "some_trusted_network_address"
            }
   }    
}
```
- The query logic must be updates to account for "dynamic" structures that could take place of the fields currently expected.

**Proposed Changes**
- Adjusted the query logic with new helper functions:
  - ```get_ip_configuration``` - This function gets called at the start of every CxPolicy and returns the fetched "ip_configuration" field, allowing it to be dynamic or not and get properly fetched regardless, and a "searchLine" value that will be used for the final result. This value is currently missing from the query and has to state dynamic/content blocks explicitly, the "searchKey" does not. 
  - ```no_authorized_networks``` - This is the policy that checks that there are no "authorized_networks" defined.
  - ```account_for_dynamic```- In case of a dynamic "authorized_networks" this function returns the "ip_configuration.dynamic" field to allow case handling. (path := ```ip_configuration.content.dynamic.authorized_networks```)
  - ```dynamic_contains``` - The biggest, final function is here to determine the full searchLine value in case of an "authorized_networks" field that contains the ```0.0.0.0``` address. It can be an array/object and dynamic/non-dynamic, and so must account for 4 separate possible combinations.


cx-andre-pereira:
This implementation is no longer relevant, the support for "dynamic" blocks is supposed to be arranged by the engine so static solutions of this sort do not make sense.

PR #7627, opened 2025-08-05
feat(query): elasticsearch domain encryption should be enabled node to node query implementation for CloudFormation/AWS

cx-ricardo-jesus:
- Currently, we have no support in CloudFormation for `Elasticsearch Domain Not Encrypted Node To Node`. This query was already implemented for Terraform.

**Proposed Changes**
- In this query, we have to cover two types of resources called `AWS::Elasticsearch::Domain` and `AWS::OpenSearchService::Domain`. For the second resource type, it is only valid to do the verifications if the field `EngineVersion` matches the pattern `^Elasticsearch_[0-9]{1}\\.[0-9]{1,2}$` in order to stay on the scope of the query.
- On the implementation of this query, I used two policies.
- The first policy check's for both of the resource types, if they do not have the field `NodeToNodeEncryptionOptions` defined, using a helper function called `node_to_node_block_not_defined`.
- The second policy checks the cases when the field `NodeToNodeEncryptionOptions.Enabled` is not defined to `true`, using an helper function called `node_to_node_encryption_not_enabled`.
- Provided eight positive samples that provides all the variations regarding the fields that return's the vulnerabilities that this query tackles.
- Provided four negative samples. All these samples have the field's with non vulnerable values for both of the resource types mentioned above.

### DIFF
--- a/assets/queries/cloudFormation/aws/elasticsearch_domain_not_encrypted_node_to_node/metadata.json
+++ b/assets/queries/cloudFormation/aws/elasticsearch_domain_not_encrypted_node_to_node/metadata.json
@@ -1,0 +1,12 @@
+{
+  "id": "43ed6fe0-edb6-43c2-97be-6501cf563d53",
+  "queryName": "Elasticsearch Domain Not Encrypted Node To Node",
+  "severity": "MEDIUM",
+  "category": "Encryption",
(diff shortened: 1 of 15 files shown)


cx-eduardo-semanas (approved):
LGTM

PR #7628, opened 2025-08-05
fix(query): web app not using TLS last version query requires minimum TLS version 1.3

cx-ricardo-jesus:
Closes #

- Currently the query only check's if the field `minTlsVersion` is set to '1.3' and it should also support the version 1.2.

**Proposed Changes**
- In order to fix this, I only made a really small change on the implementation on the query itself. Added a verification on the helper function `is_last_tls` in order to also check if the field is set to "1.2".
- Changed the results keyExpectedValue and keyActualValue fields in order to also say that is should be defined to "1.2" or "1.3".
-  Provided a extra negative sample that has the field `minTlsVersion` defined to 1.2.
- Changed two positive tests called `positive5.bicep` and `positive5.json` that had the field `minTlsVersion` defined to "1.2" and in order to make this positive samples valid, I changed the version to a valid one and that should also return a positive result.


cx-artur-ribeiro (approved):
LGTM

PR #7637, opened 2025-08-08
fix(bicep): remove references to Bicep as a platform

cx-artur-ribeiro:
- Bicep was incorrectly classified as a platform instead of an Azure Resource Manager file type;
- This caused misleading behavior in platform flags and documentation, which has now been corrected.

**Proposed Changes**
- Update documentation, removing any reference to Bicep as a Platform;
- Remove unused bicep platform related code;
- Update Test cases accordingly (e2e and unit tests);
- Fix flags documentation, removing spaces;
- Add e2e to test bicep commands failure;


cx-bruno-silva (approved):
LGTM

PR #7649, opened 2025-08-14
fix(queries): launch configuration is not encrypted resources missing support

cx-ricardo-jesus:
Closes #

- Currently, we do not have support for the "aws_instance" resource on Terraform/aws.
- Also, this query is not implemented for CloudFormation.

**Proposed Changes**
- Regarding the Terraform query changes, for the resources, I just created a variable called `supported_resources` that has the two resources that are supported in the query, and, only changed the line `resource := input.document[i].resource.aws_launch_configuration[name]` to `resource := input.document[i].resource[supported_resources[type]][name]` to do the verifications for both resources.
- Regarding the modules, I created a helper function for each one. 
- After an analysis of Terraform documentation regarding the modules equivalent to both resources, the only thing that changes between the two is the fact that the module equivalent to the `aws_instance` resource (`terraform-aws-modules/ec2-instance/aws`) does not have the input `ebs_block_device`.
- So, taking into account the information from the point above, the helper functions created for each one of the modules are similar, but the only thing that changes is the fact that we are passing the value  "aws_launch_configuration" and "aws_instance" to the `get_module_equivalent_key` function from the KICS common library depending on the type of module we are dealing with.
- I did not cover the cases when the block device is not defined because it is not in the scope of the query to detect a vulnerable result if the block device is not defined. We should return a positive result if the `encrypted` field is not defined(defaults to false) or it's defined to false.
- Regarding the CloudFormation query implementation, I created just one policy that uses a helper function called `prepare_issue` that is supposed to handle all the cases. 
- Using the same logic as the one used on the Terraform query implementation, I created a variable called `supported_resources` that contains the names of the resource types that are being analyzed on the query.
-  For each one of the resource types, the target field that is under analysis is on both resources, inside a `BlockDeviceMappings` array, and inside of that array within a field called `Ebs`, which has the field `Encrypted`. So, taking that into account, the helper function has two cases: 
    - when the `Encrypted` field is not defined (did not cover the case when Ebs is not defined, since if this block is not defined, it no longer falls within the scope of the query),
    - when the 'Encrypted' field is defined to false.
- Provided positive and negative samples that cover the cases that the query covers.


cx-artur-ribeiro (approved):
LGTM, nice work ricardo!

PR #7655, opened 2025-08-18
feat(queries): tags not copied to rds cluster snapshot query implementation for terraform/aws and CloudFormation/aws

cx-ricardo-jesus:
Closes #

- Currently, we do not have support for RDS Cluster tags not being copied to backup snapshots for Terraform and CloudFormation.

**Proposed Changes**
- For both platforms, I used the same approach, basically, I created two policies for both, one of them for the cases when the field `copy_tags_to_snapshot`\`CopyTagsToSnapshot` is defined to false, and the other policy for cases when the field is not defined at all. 
- For the CloudFormation, I covered the resources `AWS::RDS::DBInstance` and `AWS::RDS::DBCluster`. For Terraform I covered the resource `aws_rds_cluster`.
- Provided samples for all the cases and, for CloudFormation, I provided extra samples that cover the cases when the boolean parameters are strings.


cx-artur-ribeiro (approved):
LGTM

PR #7661, opened 2025-08-19
fix(query): fixed query "s3 bucket with public policy"

cx-ricardo-jesus:
Closes #

- The current query implementation does not support the following cases: 
- `aws_s3_account_public_access_block` defined with `block_public_policy` set to false and `aws_s3_bucket_public_access_block` defined with `block_public_policy` field set to true.
- `aws_s3_account_public_access_block` defined without `block_public_policy` and `aws_s3_bucket_public_access_block` defined with `block_public_policy` set to true.

- Regarding the samples, I added 3 samples, two of them covering the cases described above, and the other one covering the following case: 
- `aws_s3_account_public_access_block` not defined and `aws_s3_bucket_public_access_block` defined with `block_public_policy` field set to true.

**Proposed Changes**
- To fix the problems described above, I changed the query itself deeply.
- Now, the query only has one policy, which uses two helper functions.
- The first helper function called `prepare_account_level_status`, checks the security at the account level and returns the status of the resource `aws_s3_account_public_access_block`, which can be `unknown` (not defined), `secure`(defined with the field `block_public_account` set explicitly to true), `insecureFalse`, and `insecureNotDefined`(field `block_public_account` set to false and field `block_public_account` not defined).
- If the helper function `prepare_account_level_status` does not return "secure", it continues with the checks, but now at the bucket level, using the other helper function called `prepare_issues`. 
- In this second helper function called `prepare_issues`, if the `aws_s3_bucket_public_access_block` is not defined, it returns a vulnerable result that can change depending on the account level status.
- In the cases when the `aws_s3_bucket_public_access_block` resource is defined, it returns a positive result if the field `block_public_policy` is not defined or is set to false.
- With this new implementation, if an insecure configuration exists at the account level, but on the bucket level, the field `block_public_account` is defined to true, it does not return a vulnerable result saying that it's insecure at the account level, because in this specific case, it was requested not to return a result.


cx-artur-ribeiro (approved):
LGTM

PR #7664, opened 2025-08-20
test(query): add missing test case for S3 Bucket Allows Public Policy

cx-romeu-silva:
- The current query includes all test cases except the positive case where an `aws_s3_account_public_access_block` resource is defined, its `block_public_policy` field is missing, and no `aws_s3_bucket_public_access_block` resource is defined.

**Proposed Changes**
- Add the missing positive test case.


cx-artur-ribeiro (approved):
LGTM

PR #7667, opened 2025-08-20
feat(queries): query  IAM DB Cluster Auth Not Enabled implemented for terraform/aws and cloudFormation/aws

cx-ricardo-jesus:
Closes #

- Currently, we do not support situations where the field `iam_database_authentication_enabled`/`EnableIAMDatabaseAuthentication` is not defined or is set to false for CloudFormation and Terraform. 

**Proposed Changes**
- For both platforms, I used the same reasoning. Two policies, one for the cases when the target field is not defined, and the other for the cases when the target field is set to false.
- For both platforms, I used the same helper function called `valid_for_iam_engine_and_version_check_edited`, which is a modified version of the helper function `valid_for_iam_engine_and_version_check` from KICS common_lib.
- The helper function `valid_for_iam_engine_and_version_check_edited` checks the following points:
    - if the engine field is set to "mariadb" and the version is "10.6", "10.11", or "11.4". 
    - if the engines are `mysql`(or `mysql-postgresql`), `postgres`, `aurora-postgresql`, and `mariadb`, and there is no version defined.   
    - if the engines are `mysql`(or `mysql-postgresql`), `postgres`, and `aurora-postgresql`.
- These verifications made on the helper function were approved by the APPSEC team in the office.

- Regarding the original function present on the common_lib called `valid_for_iam_engine_and_version_check`, I made a change to also support the versions "10.11" and "11.4", which was discussed with the APPSEC team and got the approval from them too. 

- Added samples to cover all the possible scenarios.
- NOTE: on the query implementation for the Terraform platform, I did not add support for modules because I didn't find any module equivalent to the resource `aws_rds_cluster`.


assets/queries/terraform/aws/iam_db_cluster_auth_not_enabled/query.rego
```diff
@@ -0,0 +1,62 @@
...
+    valid_for_iam_engine_and_version_check_edited(resource, "engine", "engine_version")
+
+    result := {
+        "documentId": input.document[i].id,
+        "resourceType": "aws_rds_cluster",
+        "resourceName": tf_lib.get_resource_name(resource, name),
+        "searchKey": sprintf("aws_rds_cluster[%s].iam_database_authentication_enabled", [name]),
+        "issueType": "MissingAttribute",
+        "keyExpectedValue": "'iam_database_authentication_enabled' should be set to true",
+        "keyActualValue": "'iam_database_authentication_enabled' is undefined",
```
cx-artur-ribeiro: We should change the `keyActualValue` to something similar to the cloudFormation query. In this case, I believe it was a miss copy pasta. We want to tell the user that the field was incorrectly defined to false instead of "is undefined".

assets/queries/terraform/aws/iam_db_cluster_auth_not_enabled/query.rego
```diff
@@ -0,0 +1,62 @@
...
+}
+
+valid_for_iam_engine_and_version_check_edited(resource, engineVar, engineVersionVar) {
+    key_list := [engineVar, engineVersionVar]
+    contains(lower(resource[engineVar]), "mariadb")
+    supported_versions := {"10.6", "10.11", "11.4"}
+    version_check := {x | x := resource[key_list[_]]; contains(x, supported_versions[_])}
+    count(version_check) > 0
+} else {
+    engines_that_support_iam := ["aurora-postgresql", "postgres", "mysql", "mariadb"]
```
cx-artur-ribeiro, line 56: Since you added cases that cover the other engines that support iam, could you consider adding test cases for both terraform and cloudformation regarding "postgres" engine please?
I would suggest adding negative cases if you find them useful (otherwise the QA monsters will haunt you 👻)

cx-artur-ribeiro (changes requested):
Hey Ricardo, some quality of life changes and a typo 😄.
Please refer to my comments.

cx-artur-ribeiro (approved):
LGTM, thanks for the changes and the unit tests addition.

PR #7621, opened 2025-08-01
fix(query): fp for Media Type Object Without Schema -- OpenAPI/3.0

cx-andre-pereira:
Closes #

**Reason for Proposed Changes**
-

**Proposed Changes**
-
-
-

I submit this contribution under the Apache-2.0 license.

PR #7668, opened 2025-08-21
fix(query): media type object without schema -- OpenAPI 3.0 

cx-andre-pereira:
Closes #6682
- The fix for this query [#7621](https://github.com/Checkmarx/kics/pull/7621) resulted from an incorrect assessment of the query logic. 
- "The query, "[Media Type Object Without Schema](https://docs.kics.io/latest/queries/openapi-queries/f79b9d26-e945-44e7-98a1-b93f0f7a68a0/)", is meant to flag any [Media Type Object](https://swagger.io/docs/specification/v3_0/media-types/) if it does not have a schema field defined."  
- Given an excerpt such as this: 
```
requestBody:
      content:
        application/x-www-form-urlencoded:
          schema:
            type: object
            required:
              - password
            properties:
              password:
                type: string
                description: Credentials
                x-pii: {}
        text/plain: {}
```
- The query would go through all elements of ```content``` and properly flag if any of them did not contain the ```schema``` field; the issue was the fact that the "searchKey" value would then point to ```content``` rather than the "text/plain" that caused the flag to be raised making it seem at first glance that the "application/x-www-form-urlencoded" was unwarrantedly flagged.

**Proposed Changes**
- Added a "searchLine" value to ensure the query point to the specific object within the content field. 
- Adjusted "searchKey" value.
- Made the false negative tests added on the previous "fix" positive tests.


cx-artur-ribeiro (approved):
LGTM

PR #7670, opened 2025-08-21
fix(query): fixed fn for "SQL Server Database With Unrecommended Retention Days" query

cx-ricardo-jesus:
Closes #

- Currently, the query doesn't take into consideration the situations where the server/database does not have a defined retention period for the auditing settings (Microsoft.Sql/servers/auditingSettings.retentionInDays).

**Proposed Changes**
- To add support for the resource of type `Microsoft.Sql/servers/`, I added to the variable called `types` the name of its auditingSettings child resource(`Microsoft.Sql/servers/auditingSettings`).
- Also, created another variable called `dbTypes` which has the names of the resource types that the "father" resources have("Microsoft.Sql/servers/databases", "Microsoft.Sql/servers").
- This query, after the changes, still has two policies, one for the cases when the `retentionDays` field is not defined and the other one for the cases when it's defined but with a value under 90, which should also return a vulnerable result.
- On the first policy, using the walk function, I iterated through the resources and, if the resource has a type equal to the types of the array `dbTypes` (represents the father resources), it is extracted is child resources using the helper function `get_children` which is a modification from the function `get_children` from the KICS arm_lib, that serves the purpose of this query.
- So, the helper function basically returns an array that has the path to the child resource and the child resources itself.
- It has three cases. The first one is for the resources that have the child resources nested within, which happens sometimes in JSON format. The second one, its for samples that have the bicep format, that does not have the resources nested, that's why I added the ```not common_lib.valid_key(parent, "resources")``` and for the child resources it does not have the field `dependsOn`(it's a field that appears on JSON format samples). The third case is for the samples in JSON format that has the resources defined sequentially.
- After the helper function returns the values that it should return, it's checked for each child resource, if its type is one of the types present on the dbTypes array and, if it is and the resource is enabled, it checks if the field `retentionDays` is not defined.
-  For the second policy, I did the same but, the last check verifies if the field `retentionDays` is defined to a value under 90.
- Also, added samples for the `Microsoft.Sql/servers` resource type.


cx-artur-ribeiro (approved):
LGTM

PR #7680, opened 2025-08-28
fix(query): added support for a new case in "elasticsearch domain not encrypted" query

cx-ricardo-jesus:
Closes #

- The current implementation does not support the cases when the field `Enabled` is not defined inside the `NodeToNodeEncryptionOptions` block.

**Proposed Changes**
- Added two more cases on the helper function called `node_to_node_block_not_defined` for the case that was not taken into account. 
- Added samples for the case mentioned above

### DIFF
--- a/assets/queries/cloudFormation/aws/elasticsearch_domain_not_encrypted_node_to_node/query.rego
+++ b/assets/queries/cloudFormation/aws/elasticsearch_domain_not_encrypted_node_to_node/query.rego
@@ -1,0 +1,61 @@
+package Cx 
+
+import data.generic.cloudformation as cf_lib
+import data.generic.common as common_lib
+
+resources := {"AWS::Elasticsearch::Domain", "AWS::OpenSearchService::Domain"} 
+
+CxPolicy[result] {
+    document := input.document[i]
+    resource := document.Resources[name]
+    resource.Type == resources[m]
+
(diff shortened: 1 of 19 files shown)


assets/queries/cloudFormation/aws/elasticsearch_domain_not_encrypted_node_to_node/query.rego
```diff
@@ -51,11 +53,33 @@ node_to_node_encryption_not_enabled(resource, type) {
...
+} else = r {
     type == "AWS::OpenSearchService::Domain"
     regex.match("^Elasticsearch_[0-9]{1}\\.[0-9]{1,2}$", resource.EngineVersion)
     not common_lib.valid_key(resource, "NodeToNodeEncryptionOptions")
+    r := {
+        "sk": sprintf("Resources.%s.Properties", [name]),
+        "sl": common_lib.build_search_line(["Resources", name, "Properties"], []),
+    }
+} else = r {
+    type == "AWS::OpenSearchService::Domain"
```
cx-miguel-dasilva, line 79:
```suggestion
    type == "AWS::OpenSearchService::Domain"
    regex.match("^Elasticsearch_[0-9]{1}\\.[0-9]{1,2}$", resource.EngineVersion)
```

cx-miguel-silva (changes requested):

PR #7681, opened 2025-08-28
test(query): added two more samples to "App Service HTTP2 Disabled" query

cx-ricardo-jesus:
Closes #

- Currently we do not have samples that cover the cases when the `site_config` block is not defined for the resources `azurerm_linux_web_app` and `azurerm_windows_web_app`.

**Proposed Changes**
- Added the samples (positive8.tf and positive9.tf)

### DIFF
--- a/.github/workflows/go-ci.yml
+++ b/.github/workflows/go-ci.yml
@@ -40,7 +40,7 @@ jobs:
     strategy:
       matrix:
         go-version: [1.24.x]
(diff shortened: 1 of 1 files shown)


cx-eduardo-semanas (approved):
LGTM

cx-artur-ribeiro:
I've reverted windows runner image from windows-latest (2025) to windows-2022 due to disk space problems reported in the related issue https://github.com/actions/runner/issues/4012.

PR #7688, opened 2025-09-01
fix(queries): added suport modules/resources on instance uses metadata service IMDSv1 query for Terraform, CloudFormation and Ansible implementations

cx-ricardo-jesus:
Closes #

- The query implemented for the CloudFormation platform supported the resources `AWS::EC2::LaunchTemplate` and `AWS::EC2::Instance` resources, but there is no support for the `aws_launch_configuration` resource and the equivalent modules. For Ansible, it supported only the module `amazon.aws.ec2_instance` and was missing support for `community.aws.ec2_instance`, 
`community.aws.autoscaling_launch_config` and `community.aws.ec2_lc`.

**Proposed Changes**
- For the CloudFormation, it was only added test cases. 
- For the query made for Ansible, it was added the modules on the `modules` variable in order to make the same verification that already did for the module `amazon.aws.ec2_instance` but, for the other three new modules.
- For the Terraform query implementation, it was added support for modules equivalent to the resources `aws_instance` and `aws_launch_configuration`. 
- In this last query, it was added a new policy that does the necessary verification for the modules, using the helper function `is_metadata_service_enabled`, that checks if the `http_endpoint` field inside the `metadata_options` is set to `enabled` ir if the `metadata_options` block is not defined or when the `metadata_options` block is defined but the field `http_endpoint` inside is not.
- For all the queries, it was added some new test cases that cover some cases that were not covered before and that also cover the new resources/modules supported. 


cx-artur-ribeiro (approved):
LGTM, nice work!
Please make sure all cases are covered (I didn't find any case that wasn't)

PR #7694, opened 2025-09-03
fix(query): fn for API Gateway With CloudWatch Logging Disabled - terraform/aws

cx-andre-pereira:
- Currently the [API Gateway With CloudWatch Logging Disabled](https://docs.kics.io/latest/queries/terraform-queries/aws/982aa526-6970-4c59-8b9b-2ce7e019fe36/) terraform query does not account for the ```access_log_settings.destination_arn``` field of the [aws_api_gateway_stage](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/api_gateway_stage#managing-the-api-logging-cloudwatch-log-group) resource.

- This query is meant to flag whenever "CloudWatch Logs" are not enabled for a ```aws_api_gateway_stage``` resource; currently it only checks for a [```aws_cloudwatch_log_group```](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudwatch_log_group) resource that has a "name" field that references the relevant ```aws_api_gateway_stage```'s ```stage_name``` . To do this the [```haveLogs```](https://github.com/Checkmarx/kics/blob/master/assets/queries/terraform/aws/api_gateway_with_cloudwatch_logging_disabled/query.rego#L20) function checks for a specific regex match where given a ```stage_name``` like "example" the "name" of the ```aws_cloudwatch_log_group``` could be : 

  - ```name = "API-Gateway-Execution-Logs_${aws_api_gateway_rest_api.example.id}/${example}"``` 
 
- This is shown in the negative1 test with "var.stage_name" as the stage_name. The "name" also references an [```aws_api_gateway_rest_api```](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/api_gateway_rest_api) resource but that is irrelevant for the purposes of this query. 

- The missing case is associated with the ```access_log_settings.destination_arn``` field from the ```aws_api_gateway_stage``` since this field's purpose is to define where to send access logs as per the [documentation](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/api_gateway_stage#destination_arn-1). This is a glaring flaw and the field must be accounted for. 

**Proposed Changes**
- Fixed a [typo in the original message](https://github.com/Checkmarx/kics/blob/master/assets/queries/terraform/aws/api_gateway_with_cloudwatch_logging_disabled/query.rego#L16), and altered said message to be more informative (original message stated "not undefined" when it should say "undefined").

- Fixed a[ logic error](https://github.com/Checkmarx/kics/blob/master/assets/queries/terraform/aws/api_gateway_with_cloudwatch_logging_disabled/query.rego#L21) unrelated to the False Positive that caused the scan of a folder to not give results that scanning an individual file would, this was due to the fact that the auxiliary function was scanning through all "documents" in the generated payload. 

- Added new helper functions and a "master" helper function [```get_results```](https://github.com/Checkmarx/kics/pull/7694/files#diff-891c2682df8dc6d3d78794e5ebfc1cdb796ccac1165836dea15fbdca38ad0db7R25) that will define the flagging order. I decided that, in case a sample does not have a valid "stage_name" or a valid "destination_arn", it is the "destination_arn" that will be flagged since it is a resource that exists solely to properly reference a ```aws_cloudwatch_log_group```, while the stage_name will flag simply because no ```aws_cloudwatch_log_group``` resource properly referenced it and not because it is totally invalid.

- Added 2 different [```result```](https://github.com/Checkmarx/kics/pull/7694/files#diff-891c2682df8dc6d3d78794e5ebfc1cdb796ccac1165836dea15fbdca38ad0db7R13)values depending on the reason for the query flagging. This results are given by the [```does_not_have_valid_stage_name```](https://github.com/Checkmarx/kics/pull/7694/files#diff-891c2682df8dc6d3d78794e5ebfc1cdb796ccac1165836dea15fbdca38ad0db7R44) policy that is very much alike the original query logic and the new [```does_not_have_valid_destination_arn```](https://github.com/Checkmarx/kics/pull/7694/files#diff-891c2682df8dc6d3d78794e5ebfc1cdb796ccac1165836dea15fbdca38ad0db7R34) for the missing case. Note that the "destination_arn" must also respect a specific [regex pattern](https://github.com/Checkmarx/kics/pull/7694/files#diff-891c2682df8dc6d3d78794e5ebfc1cdb796ccac1165836dea15fbdca38ad0db7R67) as such : 
  - ```regexPattern := sprintf("aws_cloudwatch_log_group.%s.arn", [name])```
- Where the "name" is the literal name of the ```aws_cloudwatch_log_group``` not its "name" field.

- Added tests for the new case :
  - [negative3](https://github.com/Checkmarx/kics/pull/7694/files#diff-928f01a611840b47e175faea2f2f40b9ab62a4667de4bf3534efbb83e4e3c552) has a valid ```destination_arn```
  - [negative4](https://github.com/Checkmarx/kics/pull/7694/files#diff-4f3ea25dd69d47563c8b02c12acba86b4ed6869863d469215605f49c49952b19) has a null ```destination_arn``` but a valid ```stage_name```
  - [negative5](https://github.com/Checkmarx/kics/pull/7694/files#diff-906907dcd9e770d7f5aeceac08d4ea426441ee772fc9e22f2dfd61b566256c91) has a valid ```destination_arn``` and ```stage_name```
  - [positive2](https://github.com/Checkmarx/kics/pull/7694/files#diff-05c273673b16a2f78d593620989f1b84e75b43a142bb61e7b8c9bc69c505f9b0) has an invalid ```destination_arn``` and ```stage_name```
  - [positive3](https://github.com/Checkmarx/kics/pull/7694/files#diff-c324508ae1c033598e427242265d864dbfeb66e2754fe6f9c42599bd4bee8823) has a null ```destination_arn``` and ```stage_name```
  - [positive4](https://github.com/Checkmarx/kics/pull/7694/files#diff-9c329f38e8175a4f05c4bae9b63cb52f3d869cb48663da295507f778f9c64c85) has no ```access_log_settings``` field and an invalid ```stage_name```. 
  
 - A test case for the ```access_log_settings``` field without the ```destination_arn``` was not added since the ```destination_arn``` is a [required](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/api_gateway_stage#destination_arn-1) for said field.

Note : From my research modules cannot replace this resource since even using modules that can provide similar functions like [api-gateway](https://registry.terraform.io/modules/infrablocks/api-gateway/aws/2.1.0-rc.9), they cannot define the ```aws_api_gateway_stage``` to ```aws_cloudwatch_log_group``` interaction that causes this query to flag by themselves. 


assets/queries/terraform/aws/api_gateway_with_cloudwatch_logging_disabled/query.rego
```diff
@@ -1,25 +1,70 @@
...
+		"searchLine": common_lib.build_search_line(["resource","aws_api_gateway_stage", name, "access_log_settings", "destination_arn"],[])
+	}
+} else = ""
+
+does_not_have_valid_stage_name(resource,doc,name) = results {
+	not haveLogs_stageName(resource,doc)
+	results := {
+		"searchKey": sprintf("aws_api_gateway_stage[%s]", [name]),
+		"keyExpectedValue": sprintf("'aws_cloudwatch_log_group' for 'aws_api_gateway_stage[%s]' should be defined and use the correct naming convention",[name]),
+		"keyActualValue": sprintf("'aws_cloudwatch_log_group' for 'aws_api_gateway_stage[%s]' is undefined or not using the correct naming convention",[name]),
```
cx-artur-ribeiro: I would suggest adding the "is" to the second part of the sentence. 

```suggestion
		"keyActualValue": sprintf("'aws_cloudwatch_log_group' for 'aws_api_gateway_stage[%s]' is undefined or is not using the correct naming convention",[name]),
```

PR #7698, opened 2025-09-04
fix(query): added two new allowRules on "Generic Secret" and "Generic Token" queries from Passwords and Secrets

cx-ricardo-jesus:
Closes #

- Currently, it returns a false positive result on cases when there is a Bicep template, and a secret is retrieved from a keyVault module using the built-in function called `getSecret`.
- Also, it's detected a false positive on the following cases:
```validationToken: module_name.outputs.parameter_field```
- This last example returns a false positive when it's just pointing to a module output.

**Proposed Changes**
- For the first case, it was created a new AllowRule was created on the "Generic Secret" query that covers the cases when it's used the built-in function `getSecret`. The new allowRule has the following pattern:
```(?i)['\"]?secret[_]?(key|value)?['\"]?\\s*(:|=)\\s*[a-zA-Z]*\\.getSecret\\(\\s*[\"']?([A-Za-z0-9/~^_!@#&%(){};=?*+-<>,:;[\\]%$]*)[\"']?```
- For the second case, it was created a new allowRule on the query "Generic Token" that covers the specific cases when a parameter with a token in its name is just pointing to a module output. The pattern used on the new allowRule was based on the official outputs documentation from AzureResourceManager Bicep documentation (https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/outputs?tabs=azure-powershell#outputs-from-modules) that has a defined syntax: 
`<module-name>.outputs.<property-name>`
- The new allow rule covers the vast majority of the cases; however, it can cause a false negative, although the chances are almost nil, in cases where, for example, a token is defined on a YAML file(which can have strings defined without "") with the syntax mentioned above. 

### CODE REVIEW

File: assets/queries/common/passwords_and_secrets/regex_rules.json [hunk 1]
```diff
@@ -346,6 +350,10 @@
         {
           "description": "Avoiding next_token Var",
           "regex": "(?i)['\"]?next(_)?token['\"]?\\s*[:=]\\s*['\"]?([[A-Za-z0-9/~^_!@&%()=?*+-]+)['\"]?"
+        },
+        {
+          "description": "Avoiding reference to module outputs in Bicep",
```
cx-artur-ribeiro: Just a quick improvement, I would put "reference" as "references", like shown below: 

```suggestion
          "description": "Avoiding references to module outputs in Bicep",
```

PR #7709, opened 2025-09-09
test(query): new negative tests for Media Type Object Without Schema

cx-andre-pereira:
- The [original fix](https://github.com/Checkmarx/kics/pull/7621) and [re work](https://github.com/Checkmarx/kics/pull/7668) of said fix since it was incorrect, did not add negative tests for a "HEAD" type operation without a ```schema``` defined, these do not flag since the query logic excludes "HEAD" type operations, proving the intended behavior

**Proposed Changes**
- Added the missing test for "JSON" and "YAML" formats


cx-artur-ribeiro (approved):
LGTM

PR #7712, opened 2025-09-10
fix(query): added support for modules and added searchLine for lambda functions without letter queue for Terraform/aws

cx-ricardo-jesus:
Closes #

- The current query implementation does not have the searchLine present on the results and does not support modules

**Proposed Changes**
- Added support for modules, creating an extra policy that handles them.
- Added new test cases for the modules


cx-artur-ribeiro (approved):
LGTM

PR #7713, opened 2025-09-11
fix(query): passwords and secrets fp for run after triggers

cx-andre-pereira:
- Currently the [regex_rules](https://github.com/Checkmarx/kics/blob/master/assets/queries/common/passwords_and_secrets/regex_rules.json) from passwords and secrets are flagging the following fields : 
```
"HTTP_-_Get_OAuth_Token" : ...
"Parse_JSON_-_OAuth_Token": ...
"Try_-_Get_OAuth_Token": ...
"Catch_-_Get_OAuth_Token": ...
```
- These should be assumed as "[Microsoft.Logic workflows](https://learn.microsoft.com/en-us/azure/templates/microsoft.logic/workflows?pivots=deployment-language-arm-template#flowaccesscontrolconfigurationpolicy-1)" actions since they follow a specific [naming convention](https://blog.sandro-pereira.com/2022/02/27/logic-app-best-practices-tips-and-tricks-2-actions-naming-convention/) and can only have preset values ([7 possible](https://learn.microsoft.com/en-us/azure/logic-apps/view-workflow-status-run-history?tabs=consumption)) and not sensitive information. 
- These False Positives are happening when "actions" with "runAfter" triggers define said triggers as such: 
```
"runAfter_164": {
       "HTTP_-_Get_OAuth_Token": [
                                     "Succeeded"
                               ]
        }
```
or 
```
"runAfter_164": {
       "HTTP_-_Get_OAuth_Token": ["Succeeded"]
        }
```
- Given the limitations of the passwords_and_secrets's queries it is impossible to distinguish these cases in scenarios that rely on different naming schemes that also finish with "token". An action like "HTTP2_-_Get_OAuth_Token" would also be flagged even after this fix because it is impossible to predict how a user will name an action.

- Considering the first example it is also impossible to filter by value associated with the field like "Succeeded". Since the scan will be line by line it is likely that said value is not available; otherwise , knowing the 7 possible values, the filtering process could be applicable to many more samples and still safe enough to not cause new False Negatives.

**Proposed Changes**
- Added a new "[allowRules](https://github.com/Checkmarx/kics/blob/master/assets/queries/common/passwords_and_secrets/regex_rules.json#L293)" rule to the "[Generic Token](https://github.com/Checkmarx/kics/blob/master/assets/queries/common/passwords_and_secrets/regex_rules.json#L291)" query:
```
{
   "description": "Avoiding Run After Triggers",
   "regex": "(?i)['\"](HTTP|Parse_JSON|Try|Catch)_-_(Get_)?OAuth_Token['\"]?\\s*[:=]\\s*['\"]?([[A-Za-z0-9/~^_!@&%()=?*+-]+)['\"]?"
}
```
- Added a test "[negative56.json](https://github.com/Checkmarx/kics/pull/7713/files#diff-19306f4b4c36c0c1c3b5d2370067d46875f94b58dcbcd92a5a8c5f7685a6b0be)" with all 4 fields to be ignored.

- Finally, after https://github.com/Checkmarx/kics/pull/7698 was merged some negative test naming isues were apparent, there were 2 negative54 tests (from previous PRs) and 2 negative56 tests, these did not cause issues because their file types were different but the naming was now corrected so that there are 58 tests in total finishing with negative58.bicep.


assets/queries/common/passwords_and_secrets/regex_rules.json
```diff
@@ -346,6 +346,10 @@
         {
           "description": "Avoiding next_token Var",
           "regex": "(?i)['\"]?next(_)?token['\"]?\\s*[:=]\\s*['\"]?([[A-Za-z0-9/~^_!@&%()=?*+-]+)['\"]?"
+        },
+        {
+          "description": "Avoiding Run After Triggers",
```
cx-artur-ribeiro: Could you adjusted the description to reflect that this "only" avoids run after triggers covered by best practices, please

cx-andre-pereira: Done in https://github.com/Checkmarx/kics/pull/7713/commits/241f6f5881cf896d2d10bd9fb8aab569280e2335

PR #7726, opened 2025-09-17
test(query): fix for positive4 test to "Unknown Port Exposed To Internet" query

cx-andre-pereira:
- The positive4 test of the "Unknown Port Exposed To Internet" query had a "ipv6_cidr_blocks" field set to ["::/0"] inside an "ingress_with_cidr_blocks" of the "terraform-aws-modules/security-group/aws" module. 
- This was untended since the supported value is actually "cidr_blocks", the ipv4 version of the field. The mistake is not present in the near identical negative4 sample.

**Proposed Changes**
- Fixed the field naming and value.


cx-artur-ribeiro (approved):
LGTM

PR #7646, opened 2025-08-13
fix(query): fn for remote desktop port open to internet--terraform/aws

cx-andre-pereira:
Closes #

**Reason for Proposed Changes**
-

**Proposed Changes**
-
-
-

I submit this contribution under the Apache-2.0 license.

PR #7729, opened 2025-09-18
test(query): better testing for security group rules without description

cx-andre-pereira:
- The [original fix](https://github.com/Checkmarx/kics/pull/7646) for this query was missing a negative tests for an "aws_security_group" egress field. 
- Additionally the CxPolicy for modules was flagging as an "IncorrectValue" for the "issueType" when it should be "MissingAttribute".

**Proposed Changes**
- Added the missing test scenario 
-
-


cx-artur-ribeiro (approved):
LGTM!

PR #7731, opened 2025-09-18
test(query): test improvement for the "sensitive port" terraform queries

cx-andre-pereira:

- The testing for the 3 "sensitive port is exposed" queries are too big, the unnecessary number of samples and lack of logic behind the tests plus the way number of results becomes way higher than necessary given the current from/to port range makes them inadequate and obnoxious.

-  I decided to adjust these to not only be more minimalistic, but actually properly test all relevant scenarios; beforehand some scenarios were actually missing among all the bloat.

- To ensure proper testing we must keep in mind that for all 3 queries there are 3 conditions that must be meet for any given relevant resource to flag: 
  - The **protocol** must be set to "udp", "tcp" or "-1" (equivalent to all protocols at once)
  - The **port range** ("from_port"/"to_port") must contain a known port, all are specified by the "[tcpPortsMap](https://github.com/Checkmarx/kics/blob/b927ebf7a943e265bc5c9552e93efe888032067d/assets/libraries/common.rego#L183)"
  - **Cidr restriction**. This varies for each query since it is the one feature that distinguishes the three of them:
    -  "[Sensitive Port Is Exposed To Wide Private Network](https://github.com/Checkmarx/kics/blob/b927ebf7a943e265bc5c9552e93efe888032067d/assets/queries/terraform/aws/sensitive_port_is_exposed_to_wide_private_network/query.rego#L120)" - determined by the [isPrivateIP](https://github.com/Checkmarx/kics/blob/b927ebf7a943e265bc5c9552e93efe888032067d/assets/libraries/common.rego#L79) function
    -  "[Sensitive Port Is Exposed To Entire Network](https://github.com/Checkmarx/kics/blob/b927ebf7a943e265bc5c9552e93efe888032067d/assets/queries/terraform/aws/sensitive_port_is_exposed_to_entire_network/query.rego#L16)" - demands that the mask of the cidr is "/0" ([cidr_is_unmasked](https://github.com/Checkmarx/kics/blob/b927ebf7a943e265bc5c9552e93efe888032067d/assets/libraries/terraform.rego#L35))
    - "[Sensitive Port Is Exposed To Small Public Network](https://github.com/Checkmarx/kics/blob/b927ebf7a943e265bc5c9552e93efe888032067d/assets/queries/terraform/aws/sensitive_port_is_exposed_to_small_public_network/query.rego#L123)" - demands that ipv4 masks are within "/25" and "/29", and ipv6 within "/121" and "/125" 

**Proposed Changes**
- Implemented standard testing throughout the 3 queries only changing the "cidr" fields to adjust for each query's context.

- Changed the **negative** tests so that they test each mandatory condition for flagging individually, additionally a sample with all fields "incorrect" for flagging purposes was also included. Each sample has a comment to easily discern the reasoning for it being a negative.

- For the **positive** tests, in each of the queries all 3 protocols (-1/tcp/udp) and 2 of the “known” ports (22 and 110) are tested for. 

- The "Sensitive Port Is Exposed To Wide Private Network" has a sample for each of the 3 ipv4 private IP ranges. For ipv6 only the "fd00::/8" range was tested as a subtle nod to the fact that, as of writing, this is the only part of the private ipv6 range ("fc00::/7") that is in use; "fc00::/8" is unused but already included in our [private IP list](https://github.com/Checkmarx/kics/blob/b927ebf7a943e265bc5c9552e93efe888032067d/assets/libraries/common.rego#L80).

- For the 2 other queries, since the mask is the only relevant part of the IP addresses for their "cidr" checks, most addresses were kept identical to the ones used to test the private network query except for their mask values. 

Note that all tests include the ipv4 and ipv6 variants of the relevant fields.


cx-artur-ribeiro (approved):
LGTM

PR #7734, opened 2025-09-22
test(query): added new case for the several resources covered on the queries for terraform/aws

cx-ricardo-jesus:
Closes #

- The current test files do not cover the cases when an open IPv4 or IPv6 address is defined ("0.0.0.0/0" for IPv4 or the equivalent to IPv6).

**Proposed Changes**
- Added the cases on the test files (negative2.tf, negative3.tf and negative4.tf).


cx-artur-ribeiro (approved):
LGTM

PR #7765, opened 2025-10-12
fix(typeflag): fix bicep wrong behavior with type and excludeType flags

cx-artur-ribeiro:
- Since Bicep is a file extension associated with Azure Resource Manager (ARM) and not a separate platform, we should allow the use of the flags `-t, --type` or `--exclude-type` with azureResourceManager to include or exclude both ARM and Bicep files.

**Proposed Changes**
- Rename type name from bicep to arm;
- Add unit tests coverage to the intended behaviour.


pkg/analyzer/analyzer_test.go
```diff
@@ -436,6 +436,32 @@ func TestAnalyzer_Analyze(t *testing.T) {
...
+			excludeGitIgnore:     false,
+			MaxFileSize:          -1,
+		},
+		{
+			name:                 "analyze_test_bicep_with_exclude_type_flag",
+			paths:                []string{filepath.FromSlash("../../test/fixtures/bicep_test")},
+			wantTypes:            []string{},
+			wantExclude:          []string{},
+			typesFromFlag:        []string{""},
+			excludeTypesFromFlag: []string{"azureresourcemanager"},
```
cx-rui-araujo, line 458: Can you create tests with more than 1 platform? So, the test flow will enter in the if case that contains the `return utils.Contains`.

cx-artur-ribeiro: Done in https://github.com/Checkmarx/kics/pull/7765/commits/869f5857215b8b3e63245e5737de81fb88ed0b84

PR #7771, opened 2025-10-14
fix(queries): BETA queries naming removed for Tencent Cloud & Databricks

cx-miguel-dasilva:
**Proposed Changes**
- BETA queries naming removed for Tencent Cloud & Databricks
- fix assets transition duplication


cx-artur-ribeiro (approved):
LGTM!

PR #7773, opened 2025-10-16
feat(query): implements "Beta - Google DNS Policy Logging Disabled"

cx-andre-pereira:
- There is no query to check that a "[google_dns_policy](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/dns_policy)" resource has logging enabled through its "[enable_logging](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/dns_policy#enable_logging-1)" field.

**Proposed Changes**
- The query's logic is extremely simple, all it does is flag if the "[enable_logging](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/dns_policy#enable_logging-1)" field is missing/null or if it is set to a value that is not ```true```.

### DIFF
--- a/assets/similarityID_transition/terraform_gcp.yaml
+++ b/assets/similarityID_transition/terraform_gcp.yaml
@@ -1,0 +1,5 @@
+similarityIDChangeList:
+    - queryId: cc9e464e-5abc-4c8f-8077-a9aa7ebe6a05
+      queryName: Beta - Google DNS Policy Logging Disabled
+      observations: ""
+      change: 2


assets/queries/terraform/gcp/google_dns_policy logging disabled/query.rego
```diff
@@ -0,0 +1,47 @@
+package Cx
+
+import data.generic.common as common_lib
+import data.generic.terraform as tf_lib
+
+CxPolicy[result] {
+	doc := input.document[i]
+	dns := doc.resource["google_dns_policy"][name]
```
cx-artur-ribeiro: Can we follow the same morphology as other existent queries please?
You can keep the variable dns but I would prefer resource.
I'll leave a suggestion bellow:

```suggestion
	resource := input.document[i].google_dns_policy[name]
```

cx-andre-pereira: Done in https://github.com/Checkmarx/kics/pull/7773/commits/b2b38c1c8aff5c535e7dfc4de0627c93ac2e8451

cx-artur-ribeiro (approved):

PR #7774, opened 2025-10-17
fix(query): fix beta naming in similarityID transition docs

cx-miguel-dasilva:
**Proposed Changes**
- fix beta naming in similarityID transition docs


cx-artur-ribeiro (approved):
LGTM

PR #7781, opened 2025-10-21
feat(query): implemented query Beta - Key Vault Purge Protection Is Enabled for terraform/azure and fixed remediation problems

cx-ricardo-jesus:
- Currently, there is no query for Terraform/Azure that checks if the purge protection is enabled for an Azure Key Vault.
- Quoting [CIS_Microsoft_Azure_Foundations_Benchmark_v5.0.0](https://workbench.cisecurity.org/sections/3324211/recommendations/5454358): "`Key vaults contain object keys, secrets, and certificates. Deletion of a key vault can cause immediate data loss or loss of security functions (authentication, validation, verification, non-repudiation, etc.) supported by the key vault objects. It is recommended that the key vault be made recoverable by enabling the "purge protection" function.
- Currently, when running the remediate KICS command, an error occurs due to the variable ActualRemediationDoneNumber not being incremented for Beta queries(with the flag `experimental` set to true).

**Proposed Changes**
- Implemented the query that returns a positive result when the field `purge_protection_enabled` is not defined or not set to true.
- Regarding the metadata information, I took inspiration from a [similar query](https://docs.kics.io/latest/queries/azureresourcemanager-queries/azure/7c25f361-7c66-44bf-9b69-022acd5eb4bd/) implemented for the AzureResourceManager platform, which covers an analogous resource for AzureResourceManager, and also checks for the soft delete part. 
- Taking into account the information stated on the previous point, I think the most appropriate CWE is 530 and set the severity to High, taking into account that the similar query mentioned above, also has a HIGH severity.
- Regarding the remediation fix, basically, changed the Query struct on the `remediation.go`, to also store a boolean variable called experimental, which is extracted from the results file.
- Also, added the boolean `Experimental` variable on the `Remediation` structure.
- Basically, now, the remediationSet, which stores a set of Remediation structs, will also store the Experimental bool, which will be true or false, taking into account the value present on the `experimental` field on the results file.
- After these changes, the Beta queries will be capable of running the remediations, but, as expected, will not present any result.

### DIFF
--- a/assets/queries/terraform/gcp/google_dns_policy_logging_disabled/metadata.json
+++ b/assets/queries/terraform/gcp/google_dns_policy_logging_disabled/metadata.json
@@ -1,0 +1,14 @@
+{
+  "id": "cc9e464e-5abc-4c8f-8077-a9aa7ebe6a05",
+  "queryName": "Beta - Google DNS Policy Logging Disabled",
+  "severity": "MEDIUM",
+  "category": "Observability",
+  "descriptionText": "The 'google_dns_policy' should always have the 'enable_logging' field set to true",
+  "descriptionUrl": "https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/dns_policy#enable_logging-1",
+  "platform": "Terraform",
+  "descriptionID": "cc9e464e",
+  "cloudProvider": "gcp",
+  "cwe": "778",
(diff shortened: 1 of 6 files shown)


cx-artur-ribeiro (approved):
Insane job fixing the remediation for experimental queries!
The new query also looks well done, nothing to point out.

PR #7787, opened 2025-10-22
feat(query): implements "Beta - SQL DB Instance With Ownership Chaining Enabled"

cx-andre-pereira:
- Currently there is no query to ensure that a "[google_sql_database_instance](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance.html)" resource with a "SQLSERVER" based "[database_version](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance.html#database_version-1)" has the 'cross db ownership chaining' flag disabled.

- Quoting CIS_Google_Cloud_Platform_Foundation_Benchmark_v4.0.0 page 247: "```This flag is deprecated for all SQL Server versions in CGP. Going forward, you can't set 
its value to on. However, if you have this flag enabled, we strongly recommend that you 
either remove the flag from your database or set it to off.```" and "```Enabling cross db ownership is 
not recommended unless all of the databases hosted by the instance of SQL Server 
must participate in cross-database ownership chaining and you are aware of the 
security implications of this setting.```

**Proposed Changes**
  - Setting the "cross db ownership chaining" flag to a value that is not "off" (default value is off)

[Tenable reference](https://www.tenable.com/audits/items/CIS_Google_Cloud_Platform_Foundation_v4.0.0_L1.audit:938eca0e79ded99529f3003b9eda2e07)

### DIFF
--- a/pkg/model/model_yaml.go
+++ b/pkg/model/model_yaml.go
@@ -1,19 +1,25 @@
 package model
 
 import (
-	json "encoding/json"
+	"encoding/json"
 	"errors"
 	"path/filepath"
 	"strconv"
 
-	"github.com/Checkmarx/kics/v2/pkg/utils"
 	"github.com/rs/zerolog/log"
 	"gopkg.in/yaml.v3"
+
+	"github.com/Checkmarx/kics/v2/pkg/utils"
+)
+
+const (
+	mergeKey = "<<"
 )
 
 // UnmarshalYAML is a custom yaml parser that places line information in the payload
 func (m *Document) UnmarshalYAML(value *yaml.Node) error {
-	dpc := unmarshal(value)
+	visited := make(map[*yaml.Node]interface{})
+	dpc := unmarshalWithVisited(value, visited)
 	if mapDcp, ok := dpc.(map[string]interface{}); ok {
 		// set line information for root level objects
 		mapDcp["_kics_lines"] = getLines(value, 0)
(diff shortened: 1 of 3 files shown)


assets/queries/terraform/gcp/sql_db_instance_with_ownership_chaining_enabled/query.rego
```diff
@@ -0,0 +1,42 @@
...
+	contains(resource.database_version, "SQLSERVER")
+	results := get_results(resource, name)
+
+	result := {
+		"documentId": input.document[i].id,
+		"resourceType": "google_sql_database_instance",
+		"resourceName": tf_lib.get_resource_name(resource, name),
+		"searchKey": results.searchKey,
+		"issueType": "IncorrectValue",
+		"keyExpectedValue": sprintf("'google_sql_database_instance[%s].settings.database_flags' should be defined and set 'cross db ownership chaining' to 'off'", [name]),
```
cx-artur-ribeiro: Isn't this `expectedValue` misleading? 
Following the documentation description: "This flag is deprecated for all SQL Server versions in CGP. Going forward, you can't set its value to on. However, if you have this flag enabled, we strongly recommend that you either remove the flag from your database or set it to off", the default value is off, I believe we should say that the expected value "should be defined and set"...

Maybe change the message to something similar to other related case queries: 

```suggestion
		"keyExpectedValue": sprintf("'google_sql_database_instance[%s].settings.database_flags' should set 'cross db ownership chaining' to 'off'", [name]),
```

cx-artur-ribeiro (changes requested):
Let me know what you think of my suggestion @cx-andre-pereira .

assets/queries/terraform/gcp/sql_db_instance_with_ownership_chaining_enabled/query.rego
cx-andre-pereira: Done in https://github.com/Checkmarx/kics/pull/7787/commits/850c7c4bb584a6ec1558d3679080bc26f7149889

PR #7788, opened 2025-10-22
feat(query): implements "Beta - SQL DB Instance With Limited User Connections"

cx-andre-pereira:
- Currently there is no query to ensure that a "[google_sql_database_instance](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance.html)" resource with a "SQLSERVER" based "[database_version](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance.html#database_version-1)" has the 'user connections' flag set to ```0```.

- Quoting CIS_Google_Cloud_Platform_Foundation_Benchmark_v4.0.0 page 250: "```The user connections option specifies the maximum number of simultaneous user 
connections that are allowed on an instance of SQL Server.``` ...  ```SQL Server allows a 
maximum of 32,767 user connections. Because user connections is by default a self-configuring value, with SQL Server adjusting the maximum number of user connections 
automatically as needed, up to the maximum value allowable.``` ... ``` The default is 0, which means that the 
maximum (32,767) user connections are allowed. However if there is a number defined 
here that limits connections, SQL Server will not allow anymore above this limit. If the 
connections are at the limit, any new requests will be dropped, potentially causing lost 
data or outages for those using the database.```

**Proposed Changes**
  - Setting the "user connections" flag to a value that is not "0" (default value is 0)

[Tenable reference](https://www.tenable.com/audits/items/CIS_Google_Cloud_Platform_Foundation_v4.0.0_L1.audit:946949c5d4759f9b0165b2442df5ce1b)


assets/queries/terraform/gcp/sql_db_instance_with_limited_user_connections/metadata.json
```diff
@@ -0,0 +1,14 @@
+{
+  "id": "18cb7d28-57df-4d6b-9fb4-02828cb15660",
+  "queryName": "Beta - SQL DB Instance With Limited User Connections",
+  "severity": "MEDIUM",
+  "category": "Availability",
+  "descriptionText": "A 'google_sql_database_instance' resource based on SQLSERVER should not enable the 'user Connections' flag to prevent dropping requests or data loss in case user limit is reached",
```
cx-artur-ribeiro: I believe this description could be improved. Doesn't seem to align with the query itself, following the CIS Description. I would suggest improving to something along these lines: 

A 'google_sql_database_instance' resource based on SQLSERVER should have the user connections database flag unset or set to a non-limiting value (default 0). A small positive value may artificially limit concurrent connections and cause dropped requests or outages.

Just a suggestion, align with your preference, I just don't see the point of the description only mentioning that we should not enable the flag.

cx-andre-pereira: Adjusted description in https://github.com/Checkmarx/kics/pull/7788/commits/55010bf2b5759bd9b2fbec615cbff39517035755

cx-artur-ribeiro (dismissed):

PR #7789, opened 2025-10-22
feat(query): implements "Beta - SQL DB Instance With Global User Options"

cx-andre-pereira:
- Currently there is no query to ensure that a "[google_sql_database_instance](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance.html)" resource with a "SQLSERVER" based "[database_version](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance.html#database_version-1)" has the 'user options' flag set to ```0```.

- Quoting CIS_Google_Cloud_Platform_Foundation_Benchmark_v4.0.0 page 253: "```It is recommended that, user options database flag for Cloud SQL SQL Server 
instance should not be configured.
A user can override these defaults set with user options by using the SET statement. 
Some of these features/options could adversely affect the security of the system if 
enabled.```

**Proposed Changes**
  - Setting the "user options" flag to a value that is not "0" (default value is 0)

[Tenable reference](https://www.tenable.com/audits/items/CIS_Google_Cloud_Platform_Foundation_v4.0.0_L1.audit:dbcd48b6537a3b48bc2a3d54a1676c53)
[Info on "user options" values meaning](https://learn.microsoft.com/en-us/sql/t-sql/functions/options-transact-sql?view=sql-server-ver17)


cx-artur-ribeiro (approved):
LGTM!

cx-artur-ribeiro (approved):

PR #7802, opened 2025-10-30
perf(engine): optimize Terraform parser with directory caching and LOC-based memory calculation

cx-artur-ribeiro:
KICS experienced OOM errors when scanning large Terraform codebases due to:
- O(n²) complexity in variable resolution (rescanning all .tf files for every file parsed);
- Memory estimation based only on file count, ignoring actual file sizes;

**Proposed Changes**
- Added directory-level caching for Terraform variable resolution;
- Thread-safe cache reduces complexity from O(n²) to O(n);
- Eliminates redundant file scans (e.g., 247 terraform files: from 61,009 → 247 reads);
- Enhanced file statistics with LOC tracking;
- Provides per-platform statistics;


pkg/parser/terraform/variables.go
```diff
@@ -156,5 +153,49 @@ func getInputVariables(currentPath, fileContent, terraformVarsPath string) {
 		}
 	}
 
+	return variablesMap, nil
+}
+
+// getInputVariables now uses caching to avoid rescanning directories
+func getInputVariables(currentPath, fileContent, terraformVarsPath string) {
+	// Extract terraform vars path from file content if not provided
```
cx-miguel-dasilva, line 162: Add a comment about the inline comment

cx-artur-ribeiro: Added the comments removed here: https://github.com/Checkmarx/kics/pull/7802/commits/b1d5e41ec1f19e42cf6bfd10b5c2a5ba7bcf2b1c
Thanks for noticing Miguel.

PR #7805, opened 2025-10-31
feat(query): implements "Beta - Activity Log Alert For Create Policy Assignment Not Configured"

cx-andre-pereira:
- Currently there is no query to ensure that there is a "[azurerm_monitor_activity_log_alert](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/monitor_activity_log_alert)" resource configured to capture create policy assignment events. 

- Quoting CIS_Microsoft_Azure_Foundations_Benchmark_v5.0.0 page 227: "```Monitoring for create policy assignment events gives insight into changes done in
"Azure policy - assignments" and can reduce the time it takes to detect unsolicited
changes.```"

- Through the CIS benchmark we can also determine that detecting the target configuration boils down to checking 2 fields in the [```criteria```](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/monitor_activity_log_alert#criteria-1) block:
  - [```category```](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/monitor_activity_log_alert#category-1) should be set to "Administrative"
  - [```operation_name```](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/monitor_activity_log_alert#operation_name-1) should be set to "Microsoft.Authorization/policyAssignments/write"
- Additionally we must ensure the log alert resource does not set any "filters" that act "on Level, Status or
Caller" these can be set on the "azurerm_monitor_activity_log_alert" resource through the following fields : "[caller](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/monitor_activity_log_alert#caller-1)", "[level](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/monitor_activity_log_alert#level-1)", "[levels](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/monitor_activity_log_alert#levels-1)", "[status](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/monitor_activity_log_alert#status-1)", "[statuses](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/monitor_activity_log_alert#statuses-1)", "[sub_status](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/monitor_activity_log_alert#sub_status-1)" or "[sub_statuses](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/monitor_activity_log_alert#sub_statuses-1)". Since these could suppress or cause the log alert to miss relevant events they must not be set.
- Finally it must be ensured that the "[action_group_id](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/monitor_activity_log_alert#action_group_id-1)" field is set (inside an "[action](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/monitor_activity_log_alert#action-1)" block).

**Proposed Changes**
- The query will trigger if:
  - Both mandatory fields are correct but there is at least 1 "filter" field defined.
  - Both mandatory fields are correct and no filter is defined but the "action_group_id" field is missing.

- Additionally the query should flag the "project" itself when there is not a single log alert with both mandatory fields correct. Currently this is not possible but the query should be reworked once it is since there are plans to implement this feature eventually.

- In a given document if there are multiple "```azurerm_monitor_activity_log_alert```" and none of them fully align with the required requisites the query will prioritize flagging all the ones only missing the "action_group_id" first, then all the ones that flag due to a filter being set and, if none of the resources set the two mandatory fields to the expected values, it will flag all log alert resources. This priority is shown in the [positive4](https://github.com/Checkmarx/kics/pull/7805/files#diff-0c1b5aadcd1653320257a9a1ed6bb05c8aa64975e86d3a84ce82002b1315f750) test.


assets/queries/terraform/azure/activity_log_alert_for_create_policy_assignment_not_configured/query.rego
```diff
@@ -0,0 +1,105 @@
...
+				resource.criteria.operation_name == "Microsoft.Authorization/policyAssignments/write"}
+		}
+	logs[_] != {}
+
+} else = {"result" : "has_invalid_logs_only", "logs": resources}
+
+get_results(value) = results {					# Case of one or more resources failing due to not setting an "action.action_group_id" field
+	value.result == "has_log_without_action"
+
+	results := [z |
```
cx-artur-ribeiro: Can we find a more suitable and readable variable name please? `z` doesn't describe anything, might be hard to tackle any issue later once we find problems in this query.

PR #7820, opened 2025-11-04
fix(analyzer): handle encoded files extracted from zip archives on windows

cx-artur-ribeiro:
- KICS failed to scan Terraform plan files (tfplan.json) and other JSON/YAML files when extracted from ZIP archives on Windows, producing 0 results;
- Windows extracts large files from ZIP archives as UTF-16 LE with BOM (Byte Order Mark: 0xFF 0xFE) instead of UTF-8;
- This affected file type detection in the analyzer's `checkContent()` method for `.json`, `.yaml`, `.yml`, and `.sh` files, causing regex matching to fail and files to be skipped;
- The issue only manifested when scanning zipped folders, not individual zipped files;

**Proposed Changes**
- Added UTF-16 LE/BE to UTF-8 conversion utility functions:
  - `ReadFileToUTF8()` - reads files and converts from UTF-16 with BOM to UTF-8;
  - `ReadFileContentToUTF8()` - converts already-read byte content from UTF-16 with BOM to UTF-8;
- Updated `pkg/analyzer/analyzer.go` in `checkContent()` method to use `utils.ReadFileToUTF8()` when reading file content for regex-based type detection (affects `.json`, `.yaml`, `.yml`, `.sh` files);
- Updated `pkg/kics/service.go` in `getContent()` function to convert file content to UTF-8 before passing to parsers;
- Added comprehensive test coverage in `pkg/utils/file_reader_test.go` including the exact bug scenario (UTF-16 LE tfplan.json from ZIP);
- Added e2e test to properly test the improved behaviour.

**Usefull Prints**

When debugging, I came across the problem. When we just zipped a file directly and ran the scan, the issue didn’t exist. The problem occurred when we scanned a folder that was zipped, some files inside didn’t have UTF-8 format (as expected by os.ReadFile) and became corrupted:

image

This happened because KICS saves temporary files on Windows in the /Temp directory using UTF-16 LE encoding (sometimes). When the KICS Analyzer tried to determine which platform the file belonged to, it couldn’t pass any verification and classified the file as an unwantedFile, which resulted in its exclusion from the scan.

This only happened in specific cases where the file extension could belong to more than one platform, as shown below: 

image

Inside `checkContent` we would call `os.ReadFile`: 

image

We would then iterate over each possible platform and compare the available regex patterns to determine which platform the file belonged to:

image

The problem is now fixed for UTF-16 files (with or without BOM). This issue has been present since the initial implementation, so we might get a lot more results now 🥲.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -155,7 +155,7 @@ require (
 	github.com/boombuler/barcode v1.0.1 // indirect
 	github.com/cespare/xxhash/v2 v2.3.0 // indirect
 	github.com/chai2010/gettext-go v1.0.2 // indirect
-	github.com/containerd/containerd v1.7.28
+	github.com/containerd/containerd v1.7.29
 	github.com/cyphar/filepath-securejoin v0.4.1 // indirect
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
 	github.com/emicklei/go-restful/v3 v3.12.2 // indirect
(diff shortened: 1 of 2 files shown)


pkg/utils/file_reader.go
```diff
@@ -0,0 +1,110 @@
...
+	bytesContent, err := os.ReadFile(filepath.Clean(path))
+	if err != nil {
+		return nil, err
+	}
+	if len(bytesContent) == 0 {
+		return bytesContent, nil
+	}
+
+	// BOM detection
+	if len(bytesContent) >= 2 {
```
cx-rui-araujo: Can you call ReadFileContentToUTF8 instead?
Like: `return ReadFileContentToUTF8(bytesContent, path)`

cx-artur-ribeiro: I've changed the function to call the function `ReadFileContentToUTF8`, removing unnecessary duplicated code. Thnaks for the warning! 

I've also renamed the e2e test and zip file to be more explicit (and fixed a typo on the e2e comment) 😄

PR #7821, opened 2025-11-05
feat(query): implements "Beta - Activity Log Alert For Service Health Not Configured"

cx-andre-pereira:
- Currently there is no query to ensure that there is a "[azurerm_monitor_activity_log_alert](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/monitor_activity_log_alert)" resource configured to capture "Service Health" events.

- Quoting CIS_Microsoft_Azure_Foundations_Benchmark_v5.0.0 page 267: "```Monitoring for Service Health events provides insight into service issues, planned
maintenance, security advisories, and other changes that may affect the Azure services
and regions in use.```"

- Through the CIS benchmark we can also determine that detecting the target configuration boils down to checking 4 conditions :
  -  The "[scopes](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/monitor_activity_log_alert#scopes-1)" field should target a subscription. 
  - The [```category```](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/monitor_activity_log_alert#category-1) field should be set to "ServiceHealth"
  - In the "[events](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/monitor_activity_log_alert#events-1)" array, "Incident" should be included
  - Finally it must be ensured that the "[action_group_id](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/monitor_activity_log_alert#action_group_id-1)" field is set (inside an "[action](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/monitor_activity_log_alert#action-1)" block).

- It is expected that all "subscriptions" are associated in this manner, the data block "[azurerm_subscription](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/data-sources/subscription)" allows defining these subscriptions, and so the query should iteratively check all associated alert logs for each subscription. 

**Proposed Changes**
- The query will start by collecting all log alert type resources as well as all subscriptions defined through data azurerm_subscription blocks. 
- Then, for each subscription, the log alerts are scrubbed to search for at least one that fulfills requisites 2, 3 and 4 and references the specific subscription the auxiliary function "[at_least_one_valid_log_alert](https://github.com/Checkmarx/kics/pull/7821/files#diff-51572ff917c019edd900583bc6fbdbca287caeb091479bbba08dfd2d4e763978R42)" got called for. 
- The query will trigger differently depending on the log_alerts found, with the following priority :
  - Flag for the "action_group_id" not being set
  - Flag if the "events" array does not include "Incident"
  - Flag if the "category" is set to a value that is not "ServiceHealth"
  - Flag the subscription itself if there is not a single log alert resource associated with it. 
 
- These priorities are demonstrated through the [positive4](https://github.com/Checkmarx/kics/pull/7821/files#diff-543674ce680fbbc31c76b996bd4567afa03fd6e2092a7e69475e13f3ba464740) and [positive5](https://github.com/Checkmarx/kics/pull/7821/files#diff-a5e0aba0b73c152bf36bc7ef57dceea545746ab186f4fb774c3c72249cdeaf0a) tests and their results.


cx-eduardo-semanas (approved):
LGTM

cx-eduardo-semanas (approved):
LGTM

PR #7822, opened 2025-11-05
feat(query): implemented query to cover "Ensure Legacy Networks Do Not Exist For Older Project" for terraform/gcp

cx-ricardo-jesus:
- Currently, no query for Terraform/GCP checks if legacy networks do not exist for older projects.

**Proposed Changes**
- Implemented the query that returns a positive result when the field `auto_create_subnetworks` is not defined ir is set to true, when it's associated with a resource of type `google_project` through the field `project` or when this field is missing, through the provider defined on the template.
- The query implementation consists of two policies. 
- The first policy is for the cases when the `project` field is defined inside the resource of type `google_compute_network`. In these cases, the field `project` is defined on the following syntax `google_project.<GOOGLE_PROJECT_NAME>.id`. So the policy just checks if the google project resource name is the same as the one referenced on the field and if it is, it returns a positive result if the field `auto_create_subnetworks` is not defined or is defined to true.
- The second policy, is for the cases when the `project` field is not defined inside the `google_compute_network` resource, and in these scenarios, the resource uses the value present on the provider(for this case, it can be `google` or `google-beta`). in all the examples found through some research, the provider is defined with the following syntax: 
```
provider "google" {
  project = "my-project-id"
  region  = "us-central1"
}
```
- The value on the project field, inside the provider declaration, has to have the same value as the value present on the field `project_id` from the `google_project` resource, which differs from the first case, which uses the name of the`google_project` resource.
- Choosed the severity MEDIUM, since a [similar query](https://docs.kics.io/latest/queries/terraform-queries/gcp/59571246-3f62-4965-a96f-c7d97e269351/) also has this severity.
- The CWE that seemed the most suited for this case is the CWE CWE: 276 - Incorrect Default Permissions, because the default value of the field `auto_create_subnetworks` is unsafe.
- NOTE: after further investigation, some information about the default value of the field `auto_create_subnetworks` was found on the [google terraform provider official git repository](https://github.com/hashicorp/terraform-provider-google/blob/41cef4736a617ff48b67bcc78b2ae55b2a700a8b/google/services/compute/resource_compute_network.go#L123C5-L135C6), and its default value is set to `true` which should return a positive result.


assets/queries/terraform/gcp/legacy_networks_do_not_exist_for_older_google_projects/metadata.json
```diff
@@ -0,0 +1,14 @@
...
+  "id": "700f1049-7fa0-4cb0-971b-3efebfb6a91f",
+  "queryName": "Beta - Legacy Networks Do Not Exist For Older Google Projects",
+  "severity": "MEDIUM",
+  "category": "Insecure Configurations",
+  "descriptionText": "Legacy networks can have an impact for high network traffic projects and are subject to a single point of contention or failure.",
+  "descriptionUrl": "https://registry.terraform.io/providers/hashicorp/google/6.36.1/docs/resources/compute_network",
+  "platform": "Terraform",
+  "descriptionID": "700f1049",
+  "cloudProvider": "gcp",
+  "cwe": "693",
```
cx-eduardo-semanas: Should be updated to a CWE which isn't considered DISCOURAGED (https://cwe.mitre.org/data/definitions/693.html), a possibility would be the 276 (https://cwe.mitre.org/data/definitions/276.html)

assets/queries/terraform/gcp/legacy_networks_do_not_exist_for_older_google_projects/query.rego
```diff
@@ -0,0 +1,60 @@
+package Cx
+
+import data.generic.common as common_lib
+import data.generic.terraform as tf_lib
+
+# field project defined inside the google_compute_network_resource
```
cx-eduardo-semanas: Shouldn't it be "field project defined inside the google_compute_network resource", without the '_' before "resource"

cx-eduardo-semanas:
You should update the assets/similarityID_transition/terraform_gcp.yaml file with a new entry corresponding to this query with change: 2

cx-eduardo-semanas (approved):
LGTM

cx-eduardo-semanas (dismissed):
LGTM

cx-eduardo-semanas (approved):
LGTM

PR #7827, opened 2025-11-06
feat(query): implements "Beta - File Share Without Soft Delete"

cx-andre-pereira:
- Currently there is no query to ensure that resources of type "[azurerm_storage_account](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_account#retention_policy-2)" configure a value of 1 to 365 on their "[share_properties.retention_policy.days](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_account#days-8)" field.

- Quoting CIS_Microsoft_Azure_Foundations_Benchmark_v5.0.0 page 476: "```Important data could be accidentally deleted or removed by a malicious actor. With soft
delete enabled, the data is retained for the defined retention period before permanent
deletion, allowing for recovery of the data.```"

- Through the Terraform documentation for the resource we can see a "[retention_policy](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_account#retention_policy-2)" block left empty will assume the default value of 7 for the missing "days" field. With this all we have to do is ensure the "retention_policy" block is defined regardless of the value associated with the "days" field.  

- Note that a validation check for the "days" field (to ensure it falls between 1 and 365, inclusive) was not explicitly implemented. This is because Terraform's provider schema enforces this constraint natively — any value outside the valid range will trigger an error during the "```terraform plan```" phase.

**Proposed Changes**
- The query will flag any "azurerm_storage_account" resource that does not set the "share_properties" block, or does set it but does not set a "retention_policy" block within. 


cx-eduardo-semanas (approved):
LGTM

PR #7832, opened 2025-11-06
feat(query): implements "Beta - Storage Account With Shared Access Key"

cx-andre-pereira:
- Currently there is no query to ensure that resources of type "[azurerm_storage_account](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_account#retention_policy-2)" set the "[shared_access_key_enabled](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_account#shared_access_key_enabled-1)" field to ```false```.

- Quoting CIS_Microsoft_Azure_Foundations_Benchmark_v5.0.0 page 504: "```Microsoft Entra ID provides superior security and ease of use compared to Shared Key
and is recommended by Microsoft. To require clients to use Microsoft Entra ID for
authorizing requests, you can disallow requests to the storage account that are
authorized with Shared Key.```"

**Proposed Changes**
- The query will flag any "azurerm_storage_account" resource that does not set the "shared_access_key_enabled" field or sets it  to a value that is not ```false```.


cx-eduardo-semanas (dismissed):
LGTM

PR #7834, opened 2025-11-07
feat(query): implements "Beta - Containers Without Soft Delete"

cx-andre-pereira:
- Currently there is no query to ensure that resources of type "[azurerm_storage_account](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_account#retention_policy-2)" configure a value of 7 to 365 on their "[blob_properties.container_delete_retention_policy.days](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_account#days-1)" field.

- Quoting CIS_Microsoft_Azure_Foundations_Benchmark_v5.0.0 page 489: "```Containers can be deleted incorrectly. An attacker or malicious user may do this
deliberately in order to cause disruption. Deleting a container results in immediate data
loss. Enabling this configuration for Azure storage accounts ensures that even if
containers are deleted from the storage account, the containers are recoverable for a
specific period of time, which is defined in the "Retention policies," ranging from 7 to
365 days.```"

- Through the Terraform documentation for the resource we can see a "[container_delete_retention_policy](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_account#days-3)" block left empty will assume the default value of 7 for the missing "days" field. With this all we have to do is ensure the "delete_retention_policy" block is defined and that the value associated with the "days" field, if set, is higher than ```6```.

**Proposed Changes**
- The query will flag any "azurerm_storage_account" resource that does not set the "[blob_properties](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_account#blob_properties-1)" block, or does set it but does not set a "container_delete_retention_policy" block within. Additionally it will flag if the "[days](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_account#days-1)" field is set to a value lower than 6.

- Note that a check to ensure the days field is "valid" as in between 1 and 365 was not done since those would be found in the "terraform plan" stage. (was tested)


cx-eduardo-semanas (approved):
LGTM

PR #7836, opened 2025-11-08
feat(query): implements "Beta - Storage Account Without Delete Lock"

cx-andre-pereira:
- Currently there is no query to ensure that resources of type "[azurerm_storage_account](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_account#retention_policy-2)" are associated with a "[azurerm_management_lock](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/management_lock)" resource with "[lock_level](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/management_lock#lock_level-1)" set to ```CanNotDelete```.

- Quoting CIS_Microsoft_Azure_Foundations_Benchmark_v5.0.0 page 539: "```Azure Resource Manager CannotDelete (Delete) locks can prevent users from
accidentally or maliciously deleting a storage account. This feature ensures that while
the Storage account can still be modified or used, deletion of the Storage account
resource requires removal of the lock by a user with appropriate permissions.
This feature is a protective control for the availability of data. By ensuring that a storage
account or its parent resource group cannot be deleted without first removing the lock,
the risk of data loss is reduced.
While an automated assessment procedure exists for this recommendation, the
assessment status remains manual. Determining storage accounts that require
CannotDelete locks depends on the context and requirements of each organization and
environment.```"

- Note that an "azurerm_management_lock" can be associated to a resource directly or by associating with the "[azurerm_resource_group](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/resource_group)" that contains the target.

**Proposed Changes**
- The query will flag a given"azurerm_storage_account" resource if: 
  - There isn't a single "azurerm_management_lock" associated with it or its resource group within the project
  - There is an "azurerm_management_lock" associated with it or its resource group but "lock_level" is not set to ```CanNotDelete```.


cx-eduardo-semanas (dismissed):
LGTM

PR #7841, opened 2025-11-10
feat(query): implemented 'ensure critical contacts is configured for organization' query for terraform/gcp

cx-ricardo-jesus:
- Currently, no query for Terraform/GCP ensures that the essential contacts are configured for an organization.

**Proposed Changes**
- On this query there is only one policy, that uses the `contacts_not_configured_for_org` helper function. 
- The helper function uses other two helper functions called `is_at_organization_level` and `all_in_list` and returns true if the field `parent` is defined in a certain way that indicates that the resource is on a organization level and if the field `notification_category_subscriptions` does not have the value `ALL` defined or all the following values defined: `LEGAL`, `SECURITY`, `SUSPENSION` and `TECHNICAL`. These are the value that have to be defined to ensure that the essential contacts are configured, according to the respective CIS Workbeck [documentation](https://workbench.cisecurity.org/sections/2584065/recommendations/4175061).
- The helper function check's if the field `parent` points to an organization using two different notations. The first one is `RESOURCE_TYPE/RESOURCE_ID` with the `RESOURCE_TYPE` that can be defined with the values `projects`, `folders`, or `organizations`, defined with the value `organization`, in order to stay in the scope of the query. This information was found on the following [Google Cloud documentation](https://cloud.google.com/resource-manager/docs/manage-essential-contacts).
- After further research, I found other way of defining this field that is by, pointing to a [data source](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/organization) of type `google_organization` using the name defined for the data source, which, in all the examples that I found on my research, with the following syntax: `data.<DATA_SOURCE_TYPE>.<DATA_SOURCE_NAME>.(id|org|name|...)`.
- On the metadata of the query, I think that the most suitable CWE is the one with the number 862 - `Missing Authorization`.


cx-eduardo-semanas:
You should update the assets/similarityID_transition/terraform_gcp.yaml file with a new entry corresponding to this query with change: 2

PR #7842, opened 2025-11-11
feat(query): implemented query that checks if the use of user access administrator is not restricted for terraform/azure

cx-ricardo-jesus:
- Currently, there is no query for Terraform/Azure that checks if an resource of type `azurerm_role_assignement` is restricting the use of `User Access Administrator` role through its name or id.

**Proposed Changes**
- The query has just one policy, that uses an helper function called `get_res` that check's the two possibilities of defining the `User Access Administrator` role, by using the field `role_definition_name` and defining it to `User Access Administrator` or its [respective ID](https://learn.microsoft.com/en-us/azure/role-based-access-control/built-in-roles) (`18d7d88d-d35e-4fb5-a5c3-7773c20a72d9`).
- According to the [documentation](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/role_assignment#role_definition_name-1), `"Either role_definition_id or role_definition_name must be set."` so, I did not covered the case when both are defined.
- Additionally based on the AppSec team's feedback, the `scope` field should also be taken into account. Following this guidance, the query only covers the highest permission levels - the tenant root (`/`) and management group (`/providers/Microsoft.Management/managementGroups/<mg>`). According to the [Microsoft RBAC scope documentation](https://learn.microsoft.com/en-us/azure/role-based-access-control/scope-overview), assigning the **User Access Administratir** role at these level grants permissions over all resouces in the tenant.

### DIFF
--- a/assets/queries/terraform/azure/recovery_services_vaut_without_soft_delete/metadata.json
+++ b/assets/queries/terraform/azure/recovery_services_vaut_without_soft_delete/metadata.json
@@ -1,0 +1,14 @@
+{
+  "id": "b373043c-f3bf-40db-b67a-c982732c7781",
+  "queryName": "Beta - Recovery Services Vault Without Soft Delete",
+  "severity": "HIGH",
+  "category": "Backup",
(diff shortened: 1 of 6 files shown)


cx-eduardo-semanas (approved):
LGTM

PR #7845, opened 2025-11-13
feat(query): implements "Beta - Backup Vault Without Soft Delete"

cx-andre-pereira:
- Currently there is no query to ensure that resources of type "[azurerm_data_protection_backup_vault](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/data_protection_backup_vault)" do not set the "[```soft_delete```](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/data_protection_backup_vault)" field to ```"off"```

- Quoting CIS_Microsoft_Azure_Storage_Services_Benchmark_v1.0.0 page 59: "```Important backup data could be accidentally deleted or removed by a malicious actor.
With soft delete enabled, data is retained for at least 14 days before permanent
deletion, allowing for the recovery of the backup data.```"

**Proposed Changes**
- The query will flag any "azurerm_data_protection_backup_vault" resource that sets "soft_delete" to off. 


cx-eduardo-semanas (approved):
LGTM

PR #7848, opened 2025-11-14
feat(query): implements "Beta - Backup Vault Without Immutability"

cx-andre-pereira:
- Currently there is no query to ensure that resources of type "[azurerm_data_protection_backup_vault](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/data_protection_backup_vault)" do not set the "[```immutability```](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/data_protection_backup_vault#immutability-1)" field to ```"Disabled"```

- Quoting CIS_Microsoft_Azure_Storage_Services_Benchmark_v1.0.0 page 62: "```Enabling the immutable vault ensures that backup data is protected from unauthorized
or accidental deletion. By locking the setting and making it irreversible, malicious actors
are prevented from disabling the setting and deleting backups.```"

**Proposed Changes**
- The query will flag any "azurerm_data_protection_backup_vault" resource that sets "immutability" to ```Disabled```, or does not set the field since it is disabled by default.


cx-eduardo-semanas (approved):
LGTM

cx-eduardo-semanas (approved):
LGTM

PR #7851, opened 2025-11-14
feat(query): implements "Beta - Recovery Services Vault With Public Network Access"

cx-andre-pereira:
- Currently there is no query to ensure that resources of type "[azurerm_recovery_services_vault](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/recovery_services_vault)" set the "[```public_network_access_enabled```](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/recovery_services_vault#public_network_access_enabled-1)" field to ```false```.

- Quoting CIS_Microsoft_Azure_Storage_Services_Benchmark_v1.0.0 page 94: "```Disable public network access on Recovery Services vaults to prevent exposure to the
internet and reduce the risk of unauthorized access. Use private endpoints and Azure
Role-Based Access Control (RBAC) to securely manage access within trusted
networks.```" and "```Disabling public network access improves security by ensuring that a Recovery
Services vault is not exposed on the public internet.```".

**Proposed Changes**
- The query will flag any "azurerm_recovery_services_vault" resource that sets "public_network_access_enabled" to ```true```, or does not set the field since it is enabled by default. 


cx-eduardo-semanas (dismissed):
LGTM

PR #7852, opened 2025-11-14
feat(queries): implemented queries that checks if the tls encryption version is set to '1.2' or higher for terrraform/azure

cx-ricardo-jesus:
- Currently, some relevant resources do not have a query that checks if the TLS encryption version is defined to '1.2' or higher.

**Proposed Changes**
- Implemented 4 queries that check if the resources below have the TLS encryption defined to '1.2' or a higher version:
    - query `Beta - Function App Deployment Slot Not Using Latest TLS Encryption Version`:
        - `azurerm_linux_function_app_slot`
        - `azurerm_windows_function_app_slot`
    - query `Beta - MSSQL Not Using Latest TLS Encryption Version`:
        - `azurerm_mssql_server`
    - query `Beta - PostgreSQL Not Using Latest TLS Encryption Version`: 
       - `azurerm_postgresql_server`
    - query `Beta - Redis Cache Not Using Latest TLS Encryption Version`:
       - `azurerm_redis_cache`   

- For all the queries, the default value is '1.2', but for the Redis Cache resource, the default value is defined '1.0', which should trigger a positive result.

### DIFF
--- a/assets/queries/terraform/azure/storage_account_without_cmk/metadata.json
+++ b/assets/queries/terraform/azure/storage_account_without_cmk/metadata.json
@@ -1,0 +1,14 @@
+{
+  "id": "9bf1568d-4cd2-4581-81ef-d2efabee1178",
+  "queryName": "Beta - Storage Account Without CMK",
+  "severity": "MEDIUM",
+  "category": "Encryption",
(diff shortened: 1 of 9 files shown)


cx-eduardo-semanas (dismissed):
LGTM

PR #7863, opened 2025-11-18
feat(queries): implemented queries to check if managed identity is not enabled

cx-ricardo-jesus:
Closes #

- Currently, there is several terrraform/azure resources, that are not covered by a query that checks if their `identity` block is not enabled, i.e., by having a block `identity` defined with the `type` field with the values `SystemAssigned` or `UserAssigned`, or both(except for the resource of type `azurerm_kubernetes_cluster` that only can have one of the values at time, according to its [documentation](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster)).
- For all the resources, this case can be covered just by checking if the `identity` block is defined.
- Below is a list of queries and resources supported in each one:
    - `Beta - Container Group Managed Identity Disabled`:
        - `azurerm_container_group`
    - `Beta - Kubernetes Cluster Managed Identity Disabled`:
        - `azurerm_kubernetes_cluster` 
    - `Beta - Container App Managed Identity Disabled` 
        - `azurerm_container_app` 
    - `Beta - App Service Slot Managed Identity Disabled`
        - `azurerm_app_service_slot`
        - `azurerm_linux_web_app_slot`
        - `azurerm_windows_web_app_slot`
    - `Beta - Logic App Managed Identity Disabled`   
        - `azurerm_logic_app_standard`
    
- There is an exception in one of the queries from this PR, that is the `Beta - Redis Cache Managed Identity Is Not Set To System Assigned`, which checks if the 'System Assigned Managed Identity' is set to 'On' for Azure Redis Cache (`azurerm_redis_cache`).
- This last query covers a different scenario from CIS Benchmark, that instead of checking if the `type` field inside the `identity` block is defined to `SystemAssigned` or `UserAssigned`, it only checks if the field has the value `SystemAssigned`.

**Proposed Changes**
- Used the same implementation for all the queries and resources since, for all, just checking if the block `identity` is defined is enough because the field `type` is required and the only values that can be defined within are the ones mentioned above.
- For the last query, there is a single policy that uses a helper function called get_res, which covers the two cases that can return a positive result. The first one occurs when the block identity is not defined, and the other one occurs when the field type is defined inside the identity block, but there is no SystemAssigned value specified within the string for this field. 

### DIFF
--- a/assets/queries/terraform/azure/kubernetes_cluster_managed_identity_disabled/query.rego
+++ b/assets/queries/terraform/azure/kubernetes_cluster_managed_identity_disabled/query.rego
@@ -18,6 +18,6 @@ CxPolicy[result] {
 		"keyActualValue": "'identity' block is not defined",
 		"searchLine": common_lib.build_search_line(["resource", "azurerm_kubernetes_cluster", name], []),
         "remediationType": "addition",
-        "remediation": "identity {\n\t\ttype = \"SystemAssigned, UserAssigned\"\n\t}",
+        "remediation": "identity {\n\t\ttype = \"SystemAssigned\"\n\t}",
     }
 }


cx-eduardo-semanas (approved):
LGTM

cx-miguel-dasilva (dismissed):

cx-artur-ribeiro (approved):
LGTM

PR #7864, opened 2025-11-19
perf(engine): optimize Terraform parser with directory caching and LOC-based memory calculation

cx-artur-ribeiro:

KICS experienced OOM errors when scanning large Terraform codebases due to:
- O(n²) complexity in variable resolution (rescanning all .tf files for every file parsed);
- Memory estimation based only on file count, ignoring actual file sizes;

**Proposed Changes**
- Added directory-level caching for Terraform variable resolution;
- Thread-safe cache reduces complexity from O(n²) to O(n);
- Eliminates redundant file scans (e.g., 247 terraform files: from 61,009 → 247 reads);
- Enhanced file statistics with LOC tracking;
- Provides per-platform statistics;
- Fixed secrets inspector to not fail scan when secrets queires timeout;

### DIFF
--- a/assets/queries/terraform/gcp/ensure_essential_contacts_is_configured_for_organization/metadata.json
+++ b/assets/queries/terraform/gcp/ensure_essential_contacts_is_configured_for_organization/metadata.json
@@ -1,0 +1,14 @@
+{
+  "id": "7bd9c6a8-3b1f-495c-9752-a4a9c4e1b29f",
+  "queryName": "Beta - Ensure Essential Contacts Is Configured For Organization",
+  "severity": "LOW",
+  "category": "Access Control",
+  "descriptionText": "It is advisable to set up Essential Contacts to specify email addresses that Google Cloud can use to send important technical or security notifications.",
+  "descriptionUrl": "https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/essential_contacts_contact",
+  "platform": "Terraform",
+  "descriptionID": "7bd9c6a8",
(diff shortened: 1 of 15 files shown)


pkg/parser/terraform/variables.go
```diff
@@ -124,37 +138,69 @@ func getInputVariables(currentPath, fileContent, terraformVarsPath string) {
...
-		} else {
-			variables, errInputVariables := getInputVariablesFromFile(terraformVarsPath)
-			if errInputVariables != nil {
-				log.Error().Msgf("Error getting values from %s", terraformVarsPath)
-				log.Err(errInputVariables)
-			} else {
-				mergeMaps(variablesMap, variables)
-			}
-		}
+		cacheKey = currentPath + "|" + terraformVarsPath
```
cx-rui-araujo, line 181: If the cache key is the full path, it can sometimes become quite large. Should this be a hash instead?
Please ask the rest of the team for their opinions

cx-rui-araujo (commented):
@cx-artur-ribeiro check my comment pls

pkg/parser/terraform/variables.go
cx-artur-ribeiro: After further investigation I believe we do not need to make that effort since the paths are not going to be that long and we need a way to decode the paths if we ran into any problem, so a hash wouldn't be appropriate here (although we do not log the information at the moment).
Thanks for the great suggestion eitherway!

cx-miguel-silva (approved):

PR #7874, opened 2025-11-24
feat(query): implements "Beta - Storage Account Without CMK"

cx-andre-pereira:
- Currently there is no query to ensure a "[azurerm_storage_account](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_account)" resource has [CMK encryption](https://learn.microsoft.com/en-us/azure/storage/common/customer-managed-keys-overview) enabled.

- Quoting CIS_Microsoft_Azure_Foundations_Benchmark_v5.0.0 page 203: "```By default in Azure, data at rest tends to be encrypted using Microsoft-managed keys. If
your organization wants to control and manage encryption keys for compliance and
defense-in-depth, customer-managed keys can be established.
Configuring the storage account with the activity log export container to use CMKs
provides additional confidentiality controls on log data, as a given user must have read
permission on the corresponding storage account and must be granted decrypt
permission by the CMK.
While it is possible to automate the assessment of this recommendation, the
assessment status for this recommendation remains 'Manual' due to ideally limited
scope. The scope of application—which workloads CMK is applied to—should be
carefully considered to account for organizational capacity and targeted to workloads
with specific need for CMK.```"

- Additionally the "Audit from Azure CLI" section of the CIS Benchmark shows that not all "azurerm_storage_account" resources are to be flagged, only the ones associated with a "[azurerm_monitor_diagnostic_setting](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/monitor_diagnostic_setting)" should be targeted. 

- To enable CMK encryption for a storage account there are 2 possibilities: 
  - Setting a "[customer_managed_key](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_account#customer_managed_key-1)" field inside the "azurerm_storage_account" resource
  - Having a standalone "[azurerm_storage_account_customer_managed_key](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_account_customer_managed_key)" that is associated through its "storage_account_id" field to the target storage account.
 
**Proposed Changes**

- The query will flag a given"```azurerm_storage_account```" resource that is associated with a "```azurerm_monitor_diagnostic_setting```" resource, if it does not set the "```customer_managed_key```" field and has no association with any "```azurerm_storage_account_customer_managed_key```" resource within the same project.

### DIFF
--- a/assets/queries/terraform/azure/azure_container_registry_with_broad_permissions/metadata.json
+++ b/assets/queries/terraform/azure/azure_container_registry_with_broad_permissions/metadata.json
@@ -1,0 +1,14 @@
+{
+  "id": "77deea6a-155e-4865-bf04-153d23e488e8",
+  "queryName": "Beta - Azure Container Registry With Broad Permissions",
+  "severity": "HIGH",
+  "category": "Access Control",
+  "descriptionText": "Azure Container Registry resources should only have 'read' permissions",
+  "descriptionUrl": "https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/container_registry",
+  "platform": "Terraform",
+  "descriptionID": "77deea6a",
+  "cloudProvider": "azure",
+  "cwe": "732",
+  "riskScore": "6.0",
+  "experimental": "true"
+}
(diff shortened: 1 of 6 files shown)


cx-eduardo-semanas (dismissed):
LGTM

PR #7877, opened 2025-11-24
feat(query): new query that ensures that container instances are using private virtual networks for terraform/azure

cx-ricardo-jesus:
- Currently, no query ensures that container instances are using private virtual networks for terraform/azure.

**Proposed Changes**
- After researching through the [azurerm_container_group resource documentation](https://registry.terraform.io/providers/hashicorp/azurerm/4.54.0/docs/resources/container_group), it is clear that, In order to check if the container instance is using private virtual networks, it is imperative to check the value defined on the `ip_address_type` field is defined to `Private`. Another field that could be relevant in this scenario is checking if the field `subnet_id` is defined. However, according to the documentation, if the field `ip_address_type` is defined as `Private`, the field `subnet_id` must also be defined. Defining the field `ip_address_type` to `Private` and not defining the field `subnet_id` will result in an error, as seen below, so it is only necessary to check if the field `ip_address_type` is defined to `Private`.
image

- Regarding the query implementation itself, the query is composed of just one policy that uses the helper function `get_res`. The helper function `get_res` checks if the field `ip_address_type` is not defined or set to `Private`, resulting in a positive result if either of these scenarios occurs.


cx-eduardo-semanas (approved):
LGTM

PR #7879, opened 2025-11-25
feat(query): implemented query that ensures that kubernetes web ui is disabled (field deprecated)

cx-ricardo-jesus:
- Currently, no query covers the case when the Kubernetes Web UI (Dashboard) is not disabled for Terraform/GCP.

**Proposed Changes**
- The query only has one simple policy, which only checks if the field `addons_config.kubernetes_dashboard.disabled` is defined to false.
- If this `kubernetes_dashboard` is not defined, the default behaviour is to disable de dashboard, which should not be flagged.
- Important note is that this field was removed in version [3.0.0](https://registry.terraform.io/providers/hashicorp/google/2.20.3/docs/resources/container_cluster#kubernetes_dashboard-4) of Terraform/GCP `google_container_cluster` resource, which was released approximately 6 years ago.
- Taking into account that this field is just flagged if it's defined to false, and its absence does not return any result, it will not return any positive result on the most recent versions of this resource.
- Regarding the query itself, it will flag the cases when a resource of type `google_container_cluster` does not have the block `addons_config` defined or when the addon `kubernetes_dashboard` is not defined inside the `addons_config` block. Both these cases are only flagged when the GKE version is under 1.10 using the value present on the `min_master_version` field. For versions above or equal to 1.10, it will not return any result since in these versions, the kubernetes_dashboard is disabled by default.
- No matter the version, the query will also return a positive result if the field `addons_config.kubernetes_dashboard.disabled` is set to `false`.

### DIFF
--- a/assets/queries/terraform/gcp/kubernetes_web_ui_is_not_disabled/query.rego
+++ b/assets/queries/terraform/gcp/kubernetes_web_ui_is_not_disabled/query.rego
@@ -40,7 +40,7 @@ get_results(resource, name) = res {
         "sk": sprintf("google_container_cluster[%s]", [name]),
         "it": "MissingAttribute",
         "kev": "'kubernetes_dashboard' should be defined and disabled inside the 'addons_config_version' block for GKE versions below 1.10",
-        "kav": "'addons_config' block is not defined with the 'kubernetes_dashboard' enabled",
+        "kav": "'addons_config' block is not defined with the 'kubernetes_dashboard' disabled",
         "sl": common_lib.build_search_line(["resource", "google_container_cluster", name], [])
     }
 } else = res {


assets/queries/terraform/gcp/kubernetes_web_ui_is_not_disabled/test/negative3.tf
```diff
@@ -0,0 +1,8 @@
+# incluir versão do GKE (min_master_version) superior a 1.10
```
cx-miguel-dasilva: Please translate the comment into English

assets/queries/terraform/gcp/kubernetes_web_ui_is_not_disabled/test/positive2.tf
```diff
@@ -0,0 +1,9 @@
+# incluir versão do GKE (min_master_version) inferior a 1.10
```
cx-miguel-dasilva: Please translate the comment to english

assets/queries/terraform/gcp/kubernetes_web_ui_is_not_disabled/test/positive3.tf
```diff
@@ -0,0 +1,8 @@
+# incluir versão do GKE (min_master_version) inferior a 1.10
```
cx-miguel-dasilva: Please translate the comment to english

assets/similarityID_transition/terraform_gcp.yaml
```diff
@@ -79,6 +79,8 @@ similarityIDChangeList:
       queryName: Beta - SQL DB Instance With Exposed Show Privileges
       observations: ""
       change: 2
+    - queryId: ced11de2-e701-4e63-83ab-4fdb1ab8c5dd
+      queryName: Beta - Kubernetes Web UI Is Not Disabled
```
cx-miguel-dasilva, line 87:
```suggestion
      queryName: Beta - Kubernetes Web UI Is Not Disabled
      change: 1
      observations: ""      
```

assets/queries/terraform/gcp/kubernetes_web_ui_is_not_disabled/metadata.json
```diff
@@ -0,0 +1,14 @@
...
+  "id": "ced11de2-e701-4e63-83ab-4fdb1ab8c5dd",
+  "queryName": "Beta - Kubernetes Web UI Is Not Disabled",
+  "severity": "LOW",
+  "category": "Insecure Configurations",
+  "descriptionText": "Kubernetes Web UI (Dashboard) should be disabled when running on Kubernetes Engine",
+  "descriptionUrl": "https://registry.terraform.io/providers/hashicorp/google/2.20.3/docs/resources/container_cluster#kubernetes_dashboard-4",
+  "platform": "Terraform",
+  "descriptionID": "ced11de2",
+  "cloudProvider": "gcp",
+  "cwe": "250",
```
cx-miguel-dasilva: Review the CWE choice; 250 does not seem the best fit for this query.

cx-miguel-dasilva (dismissed):
Great Work!

PR #7885, opened 2025-11-26
feat(query): implemented query that ensures that gke version management is automated using release channels

cx-ricardo-jesus:
- Currently, no query ensures that GKE version management is automated using release channels.

**Proposed Changes**
- Taking into account the information present on the [CIS Google Kubernetes Engine (GKE) Benchmark v1.9.0](https://workbench.cisecurity.org/sections/3744750/recommendations/6213250), it is clear that the GKE should have the attribute channel set to `REGULAR` or `STABLE`, and by default, the release channels are not enabled, as shown in the screenshot's below:

image

- Taking this into account, it is clear that an GKE Cluster(represented by the resource `google_container_cluster`) should be flagged if the release channels are not defined and if the release channel defined is not either `REGULAR` or `STABLE`.
- More specifically, on the resource of type `google_container_cluster` the release channel is defined by the block `release_channel` which has a field called `channel` inside that should have the value of the relelase channel and can be defined with the values `UNSPECIFIED`, `RAPID`, `REGULAR`, `STABLE` and `EXTENDED`.
- Regarding the query implementation itself, it was composed by only one policy, which uses a helper function called `get_res`, that checks if the field `release_channel.channel` is not defined to `REGULAR` or `STABLE` or it's not defined at all.


cx-eduardo-semanas (approved):
LGTM

PR #7899, opened 2025-12-04
test(fix): Fixed the negative1 test

cx-vasco-oliveira:
- Removed the non subscription example, because there needs to be another resource other then a subscription for that to work in the first place, so the query will never reach that point.


cx-vasco-oliveira:
This pull request was a mistake, it was not meant to be to the master branch, but for the AST-116647_40_6.1.2.11_ensure_activity_log_alert_for_service_health branch instead.

PR #7907, opened 2025-12-10
feat(query): implements "Beta - Cluster Without Network Policy Support"

cx-andre-pereira:
- Currently there is no query to ensure that a "[google_container_cluster](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster)" resource allows support for a [network policy](https://kubernetes.io/docs/concepts/services-networking/network-policies/).

- Quoting CIS_Google_Kubernetes_Engine_(GKE)_Benchmark_v1.9.0 page 64: "```There are a variety of CNI plugins available for Kubernetes. If the CNI in use does not
support Network Policies it may not be possible to effectively restrict traffic in the
cluster.```

**Proposed Changes**
- The remediation suggested in the Benchmark helps bridge the gap between [CNI](https://www.cni.dev/) plugins and their compatibility with network policies and what we must actually check in the terraform IaC samples : "```To use a CNI plugin with Network Policy, enable Network Policy in GKE, and the CNI plugin will be updated.```"

- To ensure a given "google_container_cluster" resource supports use of a "network policy" it must be setup with the "[network_policy.enabled](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#enabled-7)" field set to true. The documentation seems to imply changing the "[network_policy.provider](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#provider-1)" field can be done without the need to recreate the cluster whilst the "enabled" field would require said re creation. 

- With this the query will flag the target "google_container_cluster" if it:
  - Does not set a "network_policy" block
  - Sets a "network_policy" block where "enabled" is set to a value that is not ```true``` (since "enabled" is a required field the scenario of it missing was not included in the [tests](https://github.com/Checkmarx/kics/pull/7907/files#diff-0396e8febc748855018d79f408c2ad1f4b47cb467102f7b23dfcaf802758862e))

### DIFF
--- a/assets/queries/terraform/gcp/cluster_without_network_policy_support_enabled/test/positive.tf
+++ b/assets/queries/terraform/gcp/cluster_without_network_policy_support_enabled/test/positive.tf
@@ -1,4 +1,4 @@
-resource "google_container_cluster" "negative1" {
+resource "google_container_cluster" "positive1" {
   name     = "gke-network-policy-cluster"
   location = "us-central1"
 
@@ -7,7 +7,7 @@ resource "google_container_cluster" "negative1" {
   # missing "network_policy"
 }
 
-resource "google_container_cluster" "negative2" {
+resource "google_container_cluster" "positive2" {
   name     = "gke-network-policy-cluster"
   location = "us-central1"
 


cx-eduardo-semanas (approved):
LGTM

cx-eduardo-semanas (approved):
LGTM

PR #7926, opened 2025-12-29
fix(queries): fix policy evaluation when scanning Terraform plan vs HCL files

cx-artur-ribeiro:
- KICS fails to detect policy-related security issues when scanning Terraform plan JSON files, while correctly identifying the same issues in HCL source files (.tf). 
- This inconsistency occurs because policies are represented differently in the parsed payloads:
   - In HCL source files: Policies using `jsonencode()` are parsed as escaped JSON strings (e.g., "policy": "{\"Statement\":[...]}")
   - In Terraform plan files: The same policies are already resolved as JSON objects (e.g., "policy": {"Statement": [...]})
- The current implementation in terraform.rego uses `json_unmarshal()` which only handles string inputs, causing queries to silently fail on plan files where policies are already objects. This results in various different vulnerabilities being missed when scanning Terraform plans;

**Proposed Changes**
- Update `terraform.rego` to use `common_lib.get_policy()` instead of `common_lib.json_unmarshal()` for policy parsing. The `get_policy()` function handles both string and object formats consistently;
- This ensures policy-based queries produce consistent results regardless of whether they scan HCL source files or Terraform plan JSON files.

**Note**
- There are additional queries that need the same update. That update won't be done in the context of this pull request.
- I will open a new pull request to tackle the rest of the queries that need the same update.
- A refactor to the query IAM Access Analyzer Not Enabled was carried to improve maintenance and the overall behaviour of the query. However, since the query's objective is to verify the absence of a specific resource, the `search_line` and `search_key` are still having problems detecting the same resource each time (depends on the other of the payload structure). When KICS engine is able to point to the root of the project, this query will have the expected behaviour.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM checkmarx/go:1.25.6-r0-1d9a15fffa0dba@sha256:1d9a15fffa0dba3409727b0de4a49848a02629a163709e73f55572c50d42c7de AS build_env
+FROM checkmarx/go:1.25.6-r0@sha256:8e7befa1320d506ee2e5511501196bc4afa0df731b54809a7404cdd766d3b940 AS build_env
 
 # Copy the source from the current directory to the Working Directory inside the container
 WORKDIR /app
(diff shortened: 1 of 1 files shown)


cx-eduardo-semanas (approved):
LGTM

PR #7928, opened 2025-12-31
fix(query): add supports for Microsoft.Web/sites/config on Web App Not Using TLS Last Version for AzureResourceManager

cx-ricardo-jesus:
- Currently the query `"Web App Not Using TLS Last Version"` does not take into account the scenarios where a [resource](https://learn.microsoft.com/en-us/azure/templates/microsoft.web/sites/config?pivots=deployment-language-bicep) of type `Microsoft.Web/sites/config`, which is a 'child' resource from `Microsoft.Web/sites`, does not have the field `minTlsVersion` defined or when it's defined to a version below 1.2.

**Proposed Changes**
- With these changes, the query now has two policies.
- The first policy still covers the same scenarios as before, which were the scenarios where the field `siteConfig.minTlsVersion` was not defined or when it was defined to a version that is not either `1.2` or `1.3`. The only changes in this policy are that it also searches for child resources and verifies if there are none. This is to prevent this policy from flagging the cases when the parent resource of type `Microsoft.Web/sites` does not have the field `minTlsVersion` defined, but the child resource of type `Microsoft.Web/sites/config` can have or not any vulnerable configuration, which the second policy targets.
- The second policy is for the cases when a parent resource of type `Microsoft.Web/sites` has one or more child resources of type `Microsoft.Web/sites/config` and use an helper function called `check_tls_version` to handle all three scenarios:
    - When a child resource does not have the field `minTlsVersion` defined, and the parent resource has the field defined to a version that is not either `1.2` or `1.3`.
    - When both the parent and child resource don't have the field `minTlsVersion` defined.
    - When the child resource has the field defined with a value that is not either `1.2` or `1.3`.
 - Besides all the changes mentioned above, I also made changes in order to remove the usage of the walk function, as this walk function can cause some performance degradation.
 - On the `Web App Not Using TLS Last Version`, I added the helper function web_site_resources that replaces the usage of walk() built-in function to find resources of type `Microsoft.Web/sites`.
 - The function iterates in three locations: root level resources, resources inside templates and resources nested inside other resources.
 - On the query `SQL Server Database Without Auditing`, added the helper function `sql_database_resources` that also replaces the usage of the walk() built-in function. Also added the helper function `get_outer_children_resources` that replaces the usage of the walk function inside the `get_outer_children` function.
 - Both functions in this last query support two levels of nested resources, and the helper function used in the first query only supports one level of nested resources.
 - For now, this is enough to cover the cases that the queries cover.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM checkmarx/go:1.26.0-r0@sha256:d7090ae717491522b4d39171b03b8d9ea2620164dd4dff196fa2d8fbf9dd7373 AS build_env
+FROM checkmarx/go:1.26.0-r0@sha256:d4dd4cc9866df26f1b90907d923a7c9ab3ab04c75100593c4feddc8d7650be27 AS build_env
 
 # Copy the source from the current directory to the Working Directory inside the container
 WORKDIR /app
@@ -29,7 +29,7 @@ RUN CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build \
 # Runtime image
 # Ignore no User Cmd since KICS container is stopped afer scan
 # kics-scan ignore-line
-FROM checkmarx/git:2.53.0-r0@sha256:d25c8db966ff8e39eafb3134c713bc928fe5b9edb1b55ea03b7c9f6b2112f66a
+FROM checkmarx/git:2.53.0-r0@sha256:e641ebca4adb5c2dda8055d2cd009a1220b1a3d17cac0615e9c259d5bca951d6
 
 ENV TERM xterm-256color
 


assets/queries/azureResourceManager/sql_server_database_without_auditing/test/positive5.json
cx-artur-ribeiro, line 1: The results 'search_line' for this case is "-1".

assets/queries/azureResourceManager/sql_server_database_without_auditing/test/positive7.json
cx-artur-ribeiro, line 1: The results search_line for this case is "-1" as well.

cx-ricardo-jesus: Added the query on the `azureResourceManager.yaml` file on the similarityId_transition directory to fix the `-1` value returned for the search_line

PR #7937, opened 2026-01-07
fix(bug): improve helm scanning for empty files and duplicated results

cx-miguel-dasilva:
**Proposed Changes**
- Remove empty files from Helm chart rendering that cause parsing errors
- Remove duplicated vulnerabilities during scanning for Helm charts, which may be found when a parse error occurs from double file scanning. One for the Raw k8s file and the Helm rendered file

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM checkmarx/go:1.25.6-r0@sha256:efea79da16221522f2644fea046a498bce2ee6ace505a0fea38c8e977beb3443 AS build_env
+FROM checkmarx/go:1.25.6-r0-1d9a15fffa0dba@sha256:1d9a15fffa0dba3409727b0de4a49848a02629a163709e73f55572c50d42c7de AS build_env
 
 # Copy the source from the current directory to the Working Directory inside the container
 WORKDIR /app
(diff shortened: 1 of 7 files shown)


pkg/engine/inspector.go
```diff
@@ -535,9 +537,37 @@ func (c *Inspector) DecodeQueryResults(
...
+					(*vulnerabilities)[i].SimilarityID, (*vulnerabilities)[i].QueryID)
+			}
+		}
+	}
+
+	// filter out duplicated K8s results from slice
+	filtered := make([]model.Vulnerability, 0, len(*vulnerabilities))
+	for i := 0; i < len(*vulnerabilities); i++ {
+		// Keep vulnerability if it's NOT a duplicated Helm K8s result
+		if !strings.EqualFold((*vulnerabilities)[i].Platform, kubernetesPlatformName) || (*vulnerabilities)[i].FileKind != model.KindHELM ||
```
cx-bruno-silva, line 561: Wouldn't the last condition be enough to exclude duplicate vulnerabilities?

cx-miguel-dasilva: Because this problem may affect other platforms, I think this should exclusively work for the k8s &  Helm combination for now. We do not know what may be causing this scenario for other platform results.

cx-rui-araujo (approved):
LGTM

cx-bruno-silva (approved):

PR #7942, opened 2026-01-09
fix(analyzer): add unwanted channel drain before done channel returns for correct line count

cx-artur-ribeiro:
- The analyzer's unwanted channel is buffered with capacity equal to the number of files. When all worker goroutines finish (`wg.Wait()`), the channels are closed and a done signal is sent. However, the `computeValues()` function uses a select statement that races between reading messages from unwanted and receiving the done signal.
- Once done is received, the function returned immediately, abandoning thousands of unread messages still sitting in the buffered unwanted channel. These lost exclusions meant the scan processed files that should have been excluded for huge projects.

**Proposed Changes**
- Drain the buffered `unwanted` channel before exiting;
- Update `ListSupportedCloudProviders` to match `ListSupportedPlatforms()` implementation, reducing tech dept and automatically fetch supported cloud providers from the constants;

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM checkmarx/go:1.25.6-r0@sha256:efea79da16221522f2644fea046a498bce2ee6ace505a0fea38c8e977beb3443 AS build_env
+FROM checkmarx/go:1.25.6-r0-1d9a15fffa0dba@sha256:1d9a15fffa0dba3409727b0de4a49848a02629a163709e73f55572c50d42c7de AS build_env
 
 # Copy the source from the current directory to the Working Directory inside the container
 WORKDIR /app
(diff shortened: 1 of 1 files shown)


pkg/engine/source/filesystem.go
```diff
@@ -82,7 +83,14 @@ func ListSupportedPlatforms() []string {
...
 func ListSupportedCloudProviders() []string {
-	return []string{"alicloud", "aws", "azure", "gcp", "nifcloud", "tencentcloud"}
+	cp := make([]string, len(constants.AvailableCloudProviders))
+	i := 0
+	for keys := range constants.AvailableCloudProviders {
+		cp[i] = keys
+		i++
+	}
+	sort.Strings(cp)
+	return cp
```
cx-rui-araujo: Try `maps.Keys` method from `golang.org/x/exp/maps`. Should look like this:
```
cloudProviders := maps.Keys(constants.AvailableCloudProviders)
sort.Strings(cloudProviders)
return cloudProviders
```

cx-artur-ribeiro: Updated in https://github.com/Checkmarx/kics/pull/7942/commits/12c2c804b86d8edb4654641816fd09069a1a165c

cx-eduardo-semanas (approved):
LGTM

PR #7955, opened 2026-01-26
feat(tests): improving E2E HTML report and output on remediation tests

cx-ricardo-jesus:
- Currently, when the remediation tests fail, there is no indication of which files were actually remediated. 
- This information could help in situations where the remediation tests fail, because it will be easier to see which files were remediated, decreasing the time spent analysing which scenarios are causing these tests to fail.
- The HTML reported generated when an E2E test fails, more specifically when there is an actual results vs expected result. In this scenario, a long continuous block of unorganized information regarding.
- Another problem regarding the E2E test E2E_CLI_066 was found. Before these changes, if we changed the content of the expected result, more specifically inside `e2e/fixtures/E2E_CLI_ANALYZE_RESULTS.json`, no results were output in the tests.

**Proposed Changes**
- On the `Summary` struct present inside the `utils.go` file, added the `RemediatedFiles`, which is an array of strings, meant to store the path of the files that were remediated.
- Inside the method `testRemediationQuery`, added a new map, meant to store the connections between the temporary files used to do the remediations and the original files from which those temporary files came. The key is the path to these temporary files, and the value is the path to the original file.
- The new map `tempToOriginal` will be necessary to send the path to the original file through the `RemediateFile` method inside the summary, which now takes one more argument (`original_file_name`). Within this `RemediateFile` method, this path is passed to the `writeRemediation` function to append the original filename to the new field `RemediatedFiles` in the `Summary` struct.
- Returning to the method mentioned earlier, `testRemediationQuery`, the test now returns a different message from the original, and now also mentions the files that were remediated.
- Other changes can be seen in the `remediation_test.go` and `remediate.go`, because both use the changed `RemediateFile` method that now takes one more argument.
- On the `remediation_test.go`, I passed the `filePathCopyFrom`, which is the path of the original file used for the tests.
- On the `remediate.go`, I passed an empty string, because this file handles the remediation command, and in this command, which is meant to apply a remediation to a certain file, the original file path was already sent to the `RemediateFile` method.
___
- Regarding the changes on the E2E tests, I basically made changes on the following two files: `e2e/utils/json.go` and `.github/scripts/report/main.go`. The first one writes the results of the E2E tests to a JSON file, and the second is a script that transforms the information in those JSON files and produces an HTML report.
- On `e2e/utils/json.go`, I basically added three new helper functions: `func formatPayload(payload []string)`, `func formatVulnFiles(files []map[string]interface{})` and `func toComparableFiles(queries []model.QueryResult)`. 
- The helper function `formatPayload` formats a slice of strings into a single readable string, with each line separated by a newline. Used to improve the readability of the failure message in FileCheck when the number of lines between the expected and actual output files does not match. The script will use this improved output to generate a better HTML report.
- The helper function `toComparableFiles(queries []model.QueryResult)` flattens a list of `QueryResult`, into a list of of maps, each one representing a VulnerableFile(the result from a scan in a query) enriched with it's parent QueryName so, now, the other helper function `formatVulnFiles` can take each result and pass the name of the query associated with that result.
- Regarding the changes made in the script `.github/scripts/report/main.go` I added on the TestsData struct, the field `ExpectedActual` that is of type ExpectedActual. That ExpectedActual struct has the following fields:
    - ExtraElements, which contained a list of the lines from the Expected Result and the Actual Result, that is the result of the extra elements that are given by the `require.ElementsMatch` inside the `json.go` file.
    - TestInfo, that is the information about which test is returning the error as shown below:
    
        image
   - Messages, that is, the output returned from the Messages in the require package functions used in the `json.go` file. This also stores the positive and expected results.
   - The FailOutput is the output in the test that starts with the `--- FAIL` as shown below:
       
        image
    - Both of the fields `ExtraElements` and `Messages`  have the type `ActualExpectedWithStatus` that have a list of `CodeLineStatus` that is basically a line of the output with its status. The status will be used to parse the lines that differ from the expected result and the actual result to allow them to be highlighted in the generated HTML report.
 
- In this script, I added five helper functions:
    - `cleanOutput(s string) string` cleans a line of test output by removing noise added by the go-spew library, such as type annotations (e.g. `(string)`, `(int)`, `(model.IssueType)`) and length annotations (e.g., `(len=3)`). This ensures that when lines are compared and displayed in the report, they are readable and not cluttered with internal Go type information.
   -  `extractPayloadDiffLines(failLog []string) ExpectedActual` parses the fail log output for tests that failed due to a mismatch in the number of lines between the expected and actual payload files. It extracts the expectedPayload and actualPayload sections into the Messages field, and the --- FAIL line into FailOutput, by iterating over the log lines and switching state based on known prefixes.
   - `extractExpectedActualLines(failLog []string) ExpectedActual` parses the fail log output for tests that failed due to mismatched query results (i.e. when require.ElementsMatch fails). It extracts the extra elements reported in "list A" and "list B" into the ExtraElements field, the full expected and actual query content into Messages, and the --- FAIL line into FailOutput. Each line is cleaned using cleanOutput before being stored.
   - `isDifferentNumberOfLines(failLog []string) bool` inspects the fail log to determine whether the test failed because the expected and actual output files have a different number of lines. It does so by checking if the log contains both "Expected file number of lines:" and "Actual file number of lines:" markers, which are produced by the FileCheck function in json.go.
   - `isExpectedVsActual(failLog []string) bool`  inspects the fail log to determine whether the test failed due to a query result mismatch reported by require.ElementsMatch. It checks for the presence of both "extra elements in list A:" and "extra elements in list B:" markers in the log, which are the markers used by the testify library when two slices do not match.
   - `compareMessageContent(expectedActual *ExpectedActual)` iterates over the Messages and ExtraElements fields of an ExpectedActual and compares each line at the same position between the expected and actual content. If a line differs, or if one side has more lines than the other, the differing lines are marked with Status = true. This status is later used by the HTML report template to highlight the lines that differ between the expected and actual output.
---
- To reproduce the case when there are extra elements and a comparison between an expected result and the actual result, it is only necessary to change, for example, the line in the `E2E_CLI_068_RESULT.json` in the e2e/fixtures file.
- To reproduce the scenario when there is only a different number of lines between the expected result and the actual result, a result can be deleted in a file from the e2e/fixtures directory (e.g., `E2E_CLI_032.json`).
---
- As shown below, the HTML report is clearly more readable and shows the source of the error more clearly.

- Before:
    
    image

    
    image

- After:
    image

    image

- To fix the problem mentioned above, about the analyze command E2E test, firstly, I changed this [line](https://github.com/Checkmarx/kics/pull/7955/changes/1b2f3de0031ee5eed2743cde90bcd8b2ed716ef3..abfe9bf4164d3394ca8829ea2cd7f21403c14a3f#diff-ad7234bcd572c91041d3e2832c4d042573e0f872d1a905656a83af45738a7535R84), from `if tt.Args.ExpectedAnalyzerResults != nil && arg < len(tt.Args.ExpectedResult)` to `if tt.Args.ExpectedAnalyzerResults != nil && arg < len(tt.Args.ExpectedAnalyzerResults)` in order to use the proper field inside the structure since, in the previous version of the code, the `checkExpectedResults` function was never actually executed because it was waiting for the `ExpectedResult` field of the structure to be populated and, since this test executes a analyze command, the field inside the `args` structure that was actually populated was the ExpectedAnalyzerResults not the `ExpectedResult`. Also, it is importan to note that the type of the ExpectedAnalyzrResult field has been changed from `*ResultsValidation` to `[]ResultsValidation` to allow storing multiple `ResultsValidation` objects, enabling the saving of multiple results at once, which is essential for running E2E tests that involve more than one command.
- On the `cli_test.go`, I changed the function checkExpectedAnalyzerResults to also use the FileCheck function from utils/json.go, to check the content of the results file, not just the Schema.
- Also changed the function `setFields()` used in the `FileCheck` function to have the ability to check and compare the actual and expected results from an `analyze` command, something that wasn't done before.
- Added a second analyze command to the E2E-CLI-066 test. The first command continues to analyze the swagger directory as before. The second command analyzes `positive.yaml`, a CloudFormation file. Additionally, a new CloudFormation sample file (`cloudformation_sample.yaml`) was added to the swagger directory to extend the analyze coverage in that path.


.github/scripts/report/main.go
```diff
@@ -51,6 +70,252 @@ func FindTest(tests []TestsData, testName string) (*TestsData, bool) {
...
+	return ExpectedActual{
+		TestInfo: testInfo,
+		Messages: messages,
+		FailOutput: failOutput,
+	}
+}
+
+func extractExpectedActualLines(failLog []string) ExpectedActual {
+	var extraElements ActualExpectedWithStatus
+	//var extraExpected []string // extra elements in list A
```
cx-artur-ribeiro: Remove the comments if not needed please

.github/scripts/report/main.go
```diff
@@ -51,6 +70,252 @@ func FindTest(tests []TestsData, testName string) (*TestsData, bool) {
...
+
+	for _, line := range failLog {
+		trimmed := strings.TrimSpace(line)
+		if strings.HasPrefix(trimmed, "actualPayload") {
+			state = stateMessagesActual
+		} else if strings.HasPrefix(trimmed, "expectedPayload") {
+			state = stateMessagesExpected
+		} else if strings.HasPrefix(trimmed, "--- FAIL:") {
+			state = stateFailLog
+		} else if strings.HasPrefix(trimmed, `"queries": [`) {
```
cx-artur-ribeiro: We could replace the magic strings with named constants, for example: 

Adding const in line 64 of the file: 

```go
const (
	prefixActualPayload    = "actualPayload"
	prefixExpectedPayload  = "expectedPayload"
	prefixFail             = "--- FAIL:"
	prefixQueries          = `"queries": [`
	extraElementsListA     = "extra elements in list A:"
	extraElementsListB     = "extra elements in list B:"
	suffixExpectedQueries  = "Expected Queries content: 'fixtures/{"
	suffixActualQueries    = "doesn't match the Actual Queries content: 'output/{"
	prefixTypeInterface    = "([]interface {})"
	prefixTypeVulnerable   = "(model.VulnerableFile) {"
)
```

Then on line 99:

```go
if strings.HasPrefix(trimmed, prefixActualPayload) {
	state = stateMessagesActual
} else if strings.HasPrefix(trimmed, prefixExpectedPayload) {
	state = stateMessagesExpected
} else if strings.HasPrefix(trimmed, prefixFail) {
	state = stateFailLog
} else if strings.HasPrefix(trimmed, prefixQueries) {
	state = stateNone
}
```

The same could be applied for the strings inside the switch case on line 110.

.github/scripts/report/main.go
```diff
@@ -51,6 +70,252 @@ func FindTest(tests []TestsData, testName string) (*TestsData, bool) {
...
+
+		if strings.HasPrefix(trimmed, "Test:") {
+			state = stateTestInfo
+		} else if strings.HasSuffix(trimmed, "Expected Queries content: 'fixtures/{") {
+			state = stateMessagesExpected
+		} else if strings.HasSuffix(trimmed, "doesn't match the Actual Queries content: 'output/{") {
+			state = stateMessagesActual
+		} else if strings.HasPrefix(trimmed, "--- FAIL:") {
+			state = stateFailLog
+		}
```
cx-artur-ribeiro: Same case as highlighted above, we could replace the magic strings with constants.

```go
case stateExtraA:
			if isExtraElementContentLine(trimmed) {
				cleanedLine := cleanOutput(line)
				extraElements.ExpectedContent = append(extraElements.ExpectedContent, CodeLineStatus{
					Line:   cleanedLine,
					Status: false,
				})
			}
		case stateExtraB:
			if isExtraElementContentLine(trimmed) {
				cleanedLine := cleanOutput(line)
				extraElements.ActualContent = append(extraElements.ActualContent, CodeLineStatus{
					Line:   cleanedLine,
					Status: false,
				})
			}
		case stateTestInfo:
			testInfo = append(testInfo, line)
		case stateMessagesActual:
			if !strings.HasSuffix(trimmed, suffixActualQueries) {
				messages.ActualContent = append(messages.ActualContent, CodeLineStatus{
					Line:   line,
					Status: false,
				})
			}
		case stateMessagesExpected:
			if !strings.HasSuffix(trimmed, suffixExpectedQueries) {
				messages.ExpectedContent = append(messages.ExpectedContent, CodeLineStatus{
					Line:   line,
					Status: false,
				})
			}
```

.github/scripts/report/main.go
```diff
@@ -51,6 +70,252 @@ func FindTest(tests []TestsData, testName string) (*TestsData, bool) {
...
+				})
+			}
+		case stateExtraB:
+			if !strings.HasPrefix(trimmed, "([]interface {})") && !strings.HasPrefix(trimmed, "(model.VulnerableFile) {") {
+				cleaned_extraElement_line := cleanOutput(line)
+				extraElements.ActualContent = append(extraElements.ActualContent, CodeLineStatus{
+					Line: cleaned_extraElement_line,
+					Status: false,
+				})
+			}
```
cx-artur-ribeiro, line 249: To avoid duplicated code, I would propose an auxiliar function to be called here, using the constants defined and explained on the comments above.

```go
func isExtraElementContentLine(trimmed string) bool {
	return !strings.HasPrefix(trimmed, prefixTypeInterface) && !strings.HasPrefix(trimmed, prefixTypeVulnerable)
}
```

```suggestion
		case stateExtraA:
			if isExtraElementContentLine(trimmed) {
				cleanedLine := cleanOutput(line)
				extraElements.ExpectedContent = append(extraElements.ExpectedContent, CodeLineStatus{
					Line:   cleanedLine,
					Status: false,
				})
			}
		case stateExtraB:
			if isExtraElementContentLine(trimmed) {
				cleanedLine := cleanOutput(line)
				extraElements.ActualContent = append(extraElements.ActualContent, CodeLineStatus{
					Line:   cleanedLine,
					Status: false,
				})
			}
```

.github/scripts/report/e2e-report
cx-artur-ribeiro, line 1: Why do we need the executable on the repo?

.github/scripts/report/main.go
```diff
@@ -51,6 +70,252 @@ func FindTest(tests []TestsData, testName string) (*TestsData, bool) {
...
+	var hasExpectedFileNumberLines, hasActualFileNumberLines bool
+	for _, failLogEntry := range failLog {
+		trimmedEntry := strings.TrimSpace(failLogEntry)
+		if trimmedEntry == "" {
+			continue
+		}
+		if strings.Contains(trimmedEntry, "Expected file number of lines:") {
+			hasExpectedFileNumberLines = true
+		}
+		if strings.Contains(failLogEntry, "Actual file number of lines:") {
```
cx-artur-ribeiro: We could replace the magic strings with the strings defined above in the constant block.

.github/scripts/report/main.go
```diff
@@ -51,6 +70,252 @@ func FindTest(tests []TestsData, testName string) (*TestsData, bool) {
...
+			expectedActual.Messages.ActualContent[i].Status = true
+		}
+	}
+
+	for j := range maxLenExtraElements {
+		if j >= expectedLenExtraElements || j >= actualLenExtraElements {
+			if j < expectedLenExtraElements {
+				expectedActual.ExtraElements.ExpectedContent[j].Status = true
+			}
+			if j > actualLenExtraElements {
```
cx-artur-ribeiro: if `j > actualLenExtraElements`, we never marked extra actual content as different when one list was longer. 

Shouldn't it be (like the conditions above):
 
```suggestion
			if j < actualLenExtraElements {
```

.github/scripts/report/main.go
```diff
@@ -51,6 +70,252 @@ func FindTest(tests []TestsData, testName string) (*TestsData, bool) {
...
+    }
+
+	maxLenExtraElements := expectedLenExtraElements
+	if actualLenExtraElements > maxLenExtraElements {
+		maxLenExtraElements = actualLenExtraElements
+	}
+
+	for i := range maxLen {
+		// if one of the sides does not have a line in this position, the one that exists 
+		// is tagged as the different one
```
cx-artur-ribeiro: Replace with better wording: 

```suggestion
		// If one side has no line at this index, the line that exists is marked as different.
```

.github/scripts/report/e2e-report
cx-artur-ribeiro, line 1: Why do we need this file?

PR #7994, opened 2026-03-10
feat(action): added validation for searchLine field in actions

cx-ricardo-jesus:
- Currently the GitHub actions don't validate the result from the searchLine field.

**Proposed Changes**
- Added a new job `validate-search-line` to the `go-ci` workflow, along with its supporting Python script, to automatically validate the correctness of the searchLine field in the modified KICS queries.
- On the new job I used the `dorny/paths-filter` action , to get the modified files in each PR/push to a branch, because the purpose of this actions is to validate the searchLine value for the queries changed on the relative push. This action produces the output `queries` that is a boolean that indicates whether any matching files with the path 'assets/queries/**/query.rego' (KICS query path) and the output `queries_files` that contains a JSON array with the exact paths of the changed files.
- The following steps `Set up Python`, `Set up Go` and `Validate searchLine` only run if `steps.filter.outputs.queries == true`, meaning they are entirely skipped if no query file was modified.
- The validation is made using a Python script `validate-search-line.py` that was made in Python because almost all the other scripts are also written in Python.
- The script begins by getting the path of the directory that contains the queries that were changed, using the `get_changed_queries()` helper function. This function reads the `CHANGED_QUERIES` environment variable, which is the JSON array produced by `dorny/paths-filter`. It parses the array and, for each entry ending in `/query.rego`, extracts the parent directory (the query folder) and returns of Path objects pointing to those directories.
- For each query folder, it will be used other helper function called `validate_query` will be used, which aims to validate a single directory. This helper function first of all, uses another helper function called `has_search_line_defined`, which only checks if the query contains the searchLine defined. If the query that was modified in the PR has this field defined, it starts a KICS scan command using the helper function `run_kics_scan` that only runs that command and returns True if the scan ran with no problems. After that, it uses the helper function called validate_scan_results, which reads the results from the scan that ran previously, and checks if the line is defined to `"-1"` or if the `searchLine` value is not equal to the value in the `line` field. If any of these scenarios happen, it returns False. 
- These validations mentioned above are made individually for each query; thus, if any of the queries do not have a valid value for the searchLine field in the results, it will return an error stating `::error::Some searchLine validations failed. See errors above.`.


.github/scripts/validate-search-line/validate-search-line.py
```diff
@@ -0,0 +1,166 @@
...
+        return False
+
+    return validate_scan_results(query_dir)
+
+
+def main():
+    query_dirs = get_changed_queries()
+
+    if not query_dirs:
+        print("No query.rego were changes - nothing to validate")
```
cx-artur-ribeiro: Small typo "changes" -> "changed"

```suggestion
        print("No query.rego were changed - nothing to validate")
```

.github/workflows/go-ci.yml
```diff
@@ -90,6 +90,40 @@ jobs:
...
+          filters: |
+            queries:
+              - 'assets/queries/**/query.rego'
+      - name: Set up Python
+        if: steps.filter.outputs.queries == 'true'
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        with:
+          python-version: '3.13'
+      - name: Set up Go
+        if: steps.filter.outputs.queries == 'true'
```
cx-artur-ribeiro: Why do we need the `steps.filter.outputs.queries == 'true'` in the Set up Go step here? Shouldn't it only be present on the step below?

.github/scripts/validate-search-line/validate_search_line.py
cx-artur-ribeiro, line 1: Can you change the name of the file to snake case please to align with the rest of the files present on the folder `scripts`? 
You can keep the name as `validate_search_line.py`.

.github/scripts/validate-search-line/validate-search-line.py
```diff
@@ -0,0 +1,166 @@
+#!/usr/bin/env python3
+
+import json
+import os
+import subprocess
+import sys
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parent.parent.parent.parent
```
cx-artur-ribeiro: Improve for readability please.
```suggestion
REPO_ROOT = Path(__file__).resolve().parents[3]
```
Either something along these lines or refactor to find the root in a dynamic way, using other files as reference (for example the `.gitignore` file).

Can you please check how other python scripts define the root folder or fetch the queries as well? For example, `.github/scripts/queries-validator/queries-validator.py`.

.github/scripts/validate-search-line/validate_search_line.py
```diff
@@ -0,0 +1,166 @@
...
+        print(f"  ::error file={rel_dir}::Results file not generated by scan")
+        return False
+
+    data = json.loads(results_file.read_text())
+
+    # Flatten results from all queries
+    all_results = []
+    for query in data.get("queries", []):
+        for entry in query.get("files", []):
+            all_results.append({
```
cx-artur-ribeiro: Do we want to add default values for these items? Using `line` and `search_line` as 0, as well as the `file_name` as an empty string doesn't seem like the best approach.
Maybe get the item directly without any default value.

.github/scripts/validate-search-line/validate_search_line.py
```diff
@@ -0,0 +1,166 @@
...
+    valid = True
+    for idx, r in enumerate(all_results):
+        sl = r["search_line"]
+        ln = r["line"]
+        fn = r["file_name"]
+
+        if sl == -1:
+            print(f"  ::error::Result [{idx}] {fn}: search_line is -1")
+            valid = False
+        elif sl != ln:
```
cx-artur-ribeiro, line 139: Should we add a validation for the string "-1" as well? 

```
        elif sl == "-1":
            print(f"  ::error::Result [{idx}] {fn}: search_line is '-1'")
            valid = False  
```

If we do, update the first comment on this function please.

.github/workflows/go-ci.yml
cx-miguel-dasilva: Why do you need to set up Go to run python script?

.github/scripts/validate-search-line/validate_search_line.py
```diff
@@ -0,0 +1,166 @@
...
+        "-p", str(query_dir / "test"),
+        "-o", str(results_dir),
+        "--output-name", "all_results.json",
+        "-i", query_id,
+        "-d", str(payloads_dir / "all_payloads.json"),
+        "-v",
+        "--experimental-queries",
+        "--bom",
+        "--enable-openapi-refs",
+        "--ignore-on-exit", "results",
```
cx-miguel-dasilva: Add similarityID transition flag. 

```suggestion
        "--ignore-on-exit", "results",
        "--kics_compute_new_simid",
```

.github/scripts/validate-search-line/validate-search-line.py
```diff
@@ -0,0 +1,166 @@
...
+        "--bom",
+        "--enable-openapi-refs",
+        "--ignore-on-exit", "results",
+    ]
+
+    print(f"  Running scan with query ID: {query_id}")
+
+    proc = subprocess.run(cmd, capture_output=True, text=True, cwd=str(REPO_ROOT))
+
+    if proc.returncode != 0:
```
cx-miguel-dasilva: Is this expected? KICS returns non 0 error code when it finds results. Is that taken into account?

.github/scripts/validate-search-line/validate_search_line.py
cx-miguel-dasilva: We should treat it like a JSON number on the parsing side, then we would be assured it would be a number here

docker/Dockerfile.alpine
cx-artur-ribeiro, line 1: Why change the dockerfiles go version?

.github/scripts/validate-search-line/validate_search_line.py
cx-miguel-dasilva: Why don't we use objects to read the JSON data?

PR #7995, opened 2026-03-11
fix(analyzer): improvement to dockerfile scanning

cx-andre-pereira:
The current implementation for dockerfile scanning is overly restrictive. The [GetExtension](https://github.com/Checkmarx/kics/tree/e1f23cad9640f55b963f22a116b04906b8c16ac6/pkg/utils/get_extension.go#L16) function can identify "dockerfile" type files in two ways as of now: 
  - Using the [Contains](https://github.com/Checkmarx/kics/tree/e1f23cad9640f55b963f22a116b04906b8c16ac6/pkg/utils/contains.go#L9) function on files that have [no extension set](https://github.com/Checkmarx/kics/blob/e1f23cad9640f55b963f22a116b04906b8c16ac6/pkg/utils/get_extension.go#L30), allowing detection of files that contain "Dockerfile" in a case-sensitive way, values such as: 
 ```
    - "Dockerfile"
    - "123Dockerfile"
    - "Dockerfile123" 
```
  - Checking that a file is: 
      - a valid text file and then, through the "[readPossibleDockerFile](https://github.com/Checkmarx/kics/tree/e1f23cad9640f55b963f22a116b04906b8c16ac6/pkg/utils/get_extension.go#L53C6-L53C28)" function:
        - Return true immediately if it is a file with suffix "[gitignore](https://git-scm.com/docs/gitignore)"
        - Return true if it contains a "FROM" (case-sensitive) command after skipping any commented lines "# ..."; 
    - Given all these conditions, the result for "[GetExtension](https://github.com/Checkmarx/kics/tree/e1f23cad9640f55b963f22a116b04906b8c16ac6/pkg/utils/get_extension.go#L16)" will be "```possibleDockerfile```". As a final check for the file to be included in a given scan, the analyzer's [worker](https://github.com/Checkmarx/kics/tree/e1f23cad9640f55b963f22a116b04906b8c16ac6/pkg/analyzer/analyzer.go#L417) calls the "[isDockerfile](https://github.com/Checkmarx/kics/tree/e1f23cad9640f55b963f22a116b04906b8c16ac6/pkg/analyzer/analyzer.go#L490)" function meant to ensure that only files containing the "FROM" and "RUN"(both case-sensitive) commands are scanned, files that fail this check are promptly set as "[unwanted](https://github.com/Checkmarx/kics/blob/e1f23cad9640f55b963f22a116b04906b8c16ac6/pkg/analyzer/analyzer.go#L451)".

- The way dockerfile identification is meant to work is as follows: 
  - Case-insensitive support for all "```dockerfile```", "```dockerfile.any_extension```" and "```any_name.dockerfile```" files. 
  - Check all files inside folders with the names : "```docker/```", "```dockerfile/```" or "```dockerfiles/```", also in a case-insensitive manner.
- Additionally "```.ubi8```" and "```.debian```" files should be included in the scan if they are valid docker configurations.

- The current implementation has a plethora of issues: 
  - It is case-sensitive both for "Dockerfile" identification as well as command scanning which is case-sensitive by [convention only](https://docs.docker.com/reference/dockerfile#format). 
  - Restricts dockerfile scanning to those that contain the "FROM" and "RUN" commands when the "FROM" would suffice. 
  - Does not properly skip empty lines and "[ARG](https://docs.docker.com/reference/dockerfile#understand-how-arg-and-from-interact)" type arguments when searching for the "FROM" statement in the "[readPossibleDockerFile](https://github.com/Checkmarx/kics/tree/e1f23cad9640f55b963f22a116b04906b8c16ac6/pkg/utils/get_extension.go#L53C6-L53C28)" function. 
  - Has a major redundancy by using both the "[readPossibleDockerFile](https://github.com/Checkmarx/kics/blob/e1f23cad9640f55b963f22a116b04906b8c16ac6/pkg/utils/get_extension.go#L53)" and the "[isDockerfile](https://github.com/Checkmarx/kics/blob/e1f23cad9640f55b963f22a116b04906b8c16ac6/pkg/analyzer/analyzer.go#L490)" functions. Both serve identical purposes and since the latter is only called when "readPossibleDockerFile" returns a positive result, this check should be simplified into a single function. 

**Proposed Changes**
- Reworked the "GetExtension" function so that it first :  
  - Calls the (new) "[isDockerfileExtension](https://github.com/Checkmarx/kics/pull/7995/changes#diff-e342d7fcd6b57c8f32c3a8d057b7f4b4c63e68ddfab412b5d6a18e829923533bR44)" function, here all files that are named "dockerfile.something" or "something.dockerfile" in a case-insensitive manner, or files inside the target folders mentioned before that are deemed valid dockerfile files (through "[readPossibleDockerFile](https://github.com/Checkmarx/kics/pull/7995/changes#diff-e342d7fcd6b57c8f32c3a8d057b7f4b4c63e68ddfab412b5d6a18e829923533bR85)") will be immediately determined as ".dockerfile" files. 

- The "[readPossibleDockerFile](https://github.com/Checkmarx/kics/pull/7995/changes#diff-e342d7fcd6b57c8f32c3a8d057b7f4b4c63e68ddfab412b5d6a18e829923533bR85)" function was changed to support both "ARG" commands as well as empty lines before the "FROM" command. Additionally the order of these checks was changed to enhance performance so that it is first checked if a line is [not a "FROM" command](https://github.com/Checkmarx/kics/pull/7995/changes#diff-e342d7fcd6b57c8f32c3a8d057b7f4b4c63e68ddfab412b5d6a18e829923533bR100); previously, since every line was compared to the "FROM" statement, a file with a large number of comments would take significantly longer to skim through. (tests on a dockerfile with 260 thousand comments were two times faster with this change) 

- In the [analyzer](https://github.com/Checkmarx/kics/pull/7995/changes#diff-7dc19f3d63f984f9ddd064188626e762c34bf2dc8f41f7ee0d43fa8db61cdc23) the worker was simplified, given the changes on the [utils/get_extension](https://github.com/Checkmarx/kics/pull/7995/changes#diff-e342d7fcd6b57c8f32c3a8d057b7f4b4c63e68ddfab412b5d6a18e829923533b)'s "GetExtension" function, we only need to worry about 2 results "```gitignore```" and "```.dockerfile```" , additionally the "isDockerfile" function is now gone; it was overly restrictive and served the same purpose as the "readPossibleDockerFile" function does now. 

- Unified all "dockerfile" type resource references to ".dockerfile", previously there was use of "```dockerfile```", "```Dockerfile```" and "```possibleDockerfile```", I deemed these unnecessary since they all represent the same thing, except for "possibleDockerfile" but files labeled as such could and are now determined as valid dockerfiles before being attributed a value for their "extension".  

- After initial reviews the case handling for "```gitignore```" values was simplified so that any "```.gitignore```" or any extensionless file with "```gitignore```" as a suffix will have identical flows, previously [unit-tests](https://github.com/Checkmarx/kics/pull/7995/changes#diff-67027e98d80320fd9c7c85e5510a9438add093866d1444f2210145c95c5ffd22) for the analyzer forced the latter to be added to the "[unwanted](https://github.com/Checkmarx/kics/pull/7995/changes#diff-7dc19f3d63f984f9ddd064188626e762c34bf2dc8f41f7ee0d43fa8db61cdc23L451)" path list, the test failed if we simple discarded the file in the analyze function before calling the worker (the way any invalid extension outside target folders is treated).

- ## Tests                                                                                                                                                                         
                                                                                                  
  Many new files were added to `test/fixtures/dockerfile` as well as a new folder `test/fixtures/negative_dockerfile`.                                                             
                                                                                                                                                                                   
- test/fixtures/dockerfile/                                                                                                                                                    
```                                                                                                                                         
  ├── Dockerfile-example                (pre-existing test)                               
  ├── corrupted_dockerfile              (pre-existing test)
  ├── any_name/                                                                                                                                                                                                                    
  │   ├── DOCKERfile.txt
  │   ├── Dockerfile.something                                                                                                                                                                                                     
  │   ├── any_name.debian                                                                                                                         
  │   ├── any_name.ubi8
  │   ├── dockerFILE
  │   ├── file.Dockerfile
  │   ├── file_2.DOCKERfile
  │   └── random_name
  ├── case_insensitive_tests/
  │   ├── DOCKERfile.txt
  │   ├── Dockerfile.something
  │   ├── any_name.debian
  │   ├── any_name.ubi8
  │   ├── dockerFILE
  │   ├── file.Dockerfile
  │   ├── file_2.DOCKERfile
  │   └── random_name
  ├── should_generate_payload/
  │   ├── base_image_reference
  │   ├── big_host_name
  │   ├── big_indent_from
  │   ├── full_sha_digest
  │   ├── host_ip_address
  │   ├── multiline_from_statement
  │   ├── platform_flag
  │   ├── platform_flag_and_alias
  │   ├── platform_flag_scratch
  │   ├── platform_with_arg_reference
  │   ├── use_of_host
  │   └── with_excluded_comments
  ├── test_folder_names/
  │   ├── docker/
  │   │   └── any_file.txt
  │   ├── dockerfile/
  │   │   └── any_file.txt
  │   └── dockerfiles/
  │       └── any_file.txt
  └── test_folder_names_case/
      ├── Docker/
      │   └── any_file.txt
      ├── Dockerfile/
      │   └── any_file.txt
      └── Dockerfiles/
          └── any_file.txt
```
  The files themselves include configurations with empty lines, `ARG` commands and comments before the target `FROM` command. Additionally all naming conventions are tested. Note that files inside the target folders would not be identified if they were not inside said folders (because `isDockerfileExtension` is [called](https://github.com/Checkmarx/kics/pull/7995/changes#diff-e342d7fcd6b57c8f32c3a8d057b7f4b4c63e68ddfab412b5d6a18e829923533bR44) before checking literal extension). 
  
  The `case_insensitive_tests` folder includes identical tests where all docker commands have been written in lowercase — note that most queries logic / the engine itself will not properly allow final results pointing to target lines when a query is triggered, but they still trigger similarly to comparable conventional dockerfiles (fixed in https://github.com/Checkmarx/kics/pull/8006).
  
  Lastly the "should_generate_payload" folder has 12 different **valid** variations of a docker "FROM" command to ensure the [regex used to identify dockerfiles](https://github.com/Checkmarx/kics/pull/7995/changes#diff-e342d7fcd6b57c8f32c3a8d057b7f4b4c63e68ddfab412b5d6a18e829923533bR15) do not exclude any of the samples. As the name of the folder implies all 12 samples should generate a payload ensuring the expected behavior.

- test/fixtures/negative_dockerfile/
```
  ├── CW671X02_EBM_EVENT_RULE              (pre-existing test)
  ├── not_dockerfile.debian                                                                                                                                                                                                        
  ├── not_dockerfile.txt
  ├── not_dockerfile.ubi8                                                                                                                                                                                                          
  └── should_not_generate_payload/                                                                                                           
      ├── invalid_syntax                                                                                                                                                                                                           
      ├── negative_email                                                                                                                       
      ├── negative_email_2                                                                                                                                                                                                         
      ├── negative_python
      ├── negative_python_2
      ├── negative_sparql
      └── negative_sql
```
  The `.debian` and `.ubi8` files here should not be flagged as valid dockerfiles since they do not correspond to valid docker configurations. `not_dockerfile.txt` should not be flagged since, although it contains a valid docker configuration, it has a `.txt` extension and is not named "dockerfile" (case-insensitive) explicitly — it only contains the word "dockerfile". 
  
The `should_not_generate_payload` samples test extensionless files containing `FROM` statements in non-Docker contexts (email header, Python imports, SPARQL queries, SQL) to ensure they are not misidentified as dockerfiles.

 These files are used for the [`util/get_extension` unit tests](https://github.com/Checkmarx/kics/pull/7995/changes#diff-beda894e6e1f38957b33086728fadab666b91b5a3d22817f32b9678232ddcbcd) and the new [E2E test 106](https://github.com/Checkmarx/kics/pull/7995/changes#diff-e987437b21739e745091de2fc87f9e42360a00f6e0479f2ebf3d3560e3a2c223).
 
 As a final note for regarding the tests I should point out that the [regex used to identify/exclude samples](https://github.com/Checkmarx/kics/pull/7995/changes#diff-e342d7fcd6b57c8f32c3a8d057b7f4b4c63e68ddfab412b5d6a18e829923533bR15) could never prevent false positives on Files with basic "```FROM <any_string>```" commands, often related to SQL syntax, in case the files with this text are either: 
 - Files without extension 
 - Files stored inside the target ("```docker/```", "```dockerfile/```" or "```dockerfiles/```") folders 

Moreover a new function, "[TestParser_Parse_CaseInsensitive](https://github.com/Checkmarx/kics/pull/7995/changes#diff-f5568accfdb4b8da27afe3045504007ecfe7aff76b6953a74a2521edada45a34R241)", was added to the [parser/docker/parser_test.go](https://github.com/Checkmarx/kics/pull/7995/changes#diff-f5568accfdb4b8da27afe3045504007ecfe7aff76b6953a74a2521edada45a34) unit tests; as the name implies it is meant to ensure docker samples are parsed identically regardless of casing in the [dockerfile commands](https://docs.docker.com/reference/dockerfile/) syntax.


### CODE REVIEW

File: pkg/utils/get_extension.go [hunk 1]
```diff
@@ -14,9 +14,7 @@ import (
 
 // GetExtension gets the extension of a file path
 func GetExtension(path string) (string, error) {
-	targets := []string{"Dockerfile", "tfvars"}
-
-	// Get file information
+	extDockerfile := ".dockerfile"
```
cx-miguel-dasilva: Convert the variable into a constant; by doing so, you don't need to send it as an argument to the following functions

cx-miguel-dasilva (changes requested):

pkg/utils/get_extension.go
cx-andre-pereira: Done in https://github.com/Checkmarx/kics/pull/7995/commits/ad57a0c5573b6089b47454df1add39344c948dd4

e2e/testcases/e2e-cli-106_valid_dockerfile_detected.go
```diff
@@ -0,0 +1,31 @@
+package testcases
+
+// E2E-CLI-106 - KICS  scan
+// should perform the scan successfully detect all valid dockerfile documents and return result 50
```
cx-artur-ribeiro: Either add "and" before detect or use "detecting" please. Also, we could use files instead of "documents" but I'm okay with either approach.

cx-artur-ribeiro (changes requested):
Nothing to point out besides that.

e2e/testcases/e2e-cli-106_valid_dockerfile_detected.go
cx-andre-pereira: Done in https://github.com/Checkmarx/kics/pull/7995/commits/4af5804f8121bfc607791258c9741101a8958682  with "detecting" and files instead of "documents".

pkg/parser/docker/parser.go
```diff
@@ -133,7 +133,7 @@ func (p *Parser) GetKind() model.FileKind {
 
 // SupportedExtensions returns Dockerfile extensions
 func (p *Parser) SupportedExtensions() []string {
-	return []string{"Dockerfile", ".dockerfile", ".ubi8", ".debian", "possibleDockerfile"}
```
cx-rui-araujo, line 136: Do we still need `.ubi8` and `.debian`?
You are replacing them to `.dockerfile` in GetExtension func.

pkg/remediation/scan.go
```diff
@@ -95,7 +95,7 @@ func getPayload(filePath string, content []byte, openAPIResolveReferences bool,
 	var err error
 
 	switch ext {
-	case ".dockerfile", "Dockerfile", "possibleDockerfile", ".ubi8", ".debian":
+	case ".dockerfile", ".ubi8", ".debian":
```
cx-rui-araujo: Same as the previous comment.
Do we still need `.ubi8` and `.debian`?

cx-rui-araujo (commented):
Hi @cx-andre-pereira, pls check me comments

pkg/parser/docker/parser.go
cx-andre-pereira: I left them as a precaution but they are in fact unnecessary at this point there will never be a case where GetExtension returns those values when they actually represent dockerfiles, the parser was generating even when the debian and ubi8 files were not valid dockerfiles. Done in https://github.com/Checkmarx/kics/pull/7995/commits/7ee77cc0601c68824e9f1ccc8cf033a1dba74053.

pkg/remediation/scan.go
cx-andre-pereira: Updated in https://github.com/Checkmarx/kics/pull/7995/commits/7ee77cc0601c68824e9f1ccc8cf033a1dba74053

cx-artur-ribeiro (dismissed):
LGTM, nice job André.

pkg/utils/get_extension.go
```diff
@@ -68,37 +83,14 @@ func readPossibleDockerFile(path string) bool {
...
 	// Read lines from the file
 	for scanner.Scan() {
-		if strings.HasPrefix(scanner.Text(), "FROM") {
-			return true
-		} else if strings.HasPrefix(scanner.Text(), "#") {
+		if strings.HasPrefix(scanner.Text(), "#") || strings.HasPrefix(strings.ToLower(scanner.Text()), "arg") || scanner.Text() == "" {
 			continue
 		} else {
-			return false
+			UrlRegex := `[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)`
```
cx-rui-araujo: Can you add this regex as consts?

pkg/parser/docker/parser.go
```diff
@@ -59,7 +59,7 @@ func (p *Parser) Parse(_ string, fileContent []byte) ([]model.Document, []int, e
 	for _, child := range parsed.AST.Children {
 		child.Value = strings.ToLower(child.Value)
 		if child.Value == "from" {
-			fromValue = strings.TrimPrefix(child.Original, "FROM ")
+			fromValue = child.Original[5:]
```
cx-rui-araujo, line 62: Is this gonna work if the line that contains `from` has white spaces before it?
For example, `"    from"`

cx-rui-araujo (commented):
@cx-andre-pereira check my comments pls

pkg/parser/docker/parser.go
cx-andre-pereira: The `child.Original` value already comes pre-trimmed so there is no need to remove white spaces.

pkg/utils/get_extension.go
cx-andre-pereira: Done in https://github.com/Checkmarx/kics/pull/7995/commits/2051586f1699bbb548a6fd8279e6a4d35acf2b37 .

cx-rui-araujo (dismissed):
LGTM, nice work @cx-andre-pereira 💪

janderssonse:
This would be great, any hope of seeing it moving along be rebased and move into main?

PR #8014, opened 2026-03-26
fix(workflows): update github actions to use SHA

cx-bruno-silva:
Closes #

- Github actions using tag instead of SHA

**Proposed Changes**
- Update all github actions to use SHA


cx-rui-araujo (approved):
LGTM

PR #8028, opened 2026-04-06
fix(validator): update queries validator for cwe and risk score fields

cx-artur-ribeiro:
- `cwe` field is not just a string, it's a numbered string, which was not being validated correctly.
- `riskScore` is a float number with only one decimal case;

**Proposed Changes**
- Improve queries validator script to validate cwe as a numbered string instead of just a string and float with one decimal place for riskScore;
- Fix bug in url pagination;
- Update relevant documentation with proper syntax and more information about the fields;


.github/scripts/queries-validator/metadata-schema.json
```diff
@@ -21,6 +21,16 @@
...
+        },
+        "cwe_pattern": {
+            "type": "string",
+            "minLength": 1,
+            "pattern": "^[0-9]+$"
+        },
+        "risk_score_pattern": {
+            "type": "string",
+            "minLength": 1,
+            "pattern": "^[0-9]+\\.[0-9]$"
```
cx-andre-pereira:
```suggestion
            "pattern": "^[0-9]\\.[0-9]$"
```
I don't think the ```+``` is warranted here; it implies patterns over 9.9 could be possible. 

Even if 10.0 is possible (could not find reference to confirm or deny, current highest is ```9.4``` - 6 queries), a better pattern could be ```"pattern": "^([0-9]\\.[0-9]|10\\.0)$"```

The [queries_content_test.go](https://github.com/Checkmarx/kics/blob/master/test/queries_content_test.go) seems to be the only test done on this field, it states that negative values are possible, I highly doubt that applies to the way we use the severity at least currently, also it accepts any amount of digits just like this regex, values like 100.0 are seemingly "allowed"; the restriction to 10(or 9.9) and below is only implicit, never enforced or detailed in any way as far as I can tell from our relevant [documentation](docs/creating-queries.md). 

I don´t know if i missed something on the documentation side but at least dropping the ```+``` seems worthwhile regardless. Let me know what you think.

assets/queries/ansible/aws/alb_listening_on_http/metadata.json
cx-andre-pereira, line 1: Is this change and the equivalent in ```assets/queries/ansible/aws/ami_not_encrypted/metadata.json``` accidental ? 

From what i checked only **47/1737** metadata files have a \n at the end, the pattern seems to be **not** having this newline.

docs/creating-queries.md
cx-andre-pereira, line 1: I don t know what the extra spacing on docs/creating-queries.md was for but it does not change anything visually; a visual change is very warranted though since the [Query Development Tutorial section on the official docs](https://docs.kics.io/latest/creating-queries/#query_development_tutorial) could use some separation between "```query.rego```" , "```metadata.json```" and "```test```". 


I get pushing the **riskScore** values to a newline with a hyphen but it could be deeper inline (4 white spaces vs 2, [`category` section](https://docs.kics.io/latest/creating-queries/#guidelines) uses 4 already to indent each possible value).  As it is it looks like it represents a whole new field instead of possible/recommended values, also maybe one severity per line, for improved readability, could help. 

**Comparison** (left includes changes I mentioned/ right is as is in this PR at the moment):  

Image

Master Branch : 

Image

.github/scripts/queries-validator/metadata-schema.json
cx-artur-ribeiro: Hi @cx-andre-pereira,
Thanks for taking a look at this pull request 🙏.

Risk score values must be over 0.0 and under 10.1, so you are right to question and signal the regex pattern, `+` should not be present.
I will improve the regex to only 2 digits, float-like values, as it should.

assets/queries/ansible/aws/alb_listening_on_http/metadata.json
cx-artur-ribeiro: These changes were to trigger the validation since we need to change metadata.json files in order for the action to run. Nothing was changed.

docs/creating-queries.md
cx-artur-ribeiro: The objective was to accomplish the left side comparison you provided, I'll make the necessary changes, thanks for the warning!

cx-artur-ribeiro: After the changes on latest commit, the documentation is as such: 

image

cx-andre-pereira (approved):
LGTM !

cx-rui-araujo (approved):
LGTM

cx-bruno-silva (approved):
LGTM

cx-julio-silva (approved):

PR #8030, opened 2026-04-08
fix(query): fix FP results on "IAM policy allows for data exfiltration" CloudFormation and Terraform queries

### LINKED ISSUE #7960: bug(terraform): iam_policy_allows_for_data_exfiltration  is too restrictive
iam_policy_allows_for_data_exfiltration  is too restrictive,   The `is_illegal()` function only evaluates the Action, not the Resource. So any IAM statement containing `secretsmanager:GetSecretValue` will trigger a violation, regardless of whether you specify:
•	A specific secret ARN
•	A wildcard resource
•	Condition keys   

### Actual Behavior

This should comply but it doesn’t 

data "aws_iam_policy_document" "compliant" {
  statement {
    effect = "Allow"
    actions = [
      "secretsmanager:GetSecretValue",
      "secretsmanager:DescribeSecret"
    ]
    resources = [
      "arn:aws:secretsmanager:us-east-1:123456789012:secret:my-app-secret-abc123",
      "arn:aws:secretsmanager:us-east-1:123456789012:secret:my-db-password-xyz789"
    ]
  }
}

cx-andre-pereira:
Closes #7960

- The current implementation of the 2 queries : 
  - [terraform/aws/iam_policy_allows_for_data_exfiltration](https://github.com/Checkmarx/kics/blob/0d393c145336cc9a314fdbc5f8d3dbef0e4b4b2c/assets/queries/terraform/aws/iam_policy_allows_for_data_exfiltration/query.rego)
  - [cloudformation/aws/iam_policy_allows_for_data_exfiltration](https://github.com/Checkmarx/kics/blob/6151ae02ecd5a8e955b42444ea39c8272121b7d5/assets/queries/cloudFormation/aws/iam_policy_allows_for_data_exfiltration/query.rego)
- Is not restrictive enough leading to many **False Positive** results. 
- Both in the Terraform query's [```aws_iam_policy```](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy) and the CloudFormation query's 5 target "```AWS::IAM::...```" type resources the "Resource" field is never checked to ensure exposition is global, other queries on both platforms already perform this exact check like "[```iam_policy_grants_full_permissions```](https://github.com/Checkmarx/kics/blob/0d393c145336cc9a314fdbc5f8d3dbef0e4b4b2c/assets/queries/terraform/aws/iam_policy_grants_full_permissions/query.rego#L16)" for Terraform and the "[```iam_policies_with_full_privileges```](https://github.com/Checkmarx/kics/blob/6151ae02ecd5a8e955b42444ea39c8272121b7d5/assets/queries/cloudFormation/aws/iam_policies_with_full_privileges/query.rego#L15)" for CloudFormation. 

- Without this check there is no reason to flag a given sample, it must expose all resources for the target [vulnerability](https://github.com/Checkmarx/kics/blob/0d393c145336cc9a314fdbc5f8d3dbef0e4b4b2c/assets/queries/terraform/aws/iam_policy_allows_for_data_exfiltration/metadata.json#L6) to be valid. On the relevant [Cloudsplaining documentation](https://cloudsplaining.readthedocs.io/en/latest/glossary/data-exfiltration/) this is inferred by the ```without resource constraints``` excerpt. 

- When I originally implemented these queries in https://github.com/Checkmarx/kics/pull/7631 all samples contained the relevant "```Resource: "*"```" field/value but the queries themselves would not check for this, the negative tests also did not include samples for the scenario of missing said field/value, it was simply unaccounted for.  

**Proposed Changes**
- Implemented the missing check, identical to similar queries : ```common_lib.equalsOrInArray(statement.Resource, "*")```, with the exception of the Terraform's "data" resources where the target field is slightly different : ```common_lib.equalsOrInArray(statement.resources, "*")```

- Added negative tests to show the relevant check is being done, the new negative tests avoid triggering the query by having a sample with a "Resource" value of ```arn:aws:dynamodb:us-east-1:123456789012:table/Users``` over the permissive wildcard ```*```. 

- Slight changes to positive tests were done to show that array search is supported for both queries:
  - CloudFormation - [positive5.yaml](https://github.com/Checkmarx/kics/pull/8030/changes#diff-2fb63e742de2755d59d7b31b5222c0737204736395a6ed1a87cae9854074b97c) and [positive12.json](https://github.com/Checkmarx/kics/pull/8030/changes#diff-5d92e5c953545b6b133ae6fa7e6fd36bc6ac806476eb466c95bee98f8d9a1801)
  - Terraform - [positive7.tf](https://github.com/Checkmarx/kics/pull/8030/changes#diff-91ffd5d02554e7d59c99ba49dd51e29e0443adc18c53a42405e861e441ad6cf9)

- I made the choice to test these only on positive tests since the only way to know the check is working for the array search is to have the query flag precisely because it managed to check that inside the array the wildcard target was included.  


cx-artur-ribeiro (approved):
LGTM!

PR #8033, opened 2026-04-09
fix(query): the "Image Version Using Latest" dockerfile query was not accounting for specific digest values

cx-andre-pereira:
- Currently the "[```Image Version Using Latest```](https://github.com/Checkmarx/kics/tree/53dbe657bc61a8ba4b4c7e60b2f5b5c4c2ffc49a/assets/queries/dockerfile/image_version_using_latest)" query does not account for digest values appended to the "```:latest```" tag; causing a FROM statement like : ```FROM 923847651029.dkr.ecr.eu-west-1.amazonaws.com/echo/echo-fips:latest@sha256:a1f3c2e9b0d74682f5a3c1e8b2d09f47c6e1a3b8d2f04c7e9a1b3d5f7c2e4a6b8```
to produces a **False Positive** result.

**Proposed Changes**
- Made a small adjustment to the query to ensure the target "```:latest```" tag is the suffix of the reference image being used through a [regex](https://github.com/Checkmarx/kics/pull/8033/changes#diff-bf4c4895ea586bd363ced8a135934791583956c16ae87f5ab55fd7916011a396L7) enforcing **end of line** (```$```) after any trailing whitespaces that may exist.
- Added a new sample in the [negative.dockerfile](https://github.com/Checkmarx/kics/pull/8033/changes#diff-b30e25fd2772ca8b0adc74dadb32af9ea4ba432222e671c3f6ada13eeb6642e2) test.


assets/queries/dockerfile/image_version_using_latest/query.rego
```diff
@@ -4,7 +4,7 @@ CxPolicy[result] {
 	resource := input.document[i].command[name][_]
 	resource.Cmd == "from"
 	not resource.Value[0] == "scratch"
-	contains(resource.Value[0], ":latest")
+	regex.match(":latest\\s*$", resource.Value[0])
```
cx-artur-ribeiro: Wouldn't it be the same to use rego's builtin function "endswith", since the regex is built the same?

```suggestion
	endswith(resource.Value[0], ":latest")
```

Took that idea from https://github.com/Checkmarx/kics/blob/bc5fb995ced966b95c523b686925139f567a187c/assets/queries/k8s/invalid_image/query.rego#L30.

cx-artur-ribeiro (changes requested):
Hey André, can you take a look at my comment please?

assets/queries/dockerfile/image_version_using_latest/query.rego
cx-andre-pereira: That approach is cleaner for sure, from my tests it works just the same indeed, change done in https://github.com/Checkmarx/kics/pull/8033/commits/3795843f8fcac55caf2dc30c3dfc084ca29dbb69.

cx-artur-ribeiro (approved):
LGTM

cx-artur-ribeiro (dismissed):
LGTM

cx-alon-rosenhek (dismissed):

cx-artur-ribeiro (approved):
LGTM

cx-alon-rosenhek (approved):

PR #8059, opened 2026-05-21
fix(query): skip Terraform references in metadata label validation

### LINKED ISSUE #7944: False Positive: Metadata Label Is Invalid (bc3dabb6) for Terraform local variable interpolation
**Query ID**: `bc3dabb6-fd50-40f8-b9ba-7429c9f1fb0e`
**Query Name**: Metadata Label Is Invalid
**Severity**: LOW
**Platform**: Terraform

## Problem Statement

KICS incorrectly flags Kubernetes service labels as invalid when using Terraform `local.*` variable interpolation, even though the resolved values are valid according to Kubernetes label syntax rules.

## Expected Behavior

KICS should either:
1. Resolve Terraform variables before validating label values, OR
2. Skip validation when label values contain variable interpolations, OR
3. Document this limitation in the query description

## Actual Behavior

KICS validates the literal string `local.variable_name` against the regex pattern without resolving the Terraform variable.

## Minimal Reproducible Example

```terraform
# variables.tf
variable "resource_prefix" {
  type = string
  default = "my-app-"
}

variable "name" {
  type = string
  default = "service"
}

locals {
  resource_name = "${var.resource_prefix}${var.name}"
}

# service.tf
resource "kubernetes_service_v1" "example" {
  metadata {
    name      = "my-service"
    namespace = "default"
    labels = {
      app = local.resource_name  # KICS flags this as invalid
    }
  }

  spec {
    selector = {
      app = local.resource_name
    }

    port {
      port        = 80
      target_port = 8080
    }
  }
}
```

## KICS Output

```
Metadata Label Is Invalid, Severity: LOW, Results: 1

Expected: kubernetes_service_v1[example].metadata.labels[app] has valid label
Actual: kubernetes_service_v1[example].metadata.labels[app] has invalid label
```

## Analysis

The query uses this regex: `^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?$`

The literal string `local.resource_name` fails validation, but when resolved it becomes `"my-app-service"` which is perfectly valid per Kubernetes RFC 1123.

## Impact

Affects projects using Terraform variables (`local.*`, `var.*`, `data.*`) in Kubernetes labels.

## Environment

- KICS Version: 2.1.19
- Platform: Terraform
- Provider: hashicorp/kubernetes

## Suggested Fix

Skip validation when label values contain Terraform interpolation syntax or use Terraform plan output for validation.

## Related

Similar to #591 regarding variable interpolation support.

omribz156:
Closes #7944

- The `Metadata Label Is Invalid` query currently validates unresolved Terraform references such as `local.resource_name` as literal Kubernetes label values.
- That creates a false positive before Terraform resolves the value to a valid label.

**Proposed Changes**
- Skip metadata label validation when the value is a Terraform `local.*`, `var.*`, or `data.*` reference/interpolation.
- Add a negative Terraform fixture with a Kubernetes service label using `local.resource_name`.

**Verification**
- Before the fix, direct scan of the negative fixture returned one `Metadata Label Is Invalid` result at the `labels` block.
- After the fix, direct scan of the negative fixture returned `TOTAL: 0`.
- Direct scan of the existing positive fixture still returned `TOTAL: 1` for `g**dy.l+bel`.
- `git diff --check` passed with local CRLF warnings only.

Focused Go query harness timed out locally on Windows, so I used direct CLI scans for this query.

This was implemented with Codex assistance, with the final patch kept focused and manually reviewed.

### DIFF
--- a/assets/queries/terraform/kubernetes/metadata_label_is_invalid/query.rego
+++ b/assets/queries/terraform/kubernetes/metadata_label_is_invalid/query.rego
@@ -8,6 +8,7 @@ CxPolicy[result] {
 
 	labels := resource[name].metadata.labels
 
+	not is_terraform_reference(labels[key])
 	regex.match("^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?$", labels[key]) == false
 
 	result := {
@@ -20,4 +21,10 @@ CxPolicy[result] {
 		"keyActualValue": sprintf("%s[%s].metada.labels[%s] has invalid label", [resourceType, name, key]),
 		"searchLine": common_lib.build_search_line(["resource", resourceType, name, "metadata"], ["labels", key]),
 	}
+}
+
+is_terraform_reference(label) {
+	regex.match("^\\$\\{(local|var|data)\\.[^}]+\\}$", label)
+} else {
+	regex.match("^(local|var|data)\\.[A-Za-z0-9_][A-Za-z0-9_.-]*$", label)
 }
--- a/assets/queries/terraform/kubernetes/metadata_label_is_invalid/test/negative.tf
+++ b/assets/queries/terraform/kubernetes/metadata_label_is_invalid/test/negative.tf
@@ -54,3 +54,29 @@ resource "kubernetes_pod" "test2" {
     dns_policy = "None"
   }
 }
+
+locals {
+  resource_name = "my-app-service"
+}
+
+resource "kubernetes_service_v1" "test3" {
+  metadata {
+    name      = "terraform-service-example"
+    namespace = "default"
+
+    labels = {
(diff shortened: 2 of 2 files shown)


cx-artur-ribeiro:
Hi @omribz156,
Nice contribution(s), thanks in advance for the awesome work!

Although I believe the fix you proposed was going to silently get rid of the false positives, supporting terraform locals would be the best approach in my opinion.

This PR - https://github.com/Checkmarx/kics/pull/7946 - aims to achieve that by replicating the behaviour of terraform's locals resolver. You can check more information on the PR description and the tagged issue.

I'm not sure if both solutions can coexist completely, but part of it may be a good fit.

Let me know what you think!

PR #8085, opened 2026-07-27
fix(parser): fixed looping behavior triggered during dockerfile parsing 

### LINKED ISSUE #7257: bug(docker): OOM when scanning a Dockerfile with multi-line commands and comments
Hey,
When scanning a Dockerfile with multi-line `RUN` commands that contain comments, the scan hangs until the process eventually runs out of memory.

### Expected Behavior
KICS should scan the Dockerfile without hanging or running out of memory.

### Actual Behavior
[debug.log](https://github.com/user-attachments/files/17005888/debug.log)
KICS stops printing trace logs after a few seconds and hangs, leading to the container running out of memory and crashing.

1.  Download "Dockerfile.txt" and rename it to "Dockerfile"
[Dockerfile.txt](https://github.com/user-attachments/files/17005892/Dockerfile.txt)
2.  Run KICS: `docker run -m 2G -it -v $PWD:/path checkmarx/kics:latest scan --path "/path/" --ci --log-level TRACE`
(the docker image is used to avoid running out of host memory)
3.  wait for the container to run out of memory and crash

### Specifications
-   Version: v2.1.2
-   Platform: Linux AMD64/MacOS ARM64
-   Subsystem: Docker

cx-andre-pereira:
Closes #7257

- The issue as described in #7257, was causing [OOM](https://en.wikipedia.org/wiki/Out_of_memory) issues due to unnoticed looping logic during dockerfile parsing. 

- Currently when dockerfiles containing  particularly long RUN commands, such as [dockerfile,txt](https://github.com/user-attachments/files/17005892/Dockerfile.txt), the current parsing logic of [docker_detect](https://github.com/Checkmarx/kics/blob/8ec0fa71ab2eb0195dd46a94e1d43a11d2af2ab3/pkg/detector/docker/docker_detect.go) wrongfully replace instances of “```\```” in the middle of the string and cause an exponential growth of the parsed command's length, depending on number for "```\```" characters and total lines of the command. The logic was faulty since it relied on a [regex replacement](https://github.com/Checkmarx/kics/blob/8ec0fa71ab2eb0195dd46a94e1d43a11d2af2ab3/pkg/detector/docker/docker_detect.go#L99) that was too broad. 

- Aside from causing incorrect parsing on any **multi-line RUN** command, any command with enough lines would result in an [OOM](https://en.wikipedia.org/wiki/Out_of_memory) error / "infinite" hang. 

**Proposed Changes**
- To fix the issue a [targeted removal](https://github.com/Checkmarx/kics/commit/b4081e9a182f25a9f02d14be899d598e193e4354#diff-c730cc48d70c9c17f09a8589733aec81f25e90f22aff22ab329f466f8825cfc9L99) of the last character of each line is done.

- Additionally some minor changes/fix were done to 2 queries: 
  - ["Unpinned Package Version in Apk Add"](https://github.com/Checkmarx/kics/pull/8085/changes#diff-b72a8ac5d0243da44710ef26456bb4b151dee5944dcdaa900ec78d8af9016516) - simply adjusted an unnecessarily named variable to be "```_```".
  - ["Image Version Not Explicit"](https://github.com/Checkmarx/kics/pull/8085/changes#diff-ac98493be19dd6e7c674a2ee0761276fe160b9faac935b09b3cf6761adf513ff) - fixed some significant logic errors, first the [use of a "i" variable as if it where global](https://github.com/Checkmarx/kics/pull/8085/changes#diff-ac98493be19dd6e7c674a2ee0761276fe160b9faac935b09b3cf6761adf513ffL19), this can cause other documents in a given project to influence the query and result in ```False Positives```; A similar issue exists on [line 47  ](https://github.com/Checkmarx/kics/pull/8085/changes#diff-ac98493be19dd6e7c674a2ee0761276fe160b9faac935b09b3cf6761adf513ffR47) where a given command could pass the "```build_name_exists```" check while a different command of the same name but executed in an earlier line could pass the ```"command[name][_].EndLine < line"``` check and result in a ```False Positive```, note the use of the "```i```" variable coming into play here possible making the logic extremely faulty for this section given the 2 logic flaws at play.
- Both query changes are unrelated to the issue other than the fact that, with the fix on the loop, they trigger on an internally provided sample.


cx-artur-ribeiro (approved):
LGTM

cx-alon-rosenhek (approved):

PR #8086, opened 2026-07-27
fix(query): fixed FP on default azure storage account network is too permissive for arm

cx-ricardo-jesus:
**Reason for the Proposed Changes**:
-  Currently the query does not support the scenarios when the fields `publicNetworkAccess` and `networkAcls.defaultAction` are defined inside parameters.

**Proposed Changes**:
- The only changes on the query was on the `publicNetworkAccessEnabled` and `aclsDefaultActionAllow` helper functions.
- For both I added to return a something when the fields `publicNetworkAccess` and `networkAcls.defaultAction`are defined inside parameters using the `getDefaultValueFromParametersIfPresent`  helper function from AzureResourceManager library under `assets/libraries` when the respective field's are referencing parameters (checked by using the helper function `isParameterReference` from AzureResourceManager library under `assets/libraries`).
- Added new tests to cover the cases were the fields are defined inside parameters.

### DIFF
--- a/assets/queries/azureResourceManager/default_azure_storage_account_network_access_is_too_permissive/query.rego
+++ b/assets/queries/azureResourceManager/default_azure_storage_account_network_access_is_too_permissive/query.rego
@@ -1,6 +1,7 @@
 package Cx
 
 import data.generic.common as common_lib
+import data.generic.azureresourcemanager as arm_lib
 
 CxPolicy[result] {
 	doc := input.document[i]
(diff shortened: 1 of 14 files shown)


cx-artur-ribeiro (changes requested):
Add a sample with 'app' to guarantee the pass behaviour is covered please.

cx-artur-ribeiro (dismissed):
LGTM

cx-artur-ribeiro (approved):
LGTM

cx-alon-rosenhek (approved):

PR #8087, opened 2026-07-27
fix(query): fixed FP for sql server database without auditing for arm

cx-ricardo-jesus:
**Reason for the Proposed Changes**:
-  Currently, when a Microsoft.Sql/servers/databases/auditingSettings resource is declared as a root-level resource and linked to its databases/servers parent only through dependsOn/resourceId(), with resource names built using ARM template functions such as parameters() or format(), the query returns a false positive, since it is unable to match the two resources.

**Proposed Changes**:
- To tackle the problem mentioned above, I created new helper functions that will be described below:
- `get_dependency_token(resource):` builds the string used to search for a resource's identity inside another resource's dependsOn entries. If resource.name is an ARM parameter reference (e.g. "[parameters('DatabaseName')]"), it returns the bare parameters('DatabaseName') call, since that's the part that still shows up unchanged inside a `dependsOn/resourceId()` string even when the name itself is built with format(); otherwise, it returns the literal resource.name.
- That dependencyToken variable produced by the helper function mentioned on the point above, will be taken as an argument to the other new helper function_called `get_dependent_resources`, which will use it to check if a determined resource is dependent from another "father" resource. This `get_dependent_resources` helper function, uses another two new helper functions called `leaf_type` and `type_referenced_in_dependency`.
- `leaf_type(resourceType):` returns the last segment of a resource type (e.g. "Microsoft.Sql/servers/databases" → "databases").
- `type_referenced_in_dependency(dependsOnEntry, resourceType):` checks whether a dependsOn entry references a resource of resourceType, by looking for that type's leaf segment immediately followed by a closing quote (/servers' or ('servers') inside the entry. Called from `get_dependent_resources` and from `is_child_database`.

### DIFF
--- a/assets/queries/azureResourceManager/sql_server_database_without_auditing/query.rego
+++ b/assets/queries/azureResourceManager/sql_server_database_without_auditing/query.rego
@@ -18,13 +18,6 @@ CxPolicy[result] {
 	childrenArr := [x | x := childrenArr_full[ch_index]
 						childrenArr_full[ch_index].type == types[_]]
 
-	count([x |
-		child := childrenArr[_]
-		[val, _] := arm_lib.getDefaultValueFromParametersIfPresent(doc, child.properties.state)
-		lower(val) == "enabled"
-		x := child
-	]) == 0
-
 	# Case of "child" database resource with auditingsettings as "brother" resource
 	depth_path := array.slice(path, 0, count(path)-1)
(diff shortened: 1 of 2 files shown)


cx-artur-ribeiro (approved):
LGTM

cx-alon-rosenhek (approved):

PR #8088, opened 2026-07-27
fix(query): fixed FP for website azure active directory disabled for arm

cx-ricardo-jesus:
**Reason for Proposed Changes**.
- Currently the query returns a positive result when the field `identity.type` is defined to `SystemAssigned, UserAssigned` or vice versa.

**Proposed changes**:
- Added an extra verification on `is_valid_identity` helper function and a negative test in `.json` and `.bicep` format to support the scenario.


cx-artur-ribeiro (changes requested):
Missing sample to represent the new cases.

cx-artur-ribeiro (approved):
LGTM

cx-alon-rosenhek (approved):

PR #8091, opened 2026-07-27
fix(query): fixed FP for web app not using tls last version

cx-ricardo-jesus:
- Currently, the query `Web App Not Using TLS Last Version` flags resources of type `Microsoft.Web/sites` that do not defined `properties.siteConfig.minTlsVersion` as `1.2` or `1.3` for any configurations of that resource type but, we should only flag the "web" configuration as shown on the following [documentation](https://learn.microsoft.com/en-us/azure/templates/microsoft.web/sites/config-web?pivots=deployment-language-bicep#options-for-name-property).

**Proposed Changes**
- Using the helper function `is_web_config_name`, it is now confirmed if the name of the resource is web.
- Implemented two extra helper functions, which are `get_children` and `format_config_children`. Basically, these helper functions just extend the get_children helper function from the AzureResourceManager library under `assets/libraries` to fetch the children that use the same format as the example below in the name field: 
    - `"name": "[format('{0}/{1}', <fatherResourceName>, 'web')]",`

### DIFF
--- a/assets/queries/azureResourceManager/web_app_not_using_tls_last_version/query.rego
+++ b/assets/queries/azureResourceManager/web_app_not_using_tls_last_version/query.rego
@@ -11,7 +11,7 @@ CxPolicy[result] { # resource of type Microsoft.Web/sites without child resource
 	value := site.value
 	path := site.path
 
-	children_array := arm_lib.get_children(doc, value, path)
+	children_array := get_children_ext(doc, value, path)
 
 	count(children_array) == 0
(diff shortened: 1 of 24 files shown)


cx-artur-ribeiro (changes requested):
Add a sample, for json and bicep, with 'app' instead of 'web' please.

cx-artur-ribeiro (approved):
LGTM

cx-alon-rosenhek (approved):

PR #8092, opened 2026-07-28
refactor(descriptions): remove pkg/descriptions and related CLI flag

cx-ricardo-jesus:
- Currently the descriptions package, does not fetch any cis_descriptions.

**Proposed Changes**
- Removed E2E test 046, since the purpose of this unit test is to simulate the --disable-full-descriptions, flag doesn't make sense to have it since the descriptions package has been removed.
- Renamed the E2E tests to be ordered, since one of them was removed.
- Removed any reference to the --disable-full-descriptions flag on kics documentation.
- Fully removed `pkg/descriptions`.
- Implemented a new function inside `pkg/scan/client.go` that checks if the current KICS version is up-to-date by querying the GitHub releases API. It early returns assuming it's the latest if the version is "development", an HTTP request fails, a 404 response is returned, or other errors occur reading the API response. Otherwise compares the local version against the tag_name from the release response.


pkg/scan/post_scan.go
```diff
@@ -2,14 +2,15 @@ package scan
 
 import (
 	_ "embed" // Embed kics CLI img and scan-flags
+	// "encoding/json"
+	// "fmt"
```
cx-miguel-dasilva, line 6:
```suggestion
```
Please remove unused commented deps

cx-miguel-dasilva (changes requested):
Nice work!

cx-artur-ribeiro (approved):
LGTM

PR #8095, opened 2026-07-30
fix(query): added missing case to "Last User is Root" Dockerfile query.

cx-andre-pereira:
- Currently the [implementation for the "Last User is Root" Dockerfile query](https://github.com/Checkmarx/kics/blob/9a8916683a4da61ba0b3f459fd8f0160cb235f47/assets/queries/dockerfile/last_user_is_root/query.rego) only checks for the scenario of ```USER root```, when commands like ```USER 0```, ```USER root:root``` and ```USER 0:0``` are equivalent and should also flag. 

**Proposed Changes**
- Added the [missing variations](https://github.com/Checkmarx/kics/pull/8095/changes#diff-3f1c0a7ae0668133a828c43e5d7fd8c89c24098cc8a169192fd69dba20dbd9e3) as valid inputs in the query.
- Added 3 new positive tests, one for each new variation.


cx-artur-ribeiro (dismissed):
LGTM

cx-artur-ribeiro (approved):

PR #8097, opened 2026-08-05
feat(queries): new queries to check if synapse workspace managed virtual network is enabled

cx-ricardo-jesus:
Closes #

- Currently, no queries check's if a synapse workspace managed virtual network is enabled.

**Proposed Changes**
- Implemented two queries called `"Synapse Workspace Managed Virtual Network Not Enabled"` for Terraform/Azure and AzureResourceManager.
- Both queries return a positive result if the field `managed_virtual_network_enabled` for the Terraform `azurerm_synapse_workspace` resource type and the field `managedVirtualResource` for the AzureResourceManager `Microsoft.Synapse/workspaces` resource type are not defined or defined to false/value different from `default`, respectively.


cx-artur-ribeiro (approved):
LGTM

PR #8099, opened 2026-08-14
fix(query): Fix for Missing Backslash Support on Copy_With_More_Than_Two_Arguments_Not_Ending_With_Slash query

cx-andre-pereira:
- Currently the [Copy With More Than Two Arguments Not Ending With Slash](https://github.com/Checkmarx/kics/tree/53dbe657bc61a8ba4b4c7e60b2f5b5c4c2ffc49a/assets/queries/dockerfile/copy_with_more_than_two_arguments_not_ending_with_slash) query is missing support for the use of backslashes ```\``` over forward slashes ```/``` for valid samples such as: 
 ```
# escape=`
ARG ANY_IMAGE
FROM ${ANY_IMAGE}
COPY readme.md *.xml* .\Config\example\
```
- Here the ```# escape=` ``` directive "[sets the character used to escape characters in a Dockerfile](https://docs.docker.com/reference/dockerfile/#escape)" , allows the use of backslashes in file paths (***useful for ease of use with Windows PowerShell***) . 

- Our Dockerfile parser seems to properly interpret the directive but the query [does not take backslashes into account](https://github.com/Checkmarx/kics/blob/53dbe657bc61a8ba4b4c7e60b2f5b5c4c2ffc49a/assets/queries/dockerfile/copy_with_more_than_two_arguments_not_ending_with_slash/query.rego#L13), resulting in **False Positive** results when the command ends with the valid backslash.

**Proposed Changes**
- Adjusted the query to check that the target ```Copy``` command [does not end with a slash of any kind](https://github.com/Checkmarx/kics/pull/8099/changes#diff-071ed625d424e28146885c2f721caa7ec57a64b89eee8bed7bb24c5018b4969bR14), back or forward.

- The [negative4.dockerfile](https://github.com/Checkmarx/kics/pull/8099/changes#diff-59533485a80a5eb1ece365dcb660e187d9c378be55a7f1973ce1d7d364b72710) sample makes use of double backslashes as an alternative valid method ([dockerfile - backslashes](https://docs.docker.com/reference/dockerfile/#backslashes)).

> [!Note]
> Accepting the backslash as a valid slash at the end of the command without checking for the ```# escape=` ``` directive might seem careless, but given the current parsing logic the lack of this directive will result in a command without any trailing slash on the final payload ensuring the query will flag as demonstrated through the [added positive samples](https://github.com/Checkmarx/kics/pull/8099/changes#diff-aa44e888bbe612d33ff22ea865d49b43a063519b719c9b8a44f983fd0296ddd7).


cx-artur-ribeiro (approved):
LGTM

PR #8101, opened 2026-08-19
fix(query): change to keyExpectedValue on ARM 'default_azure_storage_account_network_access_is_too_permissive' query

cx-andre-pereira:
- Currently the "[keyExpectedValue](https://github.com/Checkmarx/kics/blob/082dcacc2e36803c7bb0e6cd4a547c445004d0d2/assets/queries/azureResourceManager/default_azure_storage_account_network_access_is_too_permissive/query.rego#L24)" field on the "```default_azure_storage_account_network_access_is_too_permissive```" query states that "```publicNetworkAccess should be set to false```", when the actual value recommendation should be setting the target ```publicNetworkAccess ``` field to '```Disabled```'. 
- [Official documentation](https://learn.microsoft.com/en-us/azure/templates/microsoft.storage/storageaccounts?pivots=deployment-language-bicep) explicitly lists the valid values for the target field, those being : 
  - ``` 'Disabled' | 'Enabled' | 'SecuredByPerimeter'```. 

**Proposed Changes**
- Simply adjusted the query's ```keyExpectedValue``` to properly reflect the valid options; now it is recommended that the ```publicNetworkAccess``` field be set to '```Disabled```'. 

- Another small adjustment was made on the "[deny](https://github.com/Checkmarx/kics/pull/8101/changes#diff-018fce570508db6e1ba2821b5c7b980449e0c4285d963ec15cfcef202534d412L24)" recommendation to reflect use of single quotes (for consistency sake given the [keyActualValue(s) used](https://github.com/Checkmarx/kics/blob/082dcacc2e36803c7bb0e6cd4a547c445004d0d2/assets/queries/azureResourceManager/default_azure_storage_account_network_access_is_too_permissive/query.rego#L81)) , and the use of uppercase on the value ('**D**eny'), as per the [documented valid values](https://learn.microsoft.com/en-us/azure/templates/microsoft.storage/storageaccounts?pivots=deployment-language-bicep) on the ```defaultAction``` field.

> [!NOTE]
> The query does not explicitly handle the case for the '```SecuredByPerimeter```' value, considering that [older documentation (2021)](https://learn.microsoft.com/en-us/azure/templates/microsoft.storage/2021-06-01/storageaccounts?pivots=deployment-language-bicep)  does not mention this value as a possibility, it is likely the query did not take it into account when first developed. 
>

- It was decided that the '```SecuredByPerimeter```' value be considered unsafe and so changes to the query structure itself took place: 
  -  Adjusted the [auxiliary functions](https://github.com/Checkmarx/kics/pull/8101/changes#diff-018fce570508db6e1ba2821b5c7b980449e0c4285d963ec15cfcef202534d412L29-L74) of the query to, instead of specifically targeting the "```Enabled```"/"```Allow```" values as the only ones that should be flagged, take a more reliable approach of flagging for any value when ```publicNetworkAccess``` and/or ```defaultAction```  are not set to "```Disabled```"/"```Deny```" .

- With this change any additions to the list of possible values for each of the fields will cause the query to flag said value and properly display it on the new [keyActualValue(s)](https://github.com/Checkmarx/kics/pull/8101/changes#diff-018fce570508db6e1ba2821b5c7b980449e0c4285d963ec15cfcef202534d412L89). 
- I believe this approach to be more reliable in the sense that it's likely that future new values will also be deemed **too permissive** and, even if they are not, this way we will see **False Positive** results pop up which are substantially easier to identify over the **False Negatives** that could go by unnoticed with the current approach. 

- Added new tests for scenarios involving the '```SecuredByPerimeter```' value of the ```publicNetworkAccess``` field. 


cx-artur-ribeiro (dismissed):
LGTM

cx-ricardo-jesus (dismissed):
LGTM

cx-ricardo-jesus (approved):
LGTM

PR #8105, opened 2026-08-31
fix(query): correct keyExpectedValue and keyActualValue in redshift_not_encrypted

cx-ricardo-jesus:
- The second rule in the `redshift_not_encrypted` query had reversed `keyExpectedValue` and `keyActualValue` fields. When a cluster has encrypted = false, the finding is correctly detected, but the displayed messages on expected vs actual values were backwards, telling users the opposite of what was actually wrong.

**Proposed Changes**
- Fixed keyExpectedValue to read: "aws_redshift_cluster.encrypted should be set to true" (was incorrectly set to "false").
- Fixed keyActualValue to read: "aws_redshift_cluster.encrypted is false" (was incorrectly set to "true").
- The remediation field remains correct and consistent with the fix: before: "false", after: "true"


cx-rui-araujo (approved):
LGTM

cx-alon-rosenhek (approved):

PR #8112, opened 2026-09-07
fix(filesystem): skip cache files by extension

cx-laura-rodrigues:
**Proposed Changes**
- Skip files matching the `.cache.` extension/pattern during the filesystem walk, before they reach the parser.
- Added unit test coverage confirming `.cache.` files are excluded and no longer processed.

Jira: AST-170544

### DIFF
--- a/pkg/analyzer/analyzer.go
+++ b/pkg/analyzer/analyzer.go
@@ -74,7 +74,7 @@ var (
 	blueprintRegexTargetScope                       = regexp.MustCompile(`("targetScope"|targetScope)\s*:`)
 	blueprintRegexProperties                        = regexp.MustCompile(`("properties"|properties)\s*:`)
 	buildahRegex                                    = regexp.MustCompile(`buildah\s*from\s*\w+`)
-	dockerComposeServicesRegex                      = regexp.MustCompile(`services\s*:[\w\W]+(image|build)\s*:`)
+	dockerComposeServicesRegex                      = regexp.MustCompile(`(?m)^\s*"?services"?\s*:[\w\W]+^\s*"?(image|build)"?\s*:`)
 	crossPlaneRegex                                 = regexp.MustCompile(`"?apiVersion"?\s*:\s*(\w+\.)+crossplane\.io/v\w+\s*`)
 	knativeRegex                                    = regexp.MustCompile(`"?apiVersion"?\s*:\s*(\w+\.)+knative\.dev/v\w+\s*`)
 	pulumiNameRegex                                 = regexp.MustCompile(`name\s*:`)
--- a/pkg/analyzer/analyzer_test.go
+++ b/pkg/analyzer/analyzer_test.go
@@ -98,6 +98,23 @@ func TestAnalyzer_Analyze(t *testing.T) {
 			MaxFileSize:          -1,
 		},
 		{
+			name: "analyze_test_not_dockercompose_false_positive",
+			paths: []string{
+				filepath.FromSlash("../../test/fixtures/analyzer_test_dockercompose_false_positive/azure_marketplace.json"),
+			},
+			wantTypes: []string{},
+			wantExclude: []string{
+				filepath.FromSlash("../../test/fixtures/analyzer_test_dockercompose_false_positive/azure_marketplace.json"),
+			},
+			typesFromFlag:        []string{""},
+			excludeTypesFromFlag: []string{""},
+			wantLOC:              0,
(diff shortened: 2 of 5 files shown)


cx-artur-ribeiro (changes requested):
Hi @cx-laura-rodrigues, 
First of all, thanks for the contribution!

I believe the problem here is not catching `.cache` files, although I understand a similar solution was applied for terraform cache files on line 31 of pkg/engine/provider/filesystem.go.
The problem seems to be the misinterpretation of the files in question as belonging to the docker compose platform, when in fact they shouldn't be correlated to that platform.

Image

KICS does not have an extension per se for docker compose files, so it goes through [checkContent](https://github.com/Checkmarx/kics/blob/8046b364af4945934b4c4877df0a9939e6abe0d2/pkg/analyzer/analyzer.go#L565), which is used for json and yaml files that could belong to multiple platforms. Those files are then compared against multiple regexes to understand which platform the file content belongs to.

In this case, after scanning a file with the same content as your test files, I found that the problem is that the regex used for docker compose is too loose, causing it to match the content of a file that is not actually a docker compose file.
Below you can see the current regex compared to a file with similar content:
 
Image

Additionally, if we had a different file name but the same content, we would still hit the same symptom your pull request addresses, since the file wouldn't be excluded by name.
For that reason, I would suggest making the regex more robust in [analyzer](https://github.com/Checkmarx/kics/blob/8046b364af4945934b4c4877df0a9939e6abe0d2/pkg/analyzer/analyzer.go#L80).

For example, something along these lines:
```go
dockerComposeServicesRegex                      = regexp.MustCompile(`(?m)^\s*"?services"?\s*:[\w\W]+^\s*"?(image|build)"?\s*:`)
``` 

After testing and adding a new test file for this specific case, which I can also provide for testing purposes, I got no regressions on the files we already have enough observability into to know that KICS supports and scans correctly. We don't lose any results we already have and we fortify our content analysis by applying a better regex for comparison.

```json
[
  {
    "offer": "sample-webservices",
    "publisher": "checkmarx",
    "sku": "sku1",
    "urn": "checkmarx:sample-webservices:sku1:1.0.0",
    "version": "1.0.0"
  },
  {
    "offer": "sample-image",
    "publisher": "checkmarx",
    "sku": "sku2",
    "urn": "checkmarx:sample-image:sku2:1.0.0",
    "version": "1.0.0"
  }
]
```

Let me know what you think!
Thanks again for helping us improve kics! 🥳

pkg/analyzer/analyzer.go
```diff
@@ -77,7 +74,7 @@ var (
 	blueprintRegexTargetScope                       = regexp.MustCompile(`("targetScope"|targetScope)\s*:`)
 	blueprintRegexProperties                        = regexp.MustCompile(`("properties"|properties)\s*:`)
 	buildahRegex                                    = regexp.MustCompile(`buildah\s*from\s*\w+`)
-	dockerComposeServicesRegex                      = regexp.MustCompile(`services\s*:[\w\W]+(image|build)\s*:`)
+	dockerComposeServicesRegex                      = regexp.MustCompile(`(?m)^\s*"?services"?\s*:[\w\W]+^\s*"?(image|build)"?\s*:`)
```
cx-artur-ribeiro, line 77: My suggestion used `(?m)^\s*"?` as well as `^\s*"?`. Since we have other examples in the same file that fixed this problem:

```
serverlessProviderRegex                         = regexp.MustCompile(`(^|\n)provider\s*:`)
```

I would suggest changing it to the following:

```suggestion
	dockerComposeServicesRegex                      = regexp.MustCompile(`(^|\n)\s*"?services"?\s*:[\w\W]+(^|\n)\s*"?(image|build)"?\s*:`)
```

If you could check if everything else is okay after this change I would be gratefull 😄.

Sorry for the miss direction in my first review. It was meant to provide an example, which I used regex101 site to compile, while in go we might want to do differently, as well as maintain the solution aligned with all other regexes used in analyzer.go.

Let me know what you think @cx-laura-rodrigues.

cx-artur-ribeiro (changes requested):

pkg/analyzer/analyzer.go
cx-laura-rodrigues: All tests pass after applying your suggestion so I'm applying it :) thanks

cx-artur-ribeiro (approved):
LGTM

cx-diogo-rocha (approved):

PR #8114, opened 2026-09-08
fix(analyzer): improvement to dockerfile scanning

cx-andre-pereira:
>[!NOTE]
> A lot of links are to [Original PR](https://github.com/Checkmarx/kics/pull/7995), but diff is equivalent. 

The current implementation for dockerfile scanning is overly restrictive. The [GetExtension](https://github.com/Checkmarx/kics/tree/e1f23cad9640f55b963f22a116b04906b8c16ac6/pkg/utils/get_extension.go#L16) function can identify "dockerfile" type files in two ways as of now: 
  - Using the [Contains](https://github.com/Checkmarx/kics/tree/e1f23cad9640f55b963f22a116b04906b8c16ac6/pkg/utils/contains.go#L9) function on files that have [no extension set](https://github.com/Checkmarx/kics/blob/e1f23cad9640f55b963f22a116b04906b8c16ac6/pkg/utils/get_extension.go#L30), allowing detection of files that contain "Dockerfile" in a case-sensitive way, values such as: 
 ```
    - "Dockerfile"
    - "123Dockerfile"
    - "Dockerfile123" 
```
  - Checking that a file is: 
      - a valid text file and then, through the "[readPossibleDockerFile](https://github.com/Checkmarx/kics/tree/e1f23cad9640f55b963f22a116b04906b8c16ac6/pkg/utils/get_extension.go#L53C6-L53C28)" function:
        - Return true immediately if it is a file with suffix "[gitignore](https://git-scm.com/docs/gitignore)"
        - Return true if it contains a "FROM" (case-sensitive) command after skipping any commented lines "# ..."; 
    - Given all these conditions, the result for "[GetExtension](https://github.com/Checkmarx/kics/tree/e1f23cad9640f55b963f22a116b04906b8c16ac6/pkg/utils/get_extension.go#L16)" will be "```possibleDockerfile```". As a final check for the file to be included in a given scan, the analyzer's [worker](https://github.com/Checkmarx/kics/tree/e1f23cad9640f55b963f22a116b04906b8c16ac6/pkg/analyzer/analyzer.go#L417) calls the "[isDockerfile](https://github.com/Checkmarx/kics/tree/e1f23cad9640f55b963f22a116b04906b8c16ac6/pkg/analyzer/analyzer.go#L490)" function meant to ensure that only files containing the "FROM" and "RUN"(both case-sensitive) commands are scanned, files that fail this check are promptly set as "[unwanted](https://github.com/Checkmarx/kics/blob/e1f23cad9640f55b963f22a116b04906b8c16ac6/pkg/analyzer/analyzer.go#L451)".

- The way dockerfile identification is meant to work is as follows: 
  - Case-insensitive support for all "```dockerfile```", "```dockerfile.any_extension```" and "```any_name.dockerfile```" files. 
  - Check all files inside folders with the names : "```docker/```", "```dockerfile/```" or "```dockerfiles/```", also in a case-insensitive manner.
- Additionally "```.ubi8```" and "```.debian```" files should be included in the scan if they are valid docker configurations.

- The current implementation has a plethora of issues: 
  - It is case-sensitive both for "Dockerfile" identification as well as command scanning which is case-sensitive by [convention only](https://docs.docker.com/reference/dockerfile#format). 
  - Restricts dockerfile scanning to those that contain the "FROM" and "RUN" commands when the "FROM" would suffice. 
  - Does not properly skip empty lines and "[ARG](https://docs.docker.com/reference/dockerfile#understand-how-arg-and-from-interact)" type arguments when searching for the "FROM" statement in the "[readPossibleDockerFile](https://github.com/Checkmarx/kics/tree/e1f23cad9640f55b963f22a116b04906b8c16ac6/pkg/utils/get_extension.go#L53C6-L53C28)" function. 
  - Has a major redundancy by using both the "[readPossibleDockerFile](https://github.com/Checkmarx/kics/blob/e1f23cad9640f55b963f22a116b04906b8c16ac6/pkg/utils/get_extension.go#L53)" and the "[isDockerfile](https://github.com/Checkmarx/kics/blob/e1f23cad9640f55b963f22a116b04906b8c16ac6/pkg/analyzer/analyzer.go#L490)" functions. Both serve identical purposes and since the latter is only called when "readPossibleDockerFile" returns a positive result, this check should be simplified into a single function. 

**Proposed Changes**
- Reworked the "GetExtension" function so that it first :  
  - Calls the (new) "[isDockerfileExtension](https://github.com/Checkmarx/kics/pull/7995/changes#diff-e342d7fcd6b57c8f32c3a8d057b7f4b4c63e68ddfab412b5d6a18e829923533bR44)" function, here all files that are named "dockerfile.something" or "something.dockerfile" in a case-insensitive manner, or files inside the target folders mentioned before that are deemed valid dockerfile files (through "[readPossibleDockerFile](https://github.com/Checkmarx/kics/pull/7995/changes#diff-e342d7fcd6b57c8f32c3a8d057b7f4b4c63e68ddfab412b5d6a18e829923533bR85)") will be immediately determined as ".dockerfile" files. 

- The "[readPossibleDockerFile](https://github.com/Checkmarx/kics/pull/7995/changes#diff-e342d7fcd6b57c8f32c3a8d057b7f4b4c63e68ddfab412b5d6a18e829923533bR85)" function was changed to support both "ARG" commands as well as empty lines before the "FROM" command. Additionally the order of these checks was changed to enhance performance so that it is first checked if a line is [not a "FROM" command](https://github.com/Checkmarx/kics/pull/7995/changes#diff-e342d7fcd6b57c8f32c3a8d057b7f4b4c63e68ddfab412b5d6a18e829923533bR100); previously, since every line was compared to the "FROM" statement, a file with a large number of comments would take significantly longer to skim through. (tests on a dockerfile with 260 thousand comments were two times faster with this change) 

- In the [analyzer](https://github.com/Checkmarx/kics/pull/7995/changes#diff-7dc19f3d63f984f9ddd064188626e762c34bf2dc8f41f7ee0d43fa8db61cdc23) the worker was simplified, given the changes on the [utils/get_extension](https://github.com/Checkmarx/kics/pull/7995/changes#diff-e342d7fcd6b57c8f32c3a8d057b7f4b4c63e68ddfab412b5d6a18e829923533b)'s "GetExtension" function, we only need to worry about 2 results "```gitignore```" and "```.dockerfile```" , additionally the "isDockerfile" function is now gone; it was overly restrictive and served the same purpose as the "readPossibleDockerFile" function does now. 

- Unified all "dockerfile" type resource references to ".dockerfile", previously there was use of "```dockerfile```", "```Dockerfile```" and "```possibleDockerfile```", I deemed these unnecessary since they all represent the same thing, except for "possibleDockerfile" but files labeled as such could and are now determined as valid dockerfiles before being attributed a value for their "extension".  

- After initial reviews the case handling for "```gitignore```" values was simplified so that any "```.gitignore```" or any extensionless file with "```gitignore```" as a suffix will have identical flows, previously [unit-tests](https://github.com/Checkmarx/kics/pull/7995/changes#diff-67027e98d80320fd9c7c85e5510a9438add093866d1444f2210145c95c5ffd22) for the analyzer forced the latter to be added to the "[unwanted](https://github.com/Checkmarx/kics/pull/7995/changes#diff-7dc19f3d63f984f9ddd064188626e762c34bf2dc8f41f7ee0d43fa8db61cdc23L451)" path list, the test failed if we simple discarded the file in the analyze function before calling the worker (the way any invalid extension outside target folders is treated).

- ## Tests                                                                                                                                                                         
                                                                                                  
  Many new files were added to `test/fixtures/dockerfile` as well as a new folder `test/fixtures/negative_dockerfile`.                                                             
                                                                                                                                                                                   
- test/fixtures/dockerfile/                                                                                                                                                    
```                                                                                                                                         
  ├── Dockerfile-example                (pre-existing test)                               
  ├── corrupted_dockerfile              (pre-existing test)
  ├── any_name/                                                                                                                                                                                                                    
  │   ├── DOCKERfile.txt
  │   ├── Dockerfile.something                                                                                                                                                                                                     
  │   ├── any_name.debian                                                                                                                         
  │   ├── any_name.ubi8
  │   ├── dockerFILE
  │   ├── file.Dockerfile
  │   ├── file_2.DOCKERfile
  │   └── random_name
  ├── case_insensitive_tests/
  │   ├── DOCKERfile.txt
  │   ├── Dockerfile.something
  │   ├── any_name.debian
  │   ├── any_name.ubi8
  │   ├── dockerFILE
  │   ├── file.Dockerfile
  │   ├── file_2.DOCKERfile
  │   └── random_name
  ├── should_generate_payload/
  │   ├── base_image_reference
  │   ├── big_host_name
  │   ├── big_indent_from
  │   ├── full_sha_digest
  │   ├── host_ip_address
  │   ├── multiline_from_statement
  │   ├── platform_flag
  │   ├── platform_flag_and_alias
  │   ├── platform_flag_scratch
  │   ├── platform_with_arg_reference
  │   ├── use_of_host
  │   └── with_excluded_comments
  ├── test_folder_names/
  │   ├── docker/
  │   │   └── any_file.txt
  │   ├── dockerfile/
  │   │   └── any_file.txt
  │   └── dockerfiles/
  │       └── any_file.txt
  └── test_folder_names_case/
      ├── Docker/
      │   └── any_file.txt
      ├── Dockerfile/
      │   └── any_file.txt
      └── Dockerfiles/
          └── any_file.txt
```
  The files themselves include configurations with empty lines, `ARG` commands and comments before the target `FROM` command. Additionally all naming conventions are tested. Note that files inside the target folders would not be identified if they were not inside said folders (because `isDockerfileExtension` is [called](https://github.com/Checkmarx/kics/pull/7995/changes#diff-e342d7fcd6b57c8f32c3a8d057b7f4b4c63e68ddfab412b5d6a18e829923533bR44) before checking literal extension). 
  
  The `case_insensitive_tests` folder includes identical tests where all docker commands have been written in lowercase — note that most queries logic / the engine itself will not properly allow final results pointing to target lines when a query is triggered, but they still trigger similarly to comparable conventional dockerfiles (fixed in https://github.com/Checkmarx/kics/pull/8006).
  
  Lastly the "should_generate_payload" folder has 12 different **valid** variations of a docker "FROM" command to ensure the [regex used to identify dockerfiles](https://github.com/Checkmarx/kics/pull/7995/changes#diff-e342d7fcd6b57c8f32c3a8d057b7f4b4c63e68ddfab412b5d6a18e829923533bR15) do not exclude any of the samples. As the name of the folder implies all 12 samples should generate a payload ensuring the expected behavior.

- test/fixtures/negative_dockerfile/
```
  ├── CW671X02_EBM_EVENT_RULE              (pre-existing test)
  ├── not_dockerfile.debian                                                                                                                                                                                                        
  ├── not_dockerfile.txt
  ├── not_dockerfile.ubi8                                                                                                                                                                                                          
  └── should_not_generate_payload/                                                                                                           
      ├── invalid_syntax                                                                                                                                                                                                           
      ├── negative_email                                                                                                                       
      ├── negative_email_2                                                                                                                                                                                                         
      ├── negative_python
      ├── negative_python_2
      ├── negative_sparql
      └── negative_sql
```
  The `.debian` and `.ubi8` files here should not be flagged as valid dockerfiles since they do not correspond to valid docker configurations. `not_dockerfile.txt` should not be flagged since, although it contains a valid docker configuration, it has a `.txt` extension and is not named "dockerfile" (case-insensitive) explicitly — it only contains the word "dockerfile". 
  
The `should_not_generate_payload` samples test extensionless files containing `FROM` statements in non-Docker contexts (email header, Python imports, SPARQL queries, SQL) to ensure they are not misidentified as dockerfiles.

 These files are used for the [`util/get_extension` unit tests](https://github.com/Checkmarx/kics/pull/7995/changes#diff-beda894e6e1f38957b33086728fadab666b91b5a3d22817f32b9678232ddcbcd) and the new [E2E test 106](https://github.com/Checkmarx/kics/pull/7995/changes#diff-e987437b21739e745091de2fc87f9e42360a00f6e0479f2ebf3d3560e3a2c223).
 
 As a final note for regarding the tests I should point out that the [regex used to identify/exclude samples](https://github.com/Checkmarx/kics/pull/7995/changes#diff-e342d7fcd6b57c8f32c3a8d057b7f4b4c63e68ddfab412b5d6a18e829923533bR15) could never prevent false positives on Files with basic "```FROM <any_string>```" commands, often related to SQL syntax, in case the files with this text are either: 
 - Files without extension 
 - Files stored inside the target ("```docker/```", "```dockerfile/```" or "```dockerfiles/```") folders 

Moreover a new function, "[TestParser_Parse_CaseInsensitive](https://github.com/Checkmarx/kics/pull/7995/changes#diff-f5568accfdb4b8da27afe3045504007ecfe7aff76b6953a74a2521edada45a34R241)", was added to the [parser/docker/parser_test.go](https://github.com/Checkmarx/kics/pull/7995/changes#diff-f5568accfdb4b8da27afe3045504007ecfe7aff76b6953a74a2521edada45a34) unit tests; as the name implies it is meant to ensure docker samples are parsed identically regardless of casing in the [dockerfile commands](https://docs.docker.com/reference/dockerfile/) syntax.


### CODE REVIEW

File: pkg/engine/provider/filesystem_test.go [hunk 1]
```diff
@@ -629,31 +630,23 @@ func TestFileSystemSourceProvider_checkConditions(t *testing.T) {
 
 // TestFileSystemSourceProvider_AddExcluded tests the functions [AddExcluded()] and all the methods called by them
 func TestFileSystemSourceProvider_AddExcluded(t *testing.T) {
+	if runtime.GOOS == "windows" {
```
cx-artur-ribeiro: Instead of skipping the test due to the OS being `windows`, can we make the same solution as we have for other similar different behaviour tests, some examples already present on this codebase: 
- https://github.com/Checkmarx/kics/blob/cd467d15289298baf9005eae62411e42cb97d5a8/pkg/engine/similarity/similarity_id_test.go#L233
- https://github.com/Checkmarx/kics/blob/cd467d15289298baf9005eae62411e42cb97d5a8/e2e/utils/helper.go#L80
- https://github.com/Checkmarx/kics/blob/cd467d15289298baf9005eae62411e42cb97d5a8/pkg/report/model/cyclonedx_test.go#L81
- https://github.com/Checkmarx/kics/blob/cd467d15289298baf9005eae62411e42cb97d5a8/e2e/utils/json.go#L38
- https://github.com/Checkmarx/kics/blob/cd467d15289298baf9005eae62411e42cb97d5a8/pkg/engine/similarity/similarity_id_test.go#L233

If this is not possible, we can maintain the skip, we don't lose much since we run e2e and unit tests on other OS, but we might introduce any change to excluded files that work differently on windows that we want to be sure we catch. Paths are always something we want to test for different OS's.

cx-artur-ribeiro (changes requested):
Hi @cx-andre-pereira,
Just a small quality of life change. Other than that, I don't see any issues!

Let me know what you think!

pkg/engine/provider/filesystem_test.go
cx-andre-pereira: Definitely makes more sense to just add a windows exclusive assertion keeping the test relevant for that OS as well, done and tested locally in https://github.com/Checkmarx/kics/pull/8114/commits/66f1ceaf4d0bd2ef0d58f5596ab472515ee12926.

pkg/utils/get_extension_test.go
```diff
@@ -18,22 +18,106 @@ func TestGetExtension(t *testing.T) {
...
-			want:     "Dockerfile",
+			want:     ".dockerfile",
 			filePath: "../../Dockerfile",
 			toCreate: false,
 			err:      nil,
 		},
 		{
-			name:     "Get extension from a file not named as Dockerfile and without extension defined ('Dockerfile-example')",
-			want:     "possibleDockerfile",
-			filePath: "../../test/fixtures/dockerfile/Dockerfile-example",
```
cx-rui-araujo, line 29: Is this file being tested in other locations? If not, we can delete it and re-add this test. It's up to you whether you think it's important or whether it's already covered by other tests

pkg/engine/provider/filesystem_test.go
```diff
@@ -681,13 +667,31 @@ func TestFileSystemSourceProvider_AddExcluded(t *testing.T) {
...
+			err = fsystem.AddExcluded(tt.args.excludePaths)
 			if (err != nil) != tt.wantErr {
 				t.Errorf("AddExcluded() = %v, wantErr = %v", err, tt.wantErr)
 			}
-			got := getFSExcludes(tt.fields.fs)
-			if !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("AddExcluded() = %v, want = %v", got, tt.want)
+			got := getFSExcludes(fsystem)
+			want := tt.want
+			if tt.name == "test_too_many_levels_of_symbolic_links" && len(tt.args.excludePaths) > 0 {
```
cx-rui-araujo: Why are we changing this test? Was it flaky? 🤔

cx-rui-araujo (commented):
@cx-andre-pereira please check my review

pkg/engine/provider/filesystem_test.go
cx-andre-pereira: This change was done in https://github.com/Checkmarx/kics/pull/8114/changes/18d09f824d46ad2ef080de0ab5879c70c7ad96ec and improved in the last commit (https://github.com/Checkmarx/kics/pull/8114/commits/f7b9423c316e6b8d4bc008dcc98f523ec3e41aea) simply because i wanted to be able to run the tests locally and have them pass.

I kept getting errors even though github actions were successful, I realized that missing case handling for non-unix based OS was the cause and decided to fix that gap. As @cx-artur-ribeiro  showed in the previous comment this is a recurring occurrence/issue that can come up in the tests.

cx-rui-araujo, line 680: Why are we changing this test? Was it flaky? 🤔

pkg/utils/get_extension_test.go
cx-andre-pereira: The file itself is used for the [E2E test 106](e2e/testcases/e2e-cli-106_valid_dockerfile_detected.go) added in this PR, since it scans the whole "[fixtures/dockerfile](https://github.com/Checkmarx/kics/tree/AST-140477--Improvement-to-dockerfile-scanning/test/fixtures/dockerfile)" folder, additionally it is used in the [line_counter_test.go](https://github.com/Checkmarx/kics/blob/AST-140477--Improvement-to-dockerfile-scanning/pkg/utils/line_counter_test.go#L25) file (which i didn't realize until now). 

I didn't remove the file simply to make sure we kept all existing test, which are very few(2) in the [current fixtures/dockerfile folder](https://github.com/Checkmarx/kics/tree/master/test/fixtures/dockerfile), even if it is a relatively redundant test file given all the newly added ones I think it is worth it to keep it.

pkg/engine/provider/filesystem_test.go
cx-andre-pereira: After consideration this change was reworked in https://github.com/Checkmarx/kics/pull/8114/commits/b120fff6d52a8b75c9c7e9747fa3c099acf95e9f. Since local Windows disables symlinks by default (core.symlinks=false), a local run on a Windows environment fails.

The Windows image used to run tests on our pipeline materializes ```test/fixtures/link_test/eloop_link``` as a proper symlink and reaches the correct error message + empty array [] result, whilst local runs get a regular file containing eloop_link (os.Stat succeeds) and hit the fail condition : ```AddExcluded() = [eloop_link], want = [] fail condition```.

The test logic was kept largely the same, with the fix for re-using the ```FileSystemSourceProvider``` instance retained ([_pkg/engine/provider/filesystem_test.go:630_](https://github.com/Checkmarx/kics/pull/8114/changes/b120fff6d52a8b75c9c7e9747fa3c099acf95e9f#diff-de09d698b546eaee34f45ed9dabd91c6f73718e5ad369430717007131e0393a1L670) - https://github.com/Checkmarx/kics/pull/8114/changes/8c55d3cd5393a76e61f745ad37b73d43d97dbec7) - to avoid map excludes state leaking between ```test_too_many_levels_of_symbolic_links``` and ```test_add_excluded```, behavior that was only identified given that a local run will result in ```[eloop_link config_test], want = [config_test]``` fail condition.

cx-artur-ribeiro (approved):
Nice job André, LGTM.

cx-rui-araujo (approved):
LGTM

aidigy-com:
Thank you very much for your efforts! :+1:

Unfortunately, however, the policy has now become stricter instead of being relaxed: None of our `Dockerfile-dev`, `Dockerfile-*`, or `Dockerfile-*-dev` files in **hundreds of projects** are being scanned anymore. This is a serious problem for us. :-1:

Since we can’t simply rename our Dockerfiles to adapt them to the changed matching logic, I’m really hoping for a very short-term hotfix at your upstream end, as even working around the problem by creating custom jobs to pin the old image involves an enormous amount of work.

Regardless, I'd like to express my gratitude for your ongoing maintenance and expansion of KICS—it's an indispensable tool in our daily work on GitLab.

aidigy-com:
Apologies — only after further investigation did we notice that GitLab had not even upgraded to 2.2.0 yet:

```
[INFO] [kics] [2026-09-20T11:41:11Z] ▶ GitLab kics analyzer v6.13.0
...
```

```
docker run --rm registry.gitlab.com/security-products/kics:6.13.0 kics version
Keeping Infrastructure as Code Secure v2.1.20
```

Contrary to what I had previously assumed, the issue we are experiencing is very likely fixed by the changes of this PR. Unfortunately, we could not test against 2.2.0, as no image for it has been released yet.

Further information on the behavior of KICS for Dockerfiles with different naming and content can be found in the corresponding GitLab issue and test repository:

https://gitlab.com/gitlab-org/gitlab/-/work_items/630002

https://gitlab.com/aidigy.com/issue-630002

Of those, test-1 and test-5 represent the failure scenarios. To summarize:

- If a `Dockerfile-dev` contains only a `FROM` and `COPY` instruction, it is skipped. If it contains at least one `RUN` instruction, it is scanned.
- Exception: if it is a multi-stage `Dockerfile-dev` that uses `FROM ${IMAGE}` notation, it is skipped even if a `RUN` instruction is present

aidigy-com:
As for a multi-stage `Dockerfile-dev` that uses `FROM ${IMAGE}` notation we could not find a workaround by altering the Dockerfile alone, we have implented an override for the `kics-iac-sast` job. There we replace a `# kics-dockerfile-fix-include` with `FROM scratch AS kics-dockerfile-fix-include` before the scan, so KICS won't skip it.

See: https://gitlab.com/gitlab-org/gitlab/-/work_items/630002#note_3879546221

PR #8120, opened 2026-09-14
fix(queries): correct mismatched expected/actual values across 22 queries

cx-artur-ribeiro:
- An audit of the query catalog found 22 queries where keyExpectedValue and/or keyActualValue were reversed, identical, or repeated a recommendation instead of describing the detected state, making findings confusing or misleading to users.

**Proposed Changes**
- Fixed reversed expected/actual values in 5 queries: api_gateway_without_security_policy, sql_server_ingress_from_any_ip, iam_policy_on_user, iam_managed_policy_applied_to_a_user, and the two provisioner branches of ec2_instance_using_api_keys.
- Fixed the same systemic reversal across 8 copied secret-exposure queries (Amplify, DocDB, DMS, Directory Service), and fixed a copy/paste bug in amplify_app_oauth_token_exposed where messages referenced BasicAuthConfig.Password instead of OauthToken.
- Fixed security_requirement_object_with_wrong_scopes, which had identical expected and actual text.
- Fixed 7 queries where keyActualValue repeated a recommendation instead of reporting the detected state: missing_version_specification_in_dnf_install, the 3 API Gateway usage-plan/access-log association queries, and the 3 ecs_service_without_running_tasks variants, and fixed a %&s format typo in the Ansible ECS query.
- Fixed misleading actual-value phrasing in email_notifications_set_off.


assets/queries/cloudFormation/aws/docdb_cluster_master_password_in_plaintext/query.rego
```diff
@@ -48,7 +48,7 @@ CxPolicy[result] {
 		"searchKey": sprintf("Resources.%s.Properties.MasterUserPassword", [key]),
 		"issueType": "IncorrectValue",
 		"keyExpectedValue": sprintf("Resources.%s.Properties.MasterUserPassword must not be in plain text string", [key]),
```
cx-andre-pereira, line 50:
```suggestion
		"keyExpectedValue": sprintf("Resources.%s.Properties.MasterUserPassword must be defined as a parameter or have a secret manager referenced", [key]),
```
Maybe here we could use the original (reversed) value from the ```keyActualValue``` field.

Feels like it might be more helpful since users will already know value is in "plaintext" from the corrected ```keyActualValue```. 

The "```use parameter or secret manager reference```" excerpt provides a better "fix suggestion" of sorts. 

**Note**: same suggestion applies to ```CxPolicy``` just after this one, and a lot of other cloudformation queries with the same ```must not be in plain text string``` expected value : 
```
amplify_app_access_token_exposed
amplify_app_basic_auth_config_password_exposed
amplify_app_oauth_token_exposed
amplify_branch_basic_auth_config_password_exposed
directory_service_simple_ad_password_exposed
dms_endpoint_mongo_db_settings_password_exposed
dms_endpoint_password_exposed
```

Either way is fine by me though.

cx-andre-pereira (approved):
LGTM!

cx-andre-pereira (approved):
LGTM!

PR #8121, opened 2026-09-15
fix(detector): fix data race in dockerfile line detection for multiple dockerfiles projects

cx-artur-ribeiro:
- DetectLine mutated the shared slice behind file.LinesOriginalData in place and that pointer is shared by every query scanning the same file. Since queries run concurrently, this raced and could `panic with runtime error: slice bounds out of range [:-1]`, confirmed with go test -race and reproduced end-to-end on a real scan.

**Proposed Changes**
- pkg/detector/docker/docker_detect.go: pass `prepareDockerFileLines` a private copy of the lines instead of the shared `*file.LinesOriginalData` slice, so concurrent queries no longer mutate each other's data;
- pkg/detector/docker/docker_detect_test.go: add **TestDetectDockerLineConcurrentAccess**, which reproduces the race under -race before the fix and passes after;
- pkg/detector/detector.go, pkg/detector/detector_test.go, pkg/engine/vulnerability_builder.go: remove DetectLine.SetupLogs/logWithFields, a second unsynchronized field written by every query concurrently - dead code, since DetectLine already takes the logger as a parameter.


cx-andre-pereira (approved):
LGTM !

